PR #6, opened 2019-04-25
gNMI Cli has been added to the Dockerfile-README has been updated

adibrastegarnia:


tools/test/devicesim/Dockerfile
```diff
@@ -37,13 +37,16 @@ RUN mkdir -p $GOPATH \
       github.com/google/gnxi/gnmi_capabilities \
       github.com/google/gnxi/gnmi_get \
       github.com/google/gnxi/gnmi_set \
-      github.com/google/gnxi/gnmi_target
+      github.com/google/gnxi/gnmi_target \
+      github.com/openconfig/gnmi/cmd/gnmi_cli
```
Andrea-Campanella: I would agree with Sean here. even though the gNMI_cli is important here I think we do not want it deployed in the emulator docker container.

adibrastegarnia:
@SeanCondon @Andrea-Campanella 

Thanks for your comments. I totally agree with what you said about  running gnmi_cli outside of the Docker. So according to the comments, we have two options:
1- Add gnmi_cli to the Dockerfile as an option in case someone needs to run it inside the docker and mention gnmi_cli externally is the preferred option. I think it is a good idea to have an image that contains everything.

2- Leave the Dockerfile as it is but add the instructions for installing the gnmi_cli.

Please let me know what the group prefers and I will make the changes accordingly.

SeanCondon:
Hi Adib - thanks for coming back on this - yes add it to the Dockerfile

Andrea-Campanella:
I like it with the possibility of having both, either in docker or either on your machine.

PR #9, opened 2019-04-26
Create a minimum set of directories to have a preliminary project layout

adibrastegarnia:
Since the group is adding some preliminary packages to the repository, that would be nice to move the packages to the right place. I created a minimum set of directories to organize the code so far. We can add, remove, and change them as we move forward. We can also discuss the project layout here.


SeanCondon:
Again Adib thanks for taking the initiative. I've been talking with the other guys about this structure. I'm not convinced of how useful it is for us until we get more of a picture of what the project will be like. None of the openconfig projects use all of this structure, and I cannot find much evidence of it elsewhere either. We are having a work week to discuss this project in more detail on May 5-9, so we may decide then

Andrea-Campanella:
Adib, thanks for the initiative, the idea you are going for is definitely along the lines of what we want but nt have decided yet. I also think that this could be done on a need to need basis. given the NB-CORE-SB architecture do you think the folder structure could mimic this ?

adibrastegarnia:
@Andrea-Campanella @SeanCondon 

Thanks for your comments. I totally agree with what you said and I know we still need to think about it and then decide :) I just bring it up that the group can discuss and then decide because before merging more code and allows others to contribute in the code, we need to have a preliminary layout to distinguish public packages from binaries (i.e. main packages) and tests at least. As you pointed out, we don't need all of them and should be added on a need to need basis except those are that are trivial independent of the context.  Just my two cents :)

SeanCondon:
Deferred until a decision is made on structure

PR #10, opened 2019-04-30
device simulator extended to support gNOI

adibrastegarnia:
The device simulator has been extended to support gNOI targets. The simulator works at three different modes which allows a user  to run gNMI targets, gNOI targets, and gNMI and gNOI targets simultaneously. I still need to do some tests to make sure the targets can be created properly. Please let me know if you have any comments.


adibrastegarnia:
@SeanCondon 

Thanks for your comments. I will take care of them in the next few days. For the last comment, I think it is possible to use the certs without generating new ones.

PR #11, opened 2019-04-30
minor improvements in the devsim

adibrastegarnia:
The comments of the previous pull request have been addressed.


adibrastegarnia:
@SeanCondon 

Thanks for your suggestions. I will take care of them. Let me take a look and then update you.

adibrastegarnia:
@SeanCondon 

I almost took care of all of your comments and I will push my changes soon after doing some tests. Regarding some of your comments: 

I'm not clear why it's a manual step to provision certificates for GNOI - couldn't this be done in the setup? It looks like it would have to be run on all 3 VMs. I could not get this to work when I was running the standalone instance.>

** I  think it shouldn't part of the setup because that is one of the services of gNOI cert package (i.e it calls install rpc). Suppose we add 5 more gNOI services and a user wants to test one of them which is not certificate management service.  So we shouldn't call any service in advance if the user doesn't need it.  What do you think?**.

> What does gnoi_target command do - it gives me a seg fault when I run "gnoi_target -bind_address localhost:50002" - maybe just add some note

> **It is a binary that implements a gNOI Target with the Certificate Management service (I will add this note). This command works for me: gnoi_target  -bind_address :50002**

SeanCondon:
Hi Adib - there is one other tiny issue that I've seen - when getting the "Message of the Day" in the gNMI user manual the value is not replaced. I think it's that the SED command is not being run properly in your script because it can't find the target config from its new location. Thanks, Sean

SeanCondon:
Never mind - I pushed a patch

PR #14, opened 2019-05-04
gNMI SB implementation has been improved to make it more general

adibrastegarnia:
@Andrea-Campanella @SeanCondon 

I am working on the gNMI southbound implementation a little bit to make it more general. I am still working on it but please take a look and let me know if you have any comments. 
A few things that we need to think about: 
1.  I don't think we need a map to store targets, if it is needed please let me know. I see a usage of that in synchronizer package.  Is it possible to achieve the goal without using a map?

2. I think there should be a general struct which contains "context" and other targets should be created based on composition. I am still not sure we should have one target for all gnoi services or we should  have multiple structs where each represents a service?

 My plan is to add gnoi API based on the clients that are available at this moment as well .


SeanCondon:
Adib. Thanks again for taking the initiative with this.

I have a few Points to put to you though 
1) this project is onos-config- i dont see any need for gnoi at this point
2) firststeps will probably be merged back in master tomorrow after we review it as it is. I think once this is done, you will be able to rebase and resubmit. Also i think it would make it easier to analyse if you deal with one topic per patch.
3) devicesim is most up to date in master, so should not be updated in firststeps

Andrea-Campanella:
Hi Adib, 

thanks for the contribution and the generalisation. 

I agree with Sean I woudl leave gNOI on the southbound out of the picture at this moment. If you want it can be added as a separate branch/commit. 
Another couple of things:
1. I've updated the southbound branch with subscription (not fully tested) can you rebase on top of that ? 
2. I would divide it completely from first steps branch and base this patch is off my southbound branch only

Thanks,
Andrea

adibrastegarnia:
@Andrea-Campanella @SeanCondon 

Thanks for your comments. OK. I will open a new pull request to make the changes on southbound_gnmi branch then we will merge with master or firststeps branch later. So you can close this one.

Andrea-Campanella:
Closing this pull request, we will move it over to the new repo we are creating. Please hold on.

PR #69, opened 2019-05-09
Removed RESTCONF service

SeanCondon:


deployments/helm/templates/service.yaml
```diff
@@ -13,6 +13,4 @@ spec:
   clusterIP: None
   ports:
     - name: h2
-      port: 5150
-    - name: restconf
-      port: 8080
\ No newline at end of file
+      port: 5150
```
Andrea-Campanella: please add a line at the end of this file

PR #77, opened 2019-05-09
Path utils refactor

ray-milkey:


pkg/southbound/synchronizer/synchronizer.go
```diff
@@ -56,7 +56,13 @@ func Devicesync(changeStore *store.ChangeStore,
...
-		gnmiChange := change.GnmiChange()
+		gnmiChange, parseError := change.GnmiChange()
+
+		if parseError != nil {
+			log.Println("Parsing error for Gnmi change ", parseError)
+			continue
+		}
+
 		log.Println("Change formatted to gNMI setRequest", gnmiChange)
 		setResponse, err := southbound.Set(target, &gnmiChange)
```
SeanCondon: Don't call setResponse if you got an error above

PR #149, opened 2019-05-13
Update run.md

tomikazi:
Added options for certs.

### DIFF
--- a/docs/run.md
+++ b/docs/run.md
@@ -11,9 +11,9 @@ go run github.com/onosproject/onos-config/cmd/onos-config-manager \
     -changeStore=$HOME/go/src/github.com/onosproject/onos-config/configs/changeStore-sample.json \
     -deviceStore=$HOME/go/src/github.com/onosproject/onos-config/configs/deviceStore-sample.json \
     -networkStore=$HOME/go/src/github.com/onosproject/onos-config/configs/networkStore-sample.json \
-    -caPath tools/test/devicesim/certs/onfca.crt \
-    -keyPath tools/test/devicesim/certs/localhost.key \
-    -certPath tools/test/devicesim/certs/localhost.crt
+    -caPath $HOME/go/src/github.com/onosproject/onos-config/tools/test/devicesim/certs/onfca.crt \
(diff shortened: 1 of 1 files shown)


docs/run.md
```diff
@@ -10,7 +10,10 @@ go run github.com/onosproject/onos-config/cmd/onos-config-manager \
     -configStore=$HOME/go/src/github.com/onosproject/onos-config/configs/configStore-sample.json \
     -changeStore=$HOME/go/src/github.com/onosproject/onos-config/configs/changeStore-sample.json \
     -deviceStore=$HOME/go/src/github.com/onosproject/onos-config/configs/deviceStore-sample.json \
-    -networkStore=$HOME/go/src/github.com/onosproject/onos-config/configs/networkStore-sample.json
+    -networkStore=$HOME/go/src/github.com/onosproject/onos-config/configs/networkStore-sample.json \
+    -caPath tools/test/devicesim/certs/onfca.crt \
```
SeanCondon: Probably should be an absolute path like the store files above it

docs/run.md
```diff
@@ -20,7 +23,10 @@ docker run -p 5150:5150 -v `pwd`/configs:/etc/onos-config-manager -it onosprojec
     -configStore=/etc/onos-config-manager/configStore-sample.json \
     -changeStore=/etc/onos-config-manager/changeStore-sample.json \
     -deviceStore=/etc/onos-config-manager/deviceStore-sample.json \
-    -networkStore=/etc/onos-config-manager/networkStore-sample.json
+    -networkStore=/etc/onos-config-manager/networkStore-sample.json \
+    -caPath tools/test/devicesim/certs/onfca.crt \
```
SeanCondon: Will this know where do get the files? Did you try it out?

tomikazi: No... I wasn't really fixing this... I care that the top part is fixed... docker can be addressed separately.

SeanCondon: Sorry wasn't trying to be awkward, it's just that I thought it would be more consistent to have an absolute path. I'll try it out here with Docker image

tomikazi: No worries... I was just fixing the parts that were clearly broken, but still working on the side with Andrea on the subscribe so did not have time to test the docker bits. I figured more change would come here shortly anyway from @kuujo.

PR #157, opened 2019-05-14
Quietly ignore duplicate changes to a device Fixes #154

### LINKED ISSUE #154: Quietly ignore duplicate changes to a device
When a change is sent down through gNMI NB 3 checks need to be added to avoid duplicates
1) When the change already exists (i.e. something with the same hash exists) it need not overwrite the older version of same in Change store

2) Additionally when this change is the same as the last change made in a Configuration, the list of changes of the Configuration should not be extended

3) If this is the case for all of the targets in a gNMI set then an error should be returned for that Set. If less that all of many targets is found to already exist then the command succeeds, quietly ignoring the duplicates. The duplicates are not present in the SetResponse

SeanCondon:


pkg/northbound/gnmi/set.go
```diff
@@ -73,12 +73,16 @@ func (s *Server) Set(ctx context.Context, req *gnmi.SetRequest) (*gnmi.SetRespon
 		return nil, err
 	}
 	updateResults := make([]*gnmi.UpdateResult, 0)
-	configName := "Running"
 	for target, updates := range targetUpdates {
-		changeID, err := manager.GetManager().SetNetworkConfig(target, configName, updates, targetRemoves[target])
+		changeID, err := manager.GetManager().SetNetworkConfig(target, target, updates, targetRemoves[target])
```
Andrea-Campanella: why is this changed to double target ?

pkg/northbound/gnmi/set.go
```diff
@@ -73,12 +73,16 @@ func (s *Server) Set(ctx context.Context, req *gnmi.SetRequest) (*gnmi.SetRespon
 		return nil, err
 	}
 	updateResults := make([]*gnmi.UpdateResult, 0)
-	configName := "Running"
```
Andrea-Campanella: we ought to come up with some constants for the gNMI Nb generated configurations since gNMI can't specify nor this nor the User

pkg/northbound/gnmi/set.go
SeanCondon: The thing is when you have more than 1 target each one should have its own Configuration - not just a shared "Running" one between the 2. Therefore I have given the configuration the same name as the target - maybe should change it to "<targetName>Running". If you want to see the effect just change it back to what it was and see the Unit test fail.

pkg/northbound/gnmi/set.go
SeanCondon: I'd suggest <targetName>Running

Issue #174, opened 2019-05-16
Move redundant helper functions to utils package 


adibrastegarnia:
I see some of the helper functions such as  *getCertPoolDefault()** and *getCertPool(CaPath string)* have been repeated in multiple packages. If these functions need be used in multiple packages that would be better to be defined in *utils* package that can be used by other packages without repeating them unless there is a reason to keep them local.

Issue #173, opened 2019-05-16
Refactor code for extracting common command flags


tomikazi:
Extract code to deal with options for server address/port, certificates, keys, etc. into a single function for use by the clients.

tomikazi:
Adib's change addresses this sufficiently.

adibrastegarnia:
> Adib's change addresses this sufficiently.

Sean did :) Thanks for his work :)

PR #180, opened 2019-05-16
Rollback a newtork change Fixes #105

### LINKED ISSUE #105: Create gRPC core service
gNMI does not include all functionality, e.g rollback. We need to establish a lateral gRPC service to allow these.

SeanCondon:


cmd/admin/rollback/rollback.go
```diff
@@ -0,0 +1,74 @@
...
+
+func main() {
+	address := flag.String("address", ":5150", "address to which to send requests")
+	changeName := flag.String("changename", "", "")
+	keyPath := flag.String("keyPath", "", "path to client private key")
+	certPath := flag.String("certPath", "", "path to client certificate")
+	flag.Parse()
+
+	var opts = []grpc.DialOption{}
+	if *keyPath != "" && *certPath != "" {
```
Andrea-Campanella: we ought to extract lines 43 to 60. using them a bit all over the place

pkg/manager/getconfig.go
```diff
@@ -28,11 +28,24 @@ func (m *Manager) GetNetworkConfig(target string, configname string, path string
...
+		for _, cfg := range m.ConfigStore.Store {
+			if cfg.Device == target {
+				config = cfg
+				break
+			}
+		}
+		if config.Name == "" {
+			return make([]change.ConfigValue, 0),
+				fmt.Errorf("No Configuration found for %s", target)
+		}
```
Andrea-Campanella: extract this if and use it at line 44 with else clause

tomikazi:
Yes, there is an issue for this to be refactored. 

Thomas

cmd/admin/rollback/rollback.go
SeanCondon: This is targeted by Adib with #173 and #174

pkg/manager/getconfig.go
SeanCondon: Currently we're only using the target name to find the Config name. I was leaving this here in case we use it differently in future.
I've added in the else stmt

Issue #172, opened 2019-05-16
Pass a Context explicitly to each function - use receiver type functions


adibrastegarnia:
I think gnmi southbound client manager needs the following changes: 

1. According to the Go documents: Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. So in the clientmanager implementation the context should be passed to the functions which need it. 

2. We should use the receiver pointer and value functions to avoid sending target as an argument to the functions.

tomikazi:
I'd retitle to be more specific (to give a better idea from the issue's list)

Andrea-Campanella:
@adibrastegarnia I think you had a patch for this on the previous enos-config repo living under opennetworkinglab. Coudl you move it over to hear ? thanks!

adibrastegarnia:
@Andrea-Campanella 
Yes. My plan is to open a pull request soon and you can review it from there.

adibrastegarnia:
closed #183

PR #183, opened 2019-05-16
Pass a context explicitly to each function - use receiver type functi…

adibrastegarnia:
@Andrea-Campanella 
To address issue #172 

I tested with southbound main and I also made changes in synchronizer.

### DIFF
--- a/cmd/southbound_main/southbound_main.go
+++ b/cmd/southbound_main/southbound_main.go
@@ -15,11 +15,13 @@
 package main
 
 import (
+	"context"
 	"fmt"
-	"github.com/onosproject/onos-config/pkg/southbound"
-	"github.com/onosproject/onos-config/pkg/southbound/topocache"
 	"reflect"
 	"time"
+
+	"github.com/onosproject/onos-config/pkg/southbound"
+	"github.com/onosproject/onos-config/pkg/southbound/topocache"
 
 	"github.com/golang/protobuf/proto"
 	"github.com/openconfig/gnmi/client"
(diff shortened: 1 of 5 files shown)


pkg/southbound/defs.go
```diff
@@ -0,0 +1,76 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package southbound
+
+import (
+	"github.com/openconfig/gnmi/client"
```
Andrea-Campanella: is this needed ?

pkg/southbound/clientManager.go
```diff
@@ -233,39 +205,39 @@ func Subscribe(target Target, request *gpb.SubscribeRequest, handler client.Noti
 	q.TLS = target.Destination.TLS
 	q.NotificationHandler = handler
 	c := client.New()
-	err = c.Subscribe(target.Ctxt, q, "")
```
Andrea-Campanella: why do you need to pass the ctx and can't use the one saved in the target ? Same in all other methods.

pkg/southbound/clientManager.go
```diff
@@ -153,26 +125,26 @@ func getCertPoolDefault() *x509.CertPool {
 }
 
 // CapabilitiesWithString allows a request for the capabilities by a string - can be empty
-func CapabilitiesWithString(target Target, request string) (*gpb.CapabilityResponse, error) {
+func (target Target) CapabilitiesWithString(ctx context.Context, request string) (*gpb.CapabilityResponse, error) {
```
Andrea-Campanella: we should pass here a pointer (target *Target) not to pass the whole object

cmd/southbound_main/southbound_main.go
```diff
@@ -15,47 +15,49 @@
...
 	}
-	target, err := southbound.GetTarget(southbound.Key{Key: device.Addr})
+	ctx := context.Background()
+	target := southbound.Target{}
+
+	_, err := target.ConnectTarget(ctx, device)
 	if err != nil {
-		fmt.Println("Creating device for addr: ", device.Addr)
-		target, _, err = southbound.ConnectTarget(device)
-		if err != nil {
```
Andrea-Campanella: this if err != nil should be left there

pkg/southbound/defs.go
adibrastegarnia: Yes. For the target struct is needed. As you may noticed, I separated definitions from the actual code to make code more readable.

pkg/southbound/clientManager.go
adibrastegarnia: I followed Go rules for using context in a struct vs passing it as an argument to the functions. Here (https://golang.org/pkg/context/) recommends to pass it as an argument. If you search the web, you will find discussions around this topic and some of the potential problems that can be raised if define it in a struct. I read somewhere if we pass it using channels between processes then it is OK to define in it in the struct. Northbound interface also followed the same paradigm (i.e. passing it as an argument). Personally, I would prefer to have abstractions to minimize number of arguments that a user needs to provide to a function but I think it would be better to follow the rules.

pkg/southbound/clientManager.go
adibrastegarnia: We don't change the target fields in this function. Why do we need to define it as a pointer? I defined as a pointer in ConnectTarget since we change the fields. Am I missing something?

Andrea-Campanella: I thought that always a pointer would be better than a full object even if we do not use it since it's more lightweight.

adibrastegarnia: Thanks for your response. That makes sense. Another reason that think of is the consistency. Typically, If we have one function which uses receiver pointer, it is recommended to define the other ones using pointers as well to see a predictable behavior. I will make the changes accordingly.

cmd/southbound_main/southbound_main.go
adibrastegarnia: For which one? I have the err checking in the code for connectTarget but not for GetTarget. Do you mean handling the err for GetTarget?

pkg/southbound/clientManager.go
Andrea-Campanella: ok. is it fine if we read through a structure and always pass that into the method ? is that safe ?

pkg/southbound/defs.go
Andrea-Campanella: +1

pkg/southbound/clientManager.go
adibrastegarnia: Here (https://medium.com/@cep21/how-to-correctly-use-context-context-in-go-1-7-8f2c0fafdf39) explains when that would be OK to use it inside struct. I still need to go thorough that article and some other documents to see how can we apply it to our case. It looks like we need to use channels for communication between functions to make sure that is safe.

adibrastegarnia:
@ray-milkey 

Ray, 

Would you please take a look at travis results? It doesn't happen all the times but It happens sometimes but I don't think it is related to the changes that I made.

ray-milkey:
I see this error occasionally running locally on my Mac. I think @SeanCondon has a fix for it.

ray-milkey:
In the meantime I reran the job and it passed.

adibrastegarnia:
> In the meantime I reran the job and it passed.

Great. Thanks.

pkg/southbound/clientManager.go
adibrastegarnia: After reading a little bit about this topic, I think the current implementation that we pass context as an argument to the functions is the best option at this moment. If you have any ideas, please let me know to make it better.

tomikazi: I agree with Adib. The docs are fairly explicit about passing the context directly. This also makes it clear to the maintainers that a function is part of a request handling chain.

PR #185, opened 2019-05-17
gNMI Subscribe tests for ONCE and Stream

Andrea-Campanella:


configs/configStore-sample.json
```diff
@@ -3,8 +3,8 @@
   "Storetype": "config",
   "Store": {
     "Device1Version": {
-      "Name": "Device1Version",
-      "Device": "Device1",
```
SeanCondon: I didn't realize you had rename this - can we keep the Device1 and add on localhost as an extra config

I think it's valuable to have Device1 and Device2 configs as well as other

deployments/helm/onos-config/files/configs/changeStore-sample.json
```diff
@@ -32,6 +32,11 @@
           "Value": "abc",
           "Remove": false
         },
+        {
```
SeanCondon: This is a change on the deployments change store - did you mean to change this and or the ./configs change store?

PR #188, opened 2019-05-18
move images to its folder to organize them

adibrastegarnia:
Move images to a separate folder named "images" to organize them.


tomikazi:
Looks like an independent unit test failure.

Issue #189, opened 2019-05-19
Use gnmi generated getters to access the fields in a struct 


adibrastegarnia:
It would be better to use gnmi getters to access the fields of a struct because gNMI generated getters make sure that the receiver is not nil.

adibrastegarnia:
Closed in #190

PR #190, opened 2019-05-19
Use gnmi generated getters to access the fields in a struct (#189)- m…

adibrastegarnia:
This pull request is for:

1- To address the issue #189 
2- To Address a TODO in the set function by consolidating lines into methods.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -3,237 +3,97 @@
 
 [[projects]]
   branch = "master"
-  digest = "1:1ba1d79f2810270045c328ae5d674321db34e3aae468eb4233883b473c5c0467"
   name = "github.com/golang/glog"
   packages = ["."]
-  pruneopts = "UT"
   revision = "23def4e6c14b4da8ac2ed8007337bc5eb5007998"
 
 [[projects]]
   branch = "master"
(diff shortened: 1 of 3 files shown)


stolsma:
Just for mijself, to learn something because I'm new to Golang: what is the gookg.lock file doing and why the (for me, but maybe its important) cosmetic change from arrays of lines to oneliners for filenames throughout the file?

tomikazi:
It tracks changes in the external package dependencies. It is generated/updated by Go dep. 

Thomas

stolsma:
Ah, thanks Thomas!

adibrastegarnia:
> It would be good to add some documentation to the new methods as you create them. I'm not convinced this is productive breaking out in to functions that are only ever used once.

@SeanCondon 
Thanks for your comment. I am totally OK with adding comments to the new methods but it looks like we don't add comments to the functions that have not been exposed publicly and used for internal purposes inside a package. I didn't want to change the pattern if that is the case. Do you mean add comments in the body or somewhere else?

SeanCondon:
I think we are free to add comments anywhere they help readability. The reason I thought comments might help here is it was not clear to me why this needed refactoring

PR #193, opened 2019-05-20
Subscribe gNMI NB: Filtering for channels to send the update

Andrea-Campanella:


pkg/northbound/gnmi/subscribe.go
```diff
@@ -129,38 +129,35 @@ func broadcastNotification() {
...
-		//TODO use proper type of value, re-use code in get
-		typedValue := gnmi.TypedValue_StringVal{StringVal: changeValue.Value}
-		value := &gnmi.TypedValue{Value: &typedValue}
-		updatePath := &gnmi.Path{Elem: pathElemsRefs.Elem, Target: target}
-		update := &gnmi.Update{
-			Path: updatePath,
-			Val:  value,
-		}
-		ch <- update
+func sendUpdate(path *gnmi.Path, value string, ch chan *gnmi.Update) {
```
SeanCondon: I think you can make this safer by saying we're only outputting to this channel by making it
ch chan<- *gnmi.Update

pkg/northbound/gnmi/subscribe.go
```diff
@@ -129,38 +129,35 @@ func broadcastNotification() {
 		log.Println("Error while subscribing to updates", err)
 	}
 	for update := range changes {
```
SeanCondon: I think it would be more readable if "changes" were "changesChan". I think all channel variables should end in Chan, so you know what you're dealing with

pkg/northbound/gnmi/subscribe.go
```diff
@@ -129,38 +129,35 @@ func broadcastNotification() {
...
-			changeID := (*values)[events.ChangeID]
-			changeInternal := mgr.ChangeStore.Store[changeID]
-			err := sendUpdate(target, changeInternal, ch)
-			if err != nil {
-				log.Println("Error while parsing path ", err)
+		target, changeInternal := getChange(update, mgr)
+		//For every channel we cycle over the paths of the config change and if somebody is subscribed to it we send out
+		for path, ch := range PathToChannels {
+			for _, changeValue := range changeInternal.Config {
+				//FIXME this might prove expensive, find better way to store path and targer in channels map
```
SeanCondon: I don't mind this - you're going between 2 different paradigms for dealing with paths

pkg/northbound/gnmi/subscribe.go
```diff
@@ -129,38 +129,35 @@ func broadcastNotification() {
...
+				strPath := utils.StrPath(path)
+				if path.Target == target && strPath == changeValue.Path {
```
SeanCondon: I'm wondering about the "strPath == changeValue.Path". Should you be more inclusive than an exact match - if I gave a short path in the subscribe, I'd expect to see everything below it be a match. This is necessary to get notifications of the creation of children

pkg/northbound/gnmi/subscribe.go
Andrea-Campanella: done

pkg/northbound/gnmi/subscribe.go
Andrea-Campanella: done

PR #196, opened 2019-05-20
Unit tests for SB GNMI handling

ray-milkey:
This is a work in progress to experiment with stubbing out interfaces to external libraries for testing.


pkg/southbound/gnmiClient.go
```diff
@@ -0,0 +1,94 @@
...
+	Subscribe(ctx context.Context, q client.Query) error
+	Recv() error
+	Close() error
+	Poll() error
+}
+
+// GnmiClientImpl :
+type GnmiClientImpl struct {
+	c            client.Impl
+	get          func(ctx context.Context, r *gpb.GetRequest) (*gpb.GetResponse, error)
```
SeanCondon: It's a nifty design alright. I see the advantage for substitution during testing. I'm not a fan of functions as variables in general (just an opinion). If there are other advantages to this (other than test) then it might be worthwhile going this direction.

ray-milkey: I think those function pointers might be completely redundant now - I'll play around a little with refactoring it.

ray-milkey: I'll push a change soon that removes these function pointers. Good catch! The only place that I absolutely need function pointers is in clientManager.go where it has a factory method to allocate the clients. Tests have to be able to override that from the outside. I'm going to refactor that code too, to make it clearer what is going on.

ray-milkey: I've also moved the factory function pointers out of clientManager into the individual client shim classes. This means less pollution for the manager implementation class.

pkg/southbound/gnmiClient.go
```diff
@@ -0,0 +1,63 @@
...
+	}, err
+}
+
+// GnmiClientImpl : Default implementation of GnmiClient based on the openconfig GNMI client.
+type gnmiClientImpl struct {
+	c client.Impl
+}
+
+// Capabilities : GNMI capabilities
+func (client gnmiClientImpl) Capabilities(ctx context.Context, r *gpb.CapabilityRequest) (*gpb.CapabilityResponse, error) {
```
tomikazi: I must be missing something... is this a recursing function?

ray-milkey:
I fixed the client recursion problem Thomas found - still trying to figure out how to test the real southbound

ray-milkey:
I was able to get gnmi commands into onos-config and down to a simulated device. This should be ready to go now.

PR #200, opened 2019-05-21
adding scripts to automate running gnmi_cli commands and config manag…

adibrastegarnia:
The idea is to add some scripts to avoid copying and pasting gnmi cli commands and some other commands on terminal for testing purposes. Please let me know if you have any comments.

### DIFF
--- a/tools/test/scripts/admin_tools/net-changes.sh
+++ b/tools/test/scripts/admin_tools/net-changes.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+# list all network changes submitted through the northbound gNMI interface run'
+
+ONOS_CONFIG_ROOT=$HOME/go/src/github.com/onosproject/onos-config
+
+go run github.com/onosproject/onos-config/cmd/admin/net-changes \
+    -certPath $ONOS_CONFIG_ROOT/tools/test/devicesim/certs/client1.crt \
+    -keyPath $ONOS_CONFIG_ROOT/tools/test/devicesim/certs/client1.key
--- a/tools/test/scripts/nb_test/get.sh
+++ b/tools/test/scripts/nb_test/get.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+# Make a gNMI Get request
+
+# Path to onos-config root directory
+ONOS_CONFIG_ROOT=$HOME/go/src/github.com/onosproject/onos-config
+
+# The request message
+PROTO="path: <target: 'localhost:10161', elem: <name: 'openconfig-system:system'> elem:<name:'config'> elem: <name: 'motd-banner'>>"
+
+# Time out value
+TIME_OUT=5s
+
+gnmi_cli -get -address localhost:5150 \
+    -proto "$PROTO"  \
+    -timeout $TIME_OUT  \
+    -client_crt $ONOS_CONFIG_ROOT/tools/test/devicesim/certs/client1.crt \
+    -client_key $ONOS_CONFIG_ROOT/tools/test/devicesim/certs/client1.key \
+    -ca_crt $ONOS_CONFIG_ROOT/tools/test/devicesim/certs/onfca.crt \
+    -alsologtostderr 
--- a/tools/test/scripts/nb_test/getdevices.sh
+++ b/tools/test/scripts/nb_test/getdevices.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+# Retrieve list of all stored device names
+
+# Path to onos-config root directory
+ONOS_CONFIG_ROOT=$HOME/go/src/github.com/onosproject/onos-config
+
+# The request message
+PROTO="path: <target: '*'>" 
+
+# Time out value
+TIME_OUT=5s
+
+gnmi_cli -get -address localhost:5150 \
+    -proto "$PROTO" \
+    -timeout $TIME_OUT  \
+    -client_crt $ONOS_CONFIG_ROOT/tools/test/devicesim/certs/client1.crt \
+    -client_key $ONOS_CONFIG_ROOT/tools/test/devicesim/certs/client1.key \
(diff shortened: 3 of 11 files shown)


tools/test/scripts/nb_test/set.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+ONOS_CONFIG_ROOT=$HOME/go/src/github.com/onosproject/onos-config
+
+gnmi_cli -address localhost:5150 -set \
+    -proto "update: <path: <target: 'localhost:10161', elem: <name: 'system'> elem: <name: 'clock' > elem: <name: 'config'> elem: <name: 'timezone-name'>> val: <string_val: 'Europe/Dublin'>>" \
```
SeanCondon: the problem is the proto - were you going to hard code it like this? It will always update the timezone, and so maybe the script should have that as a name

tools/test/scripts/onos-config.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+ONOS_CONFIG_ROOT=$HOME/go/src/github.com/onosproject/onos-config
+
+go run github.com/onosproject/onos-config/cmd/onos-config-manager \
+    -configStore=$ONOS_CONFIG_ROOT/configs/configStore-sample.json \
+    -changeStore=$ONOS_CONFIG_ROOT/configs/changeStore-sample.json \
+    -deviceStore=$ONOS_CONFIG_ROOT/configs/deviceStore-sample.json \
+    -networkStore=$ONOS_CONFIG_ROOT/configs/networkStore-sample.json \
+    -caPath $ONOS_CONFIG_ROOT/tools/test/devicesim/certs/onfca.crt \
```
SeanCondon: see updated run.md - no need for caPath, keyPath, certPath now

adibrastegarnia: Sure. I will update it accordingly.

tools/test/scripts/nb_test/set.sh
adibrastegarnia: I was thinking to define it as a variable on top of the script that can be changed easily. I assume that the -proto argument can be changed by the user. Is that what you are suggesting?

adibrastegarnia: @SeanCondon 
I made the changes and introduced some variables to make the scripts more general. I couldn't test subscribe because it doesn't work and we need to take a look at it later.

Andrea-Campanella: I know I might be a bit of a pain here. I would not extract these in other commands. gnome_cli it's an external tool from ONOS-config so I woudl not tie us to it in any way. Also to me the cli is a pure testing tool so c/p is not an issue and actually enables me to learn more about what's going on If I see the whole command

tools/test/scripts/nb_test/get.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
```
Andrea-Campanella: I'd remove this

tools/test/scripts/nb_test/set.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
```
Andrea-Campanella: I'd remove this

tools/test/scripts/nb_test/subscribe.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
```
Andrea-Campanella: I'd remove this

tools/test/scripts/nb_test/getdevices.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
```
Andrea-Campanella: I'd remove this

tools/test/scripts/admin_tools/net-changes.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# list all network changes submitted through the northbound gNMI interface run'
+
+ONOS_CONFIG_ROOT=$HOME/go/src/github.com/onosproject/onos-config
+
+go run github.com/onosproject/onos-config/cmd/admin/net-changes \
```
SeanCondon: With this and all the other utils there is now no need to give keyPath and certPath - see updates in run.md. 
Not much left to script here

tools/test/scripts/nb_test/subscribe.sh
SeanCondon: @Andrea-Campanella Did you mean the shebang or the whole script?
I agree with removing the whole script - not much to script and besides it can't handle arguments for rollback etc

Andrea-Campanella: the whole thing, did not find a "file" comment

tools/test/scripts/admin_tools/net-changes.sh
adibrastegarnia: @SeanCondon @Andrea-Campanella 
I agree with what you said and I think with 
new changes no scripts will be needed  
for internal commands. In my opinion, in the long term if we don’t 
want to rely on gnmi cli, we need to equip the onos-config to 
a real CLI (not just a binary which accepts arguments) that we can 
execute different commands on top of NB. I think we can close this pull request.

SeanCondon: > @SeanCondon @Andrea-Campanella
> I agree with what you said and I think with
> new changes no scripts will be needed
> for internal commands. In my opinion, in the long term if we don’t
> want to rely on gnmi cli, we need to equip the onos-config to
> a real CLI (not just a binary which accepts arguments) that we can
> execute different commands on top of NB. I think we can close this pull request.

Thanks Adib. I think this is the best for the moment. Even with "go run .../onos-config-manager" the list of stores will soon disappear as they will be transitioned to Atomix

PR #201, opened 2019-05-21
Apply gofmt rules

ray-milkey:


SeanCondon:
Hi Ray - I don't know how I could have committed this code, and it not be caught by Travis, and then it creates a problem for you. Is there something I'm not understanding about *gofmt*

SeanCondon:
Never mind - I see the changes you made that caused this

PR #205, opened 2019-05-22
Move redundant helper functions to utils package Fixes #174

### LINKED ISSUE #174: Move redundant helper functions to utils package

SeanCondon:


docs/run.md
```diff
@@ -144,19 +144,15 @@ functionality.
 ### List Network Changes
```
Andrea-Campanella: maybe worth putting in a comment saying that if you use defaults certs these path are ok, otherwise please use --certPath and --keyPath flags.

PR #209, opened 2019-05-22
Move docker image for device simulator under the onosproject repo

ray-milkey:
Not tested yet.


ray-milkey:
This is ready to go in now

Issue #176, opened 2019-05-16
Minor suggestions


damianoneill:
See https://goreportcard.com/report/github.com/onosproject/onos-config for a few minor suggestions.

SeanCondon:
gofmt -s has now been integrated in to the Travis CI build and the Makefile script
Thanks for your input
We will keep an occasional eye on this goreportcard

damianoneill:
@SeanCondon thanks for the update, might be worth considering adding https://github.com/golangci/golangci-lint into the lint target, similar functionality as gometalinter, but significantly faster.  

```sh
$ golangci-lint run ./...
pkg/northbound/server.go:41:2: `mu` is unused (structcheck)
	mu       sync.Mutex
	^
pkg/store/change/change_test.go:29:11: `change2` is unused (deadcode)
	change1, change2, change3, change4 *Change
	         ^
cmd/profiling/profile-config.go:43:23: Error return value of `pprof.StartCPUProfile` is not checked (errcheck)
	pprof.StartCPUProfile(cpuFile)
	                     ^
cmd/southbound_main/southbound_main.go:94:18: Error return value of `target.Subscribe` is not checked (errcheck)
	target.Subscribe(ctx, req, handler)
	                ^
pkg/listener/listener_test.go:129:12: Error return value of `Unregister` is not checked (errcheck)
	Unregister(device1.Addr, true)
	          ^
pkg/listener/listener_test.go:130:12: Error return value of `Unregister` is not checked (errcheck)
	Unregister(device2.Addr, true)
	          ^
pkg/listener/listener_test.go:131:12: Error return value of `Unregister` is not checked (errcheck)
	Unregister(device3.Addr, true)
	          ^
pkg/manager/tree.go:81:16: Error return value of `addPathToTree` is not checked (errcheck)
		addPathToTree(cv.Path, cv.Value, &root)
		             ^
pkg/northbound/admin/admin.go:127:47: Error return value of `.RemoveEntry` is not checked (errcheck)
	manager.GetManager().NetworkStore.RemoveEntry(networkConfig.Name)
	                                             ^
pkg/southbound/topocache/topocache.go:58:20: Error return value of `jsonDecoder.Decode` is not checked (errcheck)
	jsonDecoder.Decode(&deviceStore)
	                  ^
pkg/store/store-api_test.go:416:11: Error return value of `os.Mkdir` is not checked (errcheck)
		os.Mkdir("testout", os.ModePerm)
		        ^
[...]

PR #213, opened 2019-05-23
#176 badges for goreport and travis build

damianoneill:


Andrea-Campanella:
Thanks for the small but great update! Look forward to more!

Issue #212, opened 2019-05-23
devicesim using GNXI project's gnmi_target does not support Subscribe


SeanCondon:
See https://github.com/google/gnxi/blob/master/gnmi/server.go line 623

Also when I try calling Subscribe I do not get this error message. Use the following to see it in action

```
gnmi_cli -address localhost:10161 \
-proto "subscribe:<mode: 0, prefix:<>, subscription:<path: <elem: <name: 'openconfig-system:system'> elem: <name: 'clock' > elem: <name: 'config'> elem: <name: 'timezone-name'>>>>" \
-timeout 5s -alsologtostderr \
-client_crt certs/client1.crt \
-client_key certs/client1.key \
-ca_crt certs/onfca.crt
```

adibrastegarnia:
We can add it in gnmi target code and overwrites. I will talk about it today.

adibrastegarnia:
@SeanCondon @tomikazi @ray-milkey @Andrea-Campanella @kuujo 

The changes that we need to made at this point are:

1- We need to return an error message to the client from gnmi target
2- We need to support subscribe. 

By looking at the code, *gnmi_target.go* which implement the target uses the *server.go* under gnmi directory in gnxi tool github repository and they overwrite Get and Set functions to provide user authentication. That means for now we just need to have gnmi_target.go in onos-config repo to support the features that we are looking for without touching the *server.go* code. 

To do that, we need to create a directory called third_party under onos-config to have gnmi_target package there. We will change Docker file later to fetch it from our repo.

SeanCondon:
Adib - I think you can put it in tools/test/devicesim/gnmi_target.
I wouldn't call it third_party because we're copying the code once off with no further reference to gnxi. We did something similar with utils from Arista

adibrastegarnia:
@SeanCondon 
OK. I will do that and open a initial pull request soon.

Andrea-Campanella:
@adibrastegarnia move this issue to the simulator repo and then close

adibrastegarnia:
@SeanCondon  @Andrea-Campanella 

Please close this issue. I cannot close it.

PR #226, opened 2019-05-23
Implement subscribe ONCE for devsim - issue #212

adibrastegarnia:
This pull request is for  supporting subscribe ONCE in the device simulator.

### DIFF
--- a/tools/test/devicesim/gnmi_user_manual.md
+++ b/tools/test/devicesim/gnmi_user_manual.md
@@ -103,13 +103,26 @@ response: <
 ```
 
 ## Run the Subscribe command
+### Subscribe ONCE
 ```bash
 gnmi_cli -address localhost:10161 \
-       -proto "subscribe:<mode: 0, prefix:<>, subscription:<path: <elem: <name: 'openconfig-system:system'> elem: <name: 'clock' > elem: <name: 'config'> elem: <name: 'timezone-name'>>>>" \
+       -proto "subscribe:<mode: 1, prefix:<>, subscription:<path: <elem: <name: 'openconfig-system:system'>  elem: <name: 'clock' > elem: <name: 'config'> elem: <name: 'timezone-name'>>>>" \
        -timeout 5s -alsologtostderr \
        -client_crt certs/client1.crt -client_key certs/client1.key -ca_crt certs/onfca.crt
 ```
-> **Note: this returns ```rpc error: code = Unimplemented desc = Subscribe is not implemented.``` and is the subject of Issue #212**
+
+This gives a response like this. 
+```bash
+{
+  "system": {
+    "clock": {
+      "config": {
+        "timezone-name": "Europe/Dublin"
+      }
+    }
+  }
+}
+```
 
 ## Troubleshooting
 
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -252,6 +252,7 @@
     "github.com/openconfig/gnmi/client/gnmi",
     "github.com/openconfig/gnmi/proto/gnmi",
     "github.com/openconfig/gnmi/proto/gnmi_ext",
+    "github.com/openconfig/ygot/ygot",
     "golang.org/x/net/context",
     "google.golang.org/grpc",
     "google.golang.org/grpc/codes",
--- a/tools/test/devicesim/Dockerfile
+++ b/tools/test/devicesim/Dockerfile
@@ -39,7 +39,6 @@ RUN mkdir -p $GOPATH \
       github.com/google/gnxi/gnmi_capabilities \
       github.com/google/gnxi/gnmi_get \
       github.com/google/gnxi/gnmi_set \
-      github.com/onosproject/onos-config/tools/test/devicesim/gnmi_target \
       github.com/openconfig/gnmi/cmd/gnmi_cli \
       github.com/google/gnxi/gnoi_target \ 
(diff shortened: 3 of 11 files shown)


adibrastegarnia:
@SeanCondon 

I ran into an issue when I build *gnmi_target.go* under *tools/devsim*. It gives me the following error: 
```
# github.com/onosproject/onos-config/vendor/github.com/google/gnxi/gnmi
../../../../vendor/github.com/google/gnxi/gnmi/model.go:34:53: undefined: ytypes.UnmarshalOpt
# github.com/onosproject/onos-config/vendor/github.com/google/gnxi/gnmi/modeldata/gostruct
../../../../vendor/github.com/google/gnxi/gnmi/modeldata/gostruct/generated.go:57:63: undefined: ytypes.UnmarshalOpt
../../../../vendor/github.com/google/gnxi/gnmi/modeldata/gostruct/generated.go:67:25: too many arguments in call to ytypes.Unmarshal
```

I can build it without any problems under go/src. That's strange. 

P.S. It looks like Makefile doesn't cover tools for build. So for now whenever you get a chance just run go build and see can you build it.

SeanCondon:
@adibrastegarnia 
I get the same error when I do:
```
go run github.com/onosproject/onos-config/tools/test/devicesim/gnmi_target
```
I don't know why it looks to this vendor folder. I think it's OK that the make does not build it. It's the Dockerfile that should build it IMHO.

I think in the run_targets.sh line 17 & 47 you should change this to "go run ..." like above

Also I think you should remove the **gnmi_target** from *go get* and *go install* in the Dockerfile, and rebuild the image

When I do this it **seems** to run for me.

Also on reflection I think it might be worth renaming yours to **gnmi_devicesim/gnmi_devicesim.go** to prevent any clashes if both of them are installed.

adibrastegarnia:
@SeanCondon 

Yes. That is strange. For now, I am developing the code under src that I can debug it easily by running just one target. Thanks for your comments.

adibrastegarnia:
@SeanCondon 

I implemented subscribe ONCE for the devsim and tested with one target and it works fine. I almost followed the same approach which is used in NB. I would prefer to merge this commit after your review and I will open separate pull requests for subscribe STREAM and POLL to make the review easier. We will keep the issue open until I implement all of the  subscribe modes.

tools/test/devicesim/gnmi_user_manual.md
```diff
@@ -103,12 +103,26 @@ response: <
...
+  "system": {
+    "clock": {
+      "config": {
+        "timezone-name": "Europe/Dublin"
+      }
+    }
+  }
+}
+```
 > **Note: this returns ```rpc error: code = Unimplemented desc = Subscribe is not implemented.``` and is the subject of Issue #212**
```
SeanCondon: This can go if the subscribe returns something correctly (maybe it's still relevant for POLL and STREAM)

tools/test/devicesim/gnmi_user_manual.md
```diff
@@ -103,12 +103,26 @@ response: <
 ```
 
 ## Run the Subscribe command
+### Subscribe ONCE
 ```bash
 gnmi_cli -address localhost:10161 \
-       -proto "subscribe:<mode: 0, prefix:<>, subscription:<path: <elem: <name: 'openconfig-system:system'> elem: <name: 'clock' > elem: <name: 'config'> elem: <name: 'timezone-name'>>>>" \
```
SeanCondon: With or without the namespace should work

adibrastegarnia:
> Adib - it doesn't work for me - I did a new build of the Docker image and still I get
> 
> ```
> rpc error: code = Unimplemented desc = Subscribe is not implemented.
> ```

@SeanCondon 
Thanks for your comments. You shouldn't get that error if you build the docker based on new file. The reason is that when we overwrite a function the original one won't be executed independent of what we have in the body. So your image still runs the previous code. What I did to make sure that I can build the project, I developed the code under GOPATH that I can build it without changing the Dockerfile. Would you please copy the whole directory (gnmi_target) to your GOPATH and run go build and then run the binary and see if it still you get the same error?

tools/test/devicesim/gnmi_user_manual.md
adibrastegarnia: @SeanCondon 
Yes it is gone I forgot to remove it from the manual. I will take care of it.

adibrastegarnia: Yes. It is gone but I forgot to remove it from the user manual

tools/test/devicesim/gnmi_user_manual.md
adibrastegarnia: Yes I know but it looks like when I remove it gnmi_cli gives me a path error (I don't think it is related to the code that I added). Let me take a closer look and then update you.

adibrastegarnia: I tested again with namespace and it works fine. Perhaps I had a typo when I was testing before. I will return the command to the original one.

SeanCondon:
Adib
I think the problem is that in the Dockerfile when it does 
```
go get -u
...
github.com/onosproject/onos-config/tools/test/devicesim/gnmi_target
```
it is actually pulling your first version of this from the Web and not the version in the current branch.
Can you change this command to get it from the current branch? You might have to have a step that copies the code over and then does a "go build/install"

PR #236, opened 2019-05-27
Fix ineffassignments reported by goreportcard

adibrastegarnia:


pkg/northbound/gnmi/get_test.go
```diff
@@ -42,6 +43,7 @@ func Test_getWithPrefixNoOtherPathsNoTarget(t *testing.T) {
 	server := setUp(false)
 
 	prefixPath, err := utils.ParseGNMIElements([]string{"test1:cont1a", "cont2a"})
+	assert.NilError(t, err, "Unexpected error")
```
SeanCondon: I think you can drop the comment on assert.Equals

adibrastegarnia: @SeanCondon 
I assume that you mean the category of the assert functions which behaves like equal because the one you commented in assert.NilError.

adibrastegarnia: @ I fixed some others as well in the same file.

PR #242, opened 2019-05-28
Added model and version types to Configuration. Fixes #230

### LINKED ISSUE #230: Need to store model data per DeviceType in onos-config
To support the validation of configuration against YANG models there needs to be a collection of Model data per device type somewhere in onos-config

## Proposal
My proposal is that the DeviceType would be an immutable object containing a sorted set of gnmi.ModelData entries. This would be stored in a store just like Change, Configuration and Network

Each Device would have a name (a tag and a version combined) and an ID (a hash of the model data). Names might be like "Stratum-1.0" or "MyDevice-1.2.3"

The existing Configuration object would be extended to point to a DeviceType.

During connection to the device we could check that the capabilities match the DeviceType preventing us sending down incompatible config to it.

On the NBI this will help validating paths against specific targets. 

It also helps in dealing with multiple versions of devices.

There is a lot more to this, which I will add later, but this is to get the ball rolling

## Alternative
An alternative approach is to extend the Configuration object with an attribute Model []gnmi.Model, so that each configuration holds its own set of Model data

SeanCondon:
Extended the Configuration object with
* Version
* DeviceType
* ModelData

removed
* User
* Description


pkg/manager/getconfig.go
```diff
@@ -65,7 +65,7 @@ func (m *Manager) GetAllDeviceIds() *[]string {
 	var deviceIds = make([]string, 0)
 
 	for _, v := range m.ConfigStore.Store {
-		deviceIds = append(deviceIds, v.Device)
+		deviceIds = append(deviceIds, v.Device+" ("+v.Version+")")
```
Andrea-Campanella: is worth extracting the config id to a class that has the to string returning this ? I think it will avoid stupid mistakes and render the code easier to run

pkg/northbound/gnmi/extensions.go
```diff
@@ -0,0 +1,27 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package gnmi
+
+const (
+	// GnmiExtensionNetwkChangeID is the extension number used in SetRequest and SetResponse
```
Andrea-Campanella: we should have a document in this patch to specify what these flags are for and especially what happens if you do not specify them.

pkg/manager/getconfig.go
SeanCondon: yeah in this particular case I had made it in to a LeafList type in the response, and was reluctant to make it any more complex
I guess I could convert in to a container with leaves (edited) 
It's only this corner case where we're getting target="*"

pkg/northbound/gnmi/extensions.go
SeanCondon: I think there will have to be a table of what they mean in run.md - I was going to prepare it for the TST tomorrow, so it will be good prep

pkg/store/configuration.go
```diff
@@ -15,25 +15,32 @@
...
 	"github.com/onosproject/onos-config/pkg/store/change"
+	"github.com/openconfig/gnmi/proto/gnmi"
+	"regexp"
 	"sort"
 	"strings"
 	"time"
 )
 
+const configurationNamePattern = `[a-zA-Z0-9\-:_]{4,40}`
+const configurationVersionPattern = `[a-zA-Z0-9_\.]{2,10}`
```
SeanCondon: are we happy with these name restrictions for version and devicetype (max 10 chars)

tomikazi: I think 10 for version is good and 40 for name should be fine.

pkg/northbound/gnmi/extensions.go
Andrea-Campanella: +1

pkg/store/configuration.go
Andrea-Campanella: +1

pkg/manager/getconfig.go
Andrea-Campanella: ok

PR #248, opened 2019-05-29
Implement subscribe poll for the devicesim

adibrastegarnia:
1- The devsim supports subscribe poll as well. 
2- Porting the gnmi server code to the repo that we add the new features that we need.


adibrastegarnia:
> When I ran this - it gives the POLL response every 5 seconds showing Dublin, but when went and changed the TZ value in another terminal to Paris, the POLL event continued giving the old Dublin value. I'd expect it to start transmitting the updated value. Even after restarting the POLL it still displays the old value "Dublin"

Thanks for catching the error. I will take a look to see what is going on.

adibrastegarnia:
> When I ran this - it gives the POLL response every 5 seconds showing Dublin, but when went and changed the TZ value in another terminal to Paris, the POLL event continued giving the old Dublin value. I'd expect it to start transmitting the updated value. Even after restarting the POLL it still displays the old value "Dublin"

@SeanCondon 

I addressed your comment and I think this part is ready to merge. Whenever you get a chance, please  test it and see will you get the expected output for subscribe poll and once. Thanks

tools/test/devicesim/Dockerfile
```diff
@@ -52,17 +52,24 @@ RUN go install -v \
...
+
+COPY target_configs target_configs
+COPY certs certs
+COPY scripts scripts
+COPY gnmi_target gnmi_target
+COPY gnmi $GOPATH/src/github.com/onosproject/onos-config/tools/test/devicesim/gnmi
 
 
+RUN cd $GOPATH/src/github.com/onosproject/onos-config/tools/test/devicesim/gnmi && go install
 RUN cd ./gnmi_target && go install
```
SeanCondon: This gives me a 
```
# _/home/devicesim/gnmi_target
./gnmi_utils.go:115:1: syntax error: non-declaration statement outside function body
```

PR #249, opened 2019-05-29
Add ability to define a name when doing gNMI Set on northbound

SeanCondon:


pkg/store/networkconfiguration.go
```diff
@@ -33,51 +31,23 @@ type NetworkConfiguration struct {
 	ConfigurationChanges map[ConfigName]change.ID
 }
 
-// CreateNetworkStoreWithName creates a NetworkConfiguration object
-func CreateNetworkStoreWithName(name string, user string) (*NetworkConfiguration, error) {
+// CreateNetworkStore creates a NetworkConfiguration object
+// If name is left blank then a name will be generated by a hash algorithm
+func CreateNetworkStore(name string, user string, changes map[ConfigName]change.ID) (*NetworkConfiguration, error) {
```
Andrea-Campanella: I'm thinking that it might be best to rename this to NetworkConfigurationStore

pkg/northbound/gnmi/set.go
```diff
@@ -129,6 +129,7 @@ func (s *Server) buildUpdateResults(targetUpdates mapTargetUpdates,
 func (s *Server) Set(ctx context.Context, req *gnmi.SetRequest) (*gnmi.SetResponse, error) {
 	// There is only one set of extensions in Set request, regardless of number of
 	// updates
+	var netwchname string // May be specified as 100 in extension
```
Andrea-Campanella: name is not very clear. Maybe netchgname to avoid confusion with ch = channel ?

pkg/store/networkconfiguration.go
SeanCondon: Good catch - It's not creating the store either - I've changed it to CreateNetworkConfiguration

PR #251, opened 2019-05-29
Hook up coveralls.io coverage analysis

ray-milkey:


ray-milkey:
I'm not sure if I understand your question. This only works when run from Travis. The script has to be invoked from the Makefile because it needs to run inside the docker container that's doing the build. The coverage files are not accessible outside of the docker.

PR #261, opened 2019-05-30
Adding a make target to run the development docker image locally

teone:
This patch adds:
- `make run-docker` to run the runtime image locally
- `make clean` to remove all the build artifacts

Let me know what you think about it.

Also I noticed that we are passing the entire repository to the docker-context, but that means that any change will force a complete rebuild of the image, while I believe we should take advantage of the docker cache and just load the "./cmd" directory in the context. In that way only changes to the code would wipe the docker cache.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,14 +2,16 @@ export CGO_ENABLED=0
 
 .PHONY: build
 
-ONOS_CONFIG_VERSION := "latest"
-ONOS_BUILD_VERSION := "stable"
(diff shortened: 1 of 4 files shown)


tomikazi:
Awesome change. Incremental builds run MUCH faster for me now.

teone:
This is not leveraging the docker cache yet, I'll take a shot at that in the next few days (SEBA permitting..)

PR #271, opened 2019-05-30
Refactored listener.go to use type struct to avoid global vars

tomikazi:


pkg/listener/listener_test.go
```diff
@@ -82,38 +94,46 @@ func Test_listlisteners(t *testing.T) {
...
 
 func Test_unregister(t *testing.T) {
-	err1 := Unregister("device5", true)
+	d := setUp()
+	err1 := d.Unregister("device5", true)
 
 	assert.Assert(t, err1 != nil, "Unexpected lack of error when unregistering non existent device")
 	assert.Assert(t, is.Contains(err1.Error(), "had not been registered"), "Unexpected error text when unregistering non existent device %s", err1)
 
-	_, err1 = Register("device6", true)
```
ray-milkey: No need for a LHS on these register calls, since you are not checking the results anyway. Go report card flags these as "ineffectual assignments"

pkg/listener/listener_test.go
```diff
@@ -82,38 +94,46 @@ func Test_listlisteners(t *testing.T) {
...
-	_, err1 = Register("device7", true)
+	_, err1 = d.Register("device6", true)
+	_, err1 = d.Register("device7", true)
 
-	err2 := Unregister("device6", true)
```
ray-milkey: Check the value of the "err2" here

pkg/listener/listener.go
```diff
@@ -28,95 +28,99 @@ import (
...
 	}
 	return listenerKeys
 }
 
-// CheckListener allows a check for a name listener
-func CheckListener(name string) bool {
-	for k := range deviceListeners {
+// HasListener returns true if the named listeners has been registered
+func (d *Dispatcher) HasListener(name string) bool {
+	for k := range d.deviceListeners {
```
SeanCondon: since it's a map could do d.deviceListeners[name] here I think

PR #283, opened 2019-05-31
Prevent duplicate names when using extension 100 with SetRequest on gNMI NB

SeanCondon:


tools/test/devicesim/gnmi_target/defs.go
```diff
@@ -17,7 +17,11 @@ package main
 import (
 	"flag"
 
+<<<<<<< HEAD
```
tomikazi: I assume this should not be part of the patch, correct?

tools/test/devicesim/gnmi_target/set.go
```diff
@@ -31,8 +31,12 @@ func (s *server) Set(ctx context.Context, req *pb.SetRequest) (*pb.SetResponse,
 		return nil, status.Error(codes.PermissionDenied, msg)
 	}
 	log.Infof("allowed a Set request: %v", msg)
+<<<<<<< HEAD
```
tomikazi: Same

tools/test/devicesim/gnmi_target/defs.go
SeanCondon: something screwed up in my fork

SeanCondon:
will push again later

PR #293, opened 2019-06-03
Use new device simulator repo name

ray-milkey:


adibrastegarnia:
@ray-milkey 

Should we move the deployment for the simulator to its repo as well?

PR #299, opened 2019-06-04
Adding ability to administratively add/remove/list devices.

tomikazi:


cmd/admin/devices/devices.go
```diff
@@ -0,0 +1,115 @@
...
+
+	} else {
+		listDevices(client, *verbose)
+	}
+}
+
+func addNewDevice(client devices.DeviceInventoryServiceClient, deviceProto string) {
+	deviceInfo := &devices.DeviceInfo{}
+	if err := proto.UnmarshalText(deviceProto, deviceInfo); err == nil {
+		fmt.Printf("Adding device %s\n", deviceInfo.Id)
```
Andrea-Campanella: should this be log ?

cmd/admin/devices/devices.go
```diff
@@ -0,0 +1,115 @@
...
+		if err != nil {
+			log.Fatalf("Unable to add device: %v", err)
+		}
+	} else {
+		log.Fatalf("Unable to parse device info from %s: %v", deviceProto, err)
+	}
+}
+
+func removeDevice(client devices.DeviceInventoryServiceClient, id string) {
+	fmt.Printf("Removing device %s\n", id)
```
Andrea-Campanella: same

docs/admin.md
```diff
@@ -17,6 +17,37 @@ go run github.com/onosproject/onos-config/cmd/admin/net-changes
...
\ No newline at end of file
+```
+
+### Adding, Removing and Listing Devices
+Until the full topology subsystem is available, there is a provisional 
+administrative interface that allows devices to be added, removed and listed via gRPC.
+A command has been provided to allow manipulating the device inventory from the command
+line using this gRPC service.
+
+To add a new device, specify the device informatio protobuf encoding as the value of the
```
Andrea-Campanella: information typo (information)

pkg/southbound/topocache/topocache.go
```diff
@@ -85,3 +81,27 @@ func LoadDeviceStore(file string, topoChannel chan<- events.TopoEvent) (*DeviceS
 
 	return &deviceStore, nil
 }
+
+// AddDevice adds the specified device to the device inventory
+func (store *DeviceStore) AddDevice(id string, device Device) error {
+	err := validateDevice(id, device)
+	if err == nil {
+		store.Store[id] = device
```
SeanCondon: It might be worth checking that this id does not exist already - otherwise the meaning here is AddOrOverwrite

cmd/admin/devices/devices.go
tomikazi: For server-side, everything should just log. For the client commands though, as we discussed, the fmt.* should be used for plain (undecorated) command output and we can continue to use log for error messages.

PR #302, opened 2019-06-05
Added in original configuration for stratum

SeanCondon:
Connects to the Stratum simulator through localhost:50001 in plain mode


SeanCondon:
Waiting on go ahead from Brian to make this stratum config public

PR #311, opened 2019-06-06
Added in an ID to the Device Store (topocache) Fixes #304

### LINKED ISSUE #304: Device name is not yet independent of device address in core
When connecting to a real device through SBI the device name must be the same as device addr in the topocache/deviceStore

What's needed is an id attribute on the Device struct and which should be the same as the value of the key in the Device store

SeanCondon:
I have renamed **localhost:10161** to **localhost-1** etc in all of the stores, to show that the Topocache device ID is not restricted to be an address
The address is only specified in the Addr field of TopoCache

Ray/Jordan I have updated the same file as you did yesterday:
```
deployments/helm/onos-config/templates/configmap.yaml
```
 I think
```
{{ printf "%s:10161" $device | quote }}
```
on lines 17 and 18 (but not 19) can become
```
{{ printf "%s" $device | quote }}
```
please look to see if this will break anything else

### DIFF
--- a/configs/configStore-sample.json
+++ b/configs/configStore-sample.json
@@ -60,9 +60,9 @@
         }
       ]
     },
-    "localhost:10161-1.0.0": {
-      "Name": "localhost:10161-1.0.0",
-      "Device": "localhost:10161",
+    "localhost-1-1.0.0": {
+      "Name": "localhost-1-1.0.0",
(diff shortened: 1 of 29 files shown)


Andrea-Campanella:
@kuujo thoughts ?

PR #313, opened 2019-06-06
Remove TODO from southbound set function

adibrastegarnia:
I think that TODO is not needed anymore unless I am missing something.


adibrastegarnia:
> not my comment originally so I can't say. It seems an extremely trivial thing to be doing a commit for - there are bigger issues to deal with I think

I was looking into subscribe for SB to see how can I test it with simulator and I found this. So I opened a pull request to avoid confusion in the code.

PR #319, opened 2019-06-10
Reimplemented tree.go and added unit tests

SeanCondon:
Also improved 
* memory performance by using []*change.ConfigValue in place of []change.ConfigValue
* stability of tests by changing store-api_test.go to use setUp() rather than TestMain


pkg/store/tree.go
```diff
@@ -0,0 +1,151 @@
+// Copyright 2019-present Open Networking Foundation.
```
Andrea-Campanella: wondering if this class can actually live in the utils. I'm thinking of this because there is no actual relevance to the store and I'll keep that "store" package for things related to atomic

SeanCondon: I think the reason I put it in store is because of the risk of circular dependencies. If I moved it to utils then utils would be dependent on store. And then if I ever needed to use anything from utils inside store I would have a circular dependency. Here I was keeping it close to where it is needed

pkg/store/store-api_test.go
```diff
@@ -187,7 +173,18 @@ var Config2Values = [11]string{
 }
 
 func TestMain(m *testing.M) {
+	//var err error
```
Andrea-Campanella: can this be removed ?

pkg/store/tree.go
Andrea-Campanella: +1 for now, we can move later

pkg/store/tree.go
```diff
@@ -0,0 +1,151 @@
...
+
+	return buf, nil
+}
+
+// addPathToTree is a recursive function that builds up a map
+// suitable for using with json.Marshal, which in turn can be used to feed in to
+// ygot.Unmarshall
+// This follows the approach in https://blog.golang.org/json-and-go "Generic JSON with interface{}"
+func addPathToTree(path, value string, nodeif *interface{}) error {
+	pathelems := strings.Split(path, "/")
```
Andrea-Campanella: we ought to make "/" a constant in order to be able to swap it if necessary

pkg/store/tree.go
```diff
@@ -0,0 +1,151 @@
...
+		if err == nil {
+			(nodemap)[pathelems[1]] = numval
+		} else if strings.ToLower(value) == "true" || strings.ToLower(value) == "false" {
+			(nodemap)[pathelems[1]] = strings.ToLower(value)
+		} else {
+			(nodemap)[pathelems[1]] = value
+		}
+	} else if strings.Contains(pathelems[1], "=") {
+		// To handle list index items
+		refinePath := strings.Join(pathelems[2:], "/")
```
Andrea-Campanella: reuse constant "/"

pkg/store/tree.go
```diff
@@ -0,0 +1,151 @@
...
+			if !ok {
+				return fmt.Errorf("Failed to convert list slice %s", listName)
+			}
+			for idx, ls := range listSliceIf {
+				lsMap, ok := ls.(map[string]interface{})
+				if !ok {
+					return fmt.Errorf("Failed to convert list slice %d", idx)
+				}
+				if (lsMap)[keyName] == keyVal {
+					fmt.Println("Found", keyName, keyVal)
```
Andrea-Campanella: is this just to print ? no assignment ?

pkg/store/store-api_test.go
SeanCondon: Removed the whole TestMain

pkg/store/tree.go
SeanCondon: done

pkg/store/tree.go
SeanCondon: done

pkg/store/tree.go
SeanCondon: This has been fixed - added corresponding new use case with **rx-power**

pkg/store/tree.go
Andrea-Campanella: I don't see it used here in 58

PR #321, opened 2019-06-11
Added in YGOT with a sample YANG file

SeanCondon:
Using the shared library approach - in modelplugin folder
- Build plugin with go build -o build/_output/testdevice.so.1.0.0 -buildmode=plugin ./modelplugin/TestDevice-1.0.0 (from inside its folder)
- Added the model registry to Manager
- Added flag to onos-config to accept 0-many plugins (-modelPlugins=modelplugin/TestDevice-1.0.0/testdevice.so.1.0.0 -modelPlugins=modelplugin/TestDevice-2.0.0/testdevice.so.2.0.0)
- Linked it up to Capabilities (NBI) (removed old pkg/models)

Haven't touched documentation yet - and it plays safe at the moment - if model is not loaded, it breaks nothing. If it is loaded then validation is enforced e.g.
```
gnmi_cli -address localhost:5150 -set     -proto "update: <path: <target: 'Device1', elem: <name: 'test1:cont1a'> elem: <name: 'cont2a'> elem: <name: 'leaf2a'>> val: <string_val: '15'>>" -timeout 5s -alsologtostderr     -client_crt pkg/southbound/testdata/client1.crt     -client_key pkg/southbound/testdata/client1.key     -ca_crt pkg/southbound/testdata/onfca.crt
```
is not allowed because 15 is outside the range for leaf2a


modelplugin/ModelGenerator.sh
```diff
@@ -0,0 +1,82 @@
+#!/bin/bash
+
+TYPEVERSION=TestDevice-2.0.0
```
Andrea-Campanella: I would parametric this script in order for it to be used with any possibile given model/org

pkg/manager/setconfig.go
```diff
@@ -104,9 +104,31 @@ func (m *Manager) SetNetworkConfig(configName store.ConfigName, updates map[stri
 	deviceConfig.Updated = time.Now()
 	m.ConfigStore.Store[configName] = deviceConfig
 
+	modelName := fmt.Sprintf("%s-%s", deviceConfig.Type, deviceConfig.Version)
+	modelPlugin, ok := m.ModelRegistry[modelName]
```
Andrea-Campanella: this modelPlugin is actually the device supported models right ? I would go for a more concrete name, otherwise it seems a random plugin.

pkg/manager/modelregistry.go
```diff
@@ -0,0 +1,79 @@
...
+
+// Capabilities returns an aggregated set of modelData in gNMI capabilities format
+// with duplicates removed
+func (m *Manager) Capabilities() []*gnmi.ModelData {
+	// Make a map - if we get duplicates overwrite them
+	modelMap := make(map[string]*gnmi.ModelData)
+	for _, model := range m.ModelRegistry {
+		_, _, modelItem := model.ModelData()
+		for _, mi := range modelItem {
+			modelName := fmt.Sprintf("%s-%s", mi.Name, mi.Version)
```
Andrea-Campanella: I woudl extract this as a static function somewhere. otherwise we may change the %s-%s and get things broken.

pkg/manager/modelregistry.go
```diff
@@ -0,0 +1,79 @@
...
+type ModelPlugin interface {
+	ModelData() (string, string, []*gnmi.ModelData)
+	UnmarshalConfigValues(jsonTree []byte) (*ygot.ValidatedGoStruct, error)
+	Validate(*ygot.ValidatedGoStruct, ...ygot.ValidationOption) error
+}
+
+// RegisterModelPlugin adds an external model plugin to the model registry at startup
+func (m *Manager) RegisterModelPlugin(moduleName string) error {
+	log.Info("Loading module", moduleName)
+	modelPluginModule, err := plugin.Open(moduleName)
```
Andrea-Campanella: woudl be interesting to break this apart in order to have a REST api where a modelPlugin can be dropped in as a jar and added as information for a device.

modelplugin/ModelGenerator.sh
SeanCondon: Can I create an issue to do this as a follow up? I didn't want to make this patch too big

pkg/manager/setconfig.go
SeanCondon: done

pkg/manager/modelregistry.go
SeanCondon: done

pkg/manager/modelregistry.go
SeanCondon: Can I create an issue to do this as a follow up? I didn't want to make this patch too big

modelplugin/ModelGenerator.sh
Andrea-Campanella: +1 on issue for this

pkg/manager/modelregistry.go
Andrea-Campanella: +1 also, not majorly important as of right now.

Issue #322, opened 2019-06-11
At startup if one of the devices listed in deviceStore is not contactable others halt


SeanCondon:
On startup of onos-config the system will try to connect to all of the devices in deviceStore.
Currently there are 4 devices
* 3 devicesim simulators
* 1 stratum device (localhost:50001)

The synchronizer factory can try connecting to these in any order - if one can't be contacted though the others can't proceed, meaning that there is behavior each time you start onos-config

Andrea-Campanella:
As per our discussion factory.go should spawn a routine passing an error channel to give back errors and timeouts.

Andrea-Campanella:
when #332 is merged we can close this in favour of the #333

PR #332, opened 2019-06-12
Lower timeouts for devices, remove return in Factory

Andrea-Campanella:
Lowering the device timeouts and not breaking the loop if one device fails connectivity. 
Partial Fix for: #322

### DIFF
--- a/configs/deviceStore-sample.json
+++ b/configs/deviceStore-sample.json
@@ -8,26 +8,26 @@
       "SoftwareVersion": "1.0.0",
       "User": "devicesim",
       "Pwd": "notused",
-      "Timeout": 10000000000
+      "Timeout": 5
     },
     "localhost-2": {
       "ID": "localhost-2",
       "Addr": "localhost:10162",
       "SoftwareVersion": "1.0.0",
-      "Timeout": 10000000000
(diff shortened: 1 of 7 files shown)


configs/deviceStore-sample.json
```diff
@@ -8,26 +8,26 @@
       "SoftwareVersion": "1.0.0",
       "User": "devicesim",
       "Pwd": "notused",
-      "Timeout": 10000000000
+      "Timeout": 5
```
SeanCondon: values are nanosecond - 5ns is too short - I suggest 5 seconds

pkg/southbound/clientManager.go
```diff
@@ -41,7 +41,7 @@ func createDestination(device topocache.Device) (*client.Destination, DeviceID)
 	d := &client.Destination{}
 	d.Addrs = []string{device.Addr}
 	d.Target = device.Target
-	d.Timeout = time.Duration(device.Timeout * time.Second)
+	d.Timeout = time.Duration(device.Timeout) * time.Second
```
SeanCondon: this is going to multiply it by 1,000,000,000

PR #343, opened 2019-06-13
Deleted the old admin and diags commands replaced by the new 'onos' CLI

tomikazi:


Andrea-Campanella:
I'll do some testing of the CLI today and then approve this based on that.

PR #344, opened 2019-06-13
Added docs for onos-cli container to deployment.md

tomikazi:


docs/deployment.md
```diff
@@ -250,6 +250,30 @@ Clients must connect through the HTTPS port using the certificates with which th
...
+
+```bash
+> kubectl run onos-cli --rm -it --image onosproject/onos-cli:latest --image-pull-policy "IfNotPresent" --restart "Never"
+```
+This command will run the CLI image as a Deployment and log into the bash shell.
+Once you've joined the container, you can connect to the `onos-config` server by running:
+
+```bash
+> onos config set address something-else-onos-config:5150
+something-else-onos-config:5150
```
Andrea-Campanella: why is it needed two times ? also can you provide a working command assuming onos-config as the name of the deployment ?

tomikazi: The command outputs that. Only the line prefixed with > is the CLI (unless it would have \ at the end)

PR #346, opened 2019-06-13
Smoke test framework in golang

ray-milkey:


cmd/smoke-test/smoketest.go
```diff
@@ -0,0 +1,204 @@
...
+	for _, pathElement := range pathElements {
+		protoBuilder.WriteString("elem: <name: '")
+		protoBuilder.WriteString(pathElement)
+		protoBuilder.WriteString("'> ")
+	}
+	protoBuilder.WriteString(">")
+	return protoBuilder.String()
+}
+
+func makeSetPathProto(target string, path string, value string) string {
```
Andrea-Campanella: you might want to consolidate this method and the one above. Avoid redundancy and error prone code

cmd/smoke-test/smoketest.go
```diff
@@ -0,0 +1,204 @@
...
+	protoBuilder.WriteString("', ")
+
+	pathElements := utils.SplitPath(path)
+	for _, pathElement := range pathElements {
+		protoBuilder.WriteString("elem: <name: '")
+		protoBuilder.WriteString(pathElement)
+		protoBuilder.WriteString("'> ")
+	}
+	protoBuilder.WriteString(">")
+	protoBuilder.WriteString(" val: <string_val: '")
```
Andrea-Campanella: just add this to the builder.

cmd/smoke-test/smoketest.go
```diff
@@ -0,0 +1,204 @@
...
+	protoBuilder.WriteString(value)
+	protoBuilder.WriteString("'>>")
+	return protoBuilder.String()
+}
+
+func findPathValue(response *gpb.GetResponse, path string) (string, error) {
+	if len(response.Notification) != 1 {
+		return "", errors.New("response notifications must have one entry")
```
Andrea-Campanella: just out of my curiosity, why not fmt.Errorf ?

ray-milkey: It should be. That seems to be the standard.

cmd/smoke-test/smoketest.go
ray-milkey: This was done in a subsequent set of changes

cmd/smoke-test/smoketest.go
ray-milkey: done

cmd/smoke-test/smoketest.go
ray-milkey: seems to have been refactored out

Andrea-Campanella: +1

cmd/smoke-test/smoketest.go
Andrea-Campanella: nice!

cmd/smoke-test/Dockerfile
```diff
@@ -0,0 +1,18 @@
+ARG ONOS_BUILD_VERSION=stable
```
Andrea-Campanella: Just out of curiosity, why are you creating a docker image just for the test ? what does this have that we can't move to the base docker image ?

cmd/smoke-test/smoketest_test.go
```diff
@@ -0,0 +1,183 @@
...
+
+	setTZRequest := &gpb.SetRequest{}
+	if err := proto.UnmarshalText(protoPath, setTZRequest); err != nil {
+		return err
+	}
+
+	_, err := c.(*gclient.Client).Set(ctx, setTZRequest)
+	return err
+}
+
```
Andrea-Campanella: I know this is WIP but it's a great start. I'm ok with adding subscribe ONCE, STREAM and POLL in a subsequent patch, or here if you prefer.

cmd/smoke-test/smoketest_test.go
```diff
@@ -0,0 +1,183 @@
...
+		Addrs:       q.Addrs,
+		Target:      q.Target,
+		Timeout:     q.Timeout,
+		Credentials: q.Credentials,
+		TLS:         q.TLS,
+	})
+	assert.NilError(t, err)
+	assert.Assert(t, c != nil, "Fetching client returned nil")
+
+	// First lookup should return an error - the path has not been given a value yet
```
Andrea-Campanella: my suggestion to make this modular is to have these as functions down below in the class and store them in an array in this "main" method. Iterating over the array and calling the function woudl execute the tests. anybody who wants to add his own could just writ a function and add it to that array.

ray-milkey:
Subsumed into the integration test framework changes

PR #349, opened 2019-06-13
Added in the Devicesim model plugin

SeanCondon:
I'm leaving aside the running of these plugins in the docker image for the moment, because the Makefile is going to change so much.


Makefile
```diff
@@ -21,7 +21,7 @@ deps: # @HELP ensure that the required dependencies are in place
 lint: # @HELP run the linters for Go source code
 	golint -set_exit_status github.com/onosproject/onos-config/pkg/...
 	golint -set_exit_status github.com/onosproject/onos-config/cmd/...
-	golint github.com/onosproject/onos-config/modelplugin/...
+	# golint github.com/onosproject/onos-config/modelplugin/...
```
Andrea-Campanella: can it be just removed ?

docs/run.md
```diff
@@ -10,7 +10,10 @@ go run github.com/onosproject/onos-config/cmd/onos-config \
     -configStore=$HOME/go/src/github.com/onosproject/onos-config/configs/configStore-sample.json \
     -changeStore=$HOME/go/src/github.com/onosproject/onos-config/configs/changeStore-sample.json \
     -deviceStore=$HOME/go/src/github.com/onosproject/onos-config/configs/deviceStore-sample.json \
-    -networkStore=$HOME/go/src/github.com/onosproject/onos-config/configs/networkStore-sample.json
+    -networkStore=$HOME/go/src/github.com/onosproject/onos-config/configs/networkStore-sample.json \
+    -modelPlugin=$HOME/go/src/github.com/onosproject/onos-config/modelplugin/TestDevice-1.0.0/testdevice.so.1.0.0 \
+    -modelPlugin=$HOME/go/src/github.com/onosproject/onos-config/modelplugin/TestDevice-2.0.0/testdevice.so.1.0.0 \
```
Andrea-Campanella: can we add just the emulator one ? the test device is not really present. Useful for testing though.

modelplugin/yang/ietf-interfaces@2014-05-08.yang
```diff
@@ -0,0 +1,111 @@
+module ietf-interfaces {
```
Andrea-Campanella: these files are because they cannot be pulled from openconfig public repo ?

SeanCondon: Yes - I found it very difficult to get these particular versions of the YANG files - had to go back through the Git history in open configs to get these - they are a snapshot in time, and so I wanted to keep them versioned. In fact I want to rename ietf-interfaces because it has to be modified - only the typedefs are needed from it and I had to comment out the containers to get it to compile

docs/run.md
SeanCondon: Individual users can add what they want when they run onos-config - I think it will be important to include all for testing purposes. I will also add a gRPC to manage these plugins

tomikazi:
Sean, sorry about that, but my PR cause a slight conflict on the Makefile here.

Makefile
SeanCondon: Done

SeanCondon:
> Sean, sorry about that, but my PR cause a slight conflict on the Makefile here.

no problem - I will rebase

docs/run.md
Andrea-Campanella: ok, I agree.

modelplugin/yang/ietf-interfaces@2014-05-08.yang
Andrea-Campanella: ok, it really shows that they want you to use the latest version! Interesting on the change/comment, why is that ?

SeanCondon: I changed ietf-interfaces@2014-05-08.yang to ietf-interfaces@2014-05-08.modified with a big comment at the top.
I haven't found which typedef exactly that the openconfig guys wanted from ietf-interfaces, but they redefine interfaces

Issue #347, opened 2019-06-13
Fix our build scheme


tomikazi:
**Is your feature request related to a problem? Please describe.**
We need to simplify our build process.

**Describe the solution you'd like**
* Update build.md to document set of tools required for local build
  * golang, go dep, golint, go vet?, Docker
* Retrofit golang-build to just protoc-build
* Clean-up the makefile
  * Direct invocation of protoc-build as a separate target
* Building docker images for onos-config and onos-cli separate targets and not default
* Default make target should do deps, compile, test, vet and lint
* Update travis stuff

PR #350, opened 2019-06-13
Streamlined the Makefile to address #347

tomikazi:


docs/build.md
```diff
@@ -1,48 +1,52 @@
...
+lint                            run the linters for Go source code
+onos-cli-docker                 build onos-cli Docker image
+onos-config-docker              build onos-config Docker image
+protos                          compile the protobuf files (using protoc-go Docker)
+run-docker                      run onos-config docker image
+test                            run the unit tests and source code validation
+vet                             examines Go source code and reports suspicious constructs
 ```
 
-See [run.md](run.md) for details on how to run the applications.
```
Andrea-Campanella: might still be needed ?

tomikazi: I don't think so. The build page is now fairly concise and it should not be difficult to back out and go to the 'how to run' page.

PR #353, opened 2019-06-13
Support DEBUG variable to build binaries with debugging support enabled

kuujo:
This PR adds support for building binaries with support for debugging enabled. This is done by supporting a `DEBUG` environment variable in the Makefile. When set, the `go build` command for onos-config is modified to add flags for the compiler to enable remote debugging. Additionally, the Dockerfile is updated to support the `DEBUG` flag through build arguments, so a debug image can be built.

This is just a first pass at this to ensure we can properly support debugging in k8s. I'm open to suggestions. I actually haven't been able to test this change in k8s because the `make images` command is failing for me (I'll open another bug report for that issue) on this PR and in master.

@ray-milkey once we figure this out, I think the Docker-y thing to do would be to always produce both an `onosproject/onos-config:x.x.x` image _and_ an `onosproject/onos-config:x.x.x-debug` image so users can deploy clusters with debugging enabled by changing the image, e.g. `helm install --set image="onosproject/onos-config:1.2.3-debug"`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,15 +3,15 @@ export CGO_ENABLED=0
 .PHONY: build
 
 ONOS_CONFIG_VERSION := latest
+ONOS_CONFIG_DEBUG_VERSION := debug
 ONOS_BUILD_VERSION := stable
-ONOS_CONFIG_BUILD_FLAGS := ""
 
 build: # @HELP build the Go binaries and run all validations (default)
 build: test
 ifndef DEBUG
 	go build -o build/_output/onos-config ./cmd/onos-config
 else
-	go build -gcflags "-N -l" -o build/_output/onos-config ./cmd/onos-config
+	go build -gcflags "all=-N -l" -o build/_output/onos-config ./cmd/onos-config
 endif
 	go build -o build/_output/onos ./cmd/onos
 	go build -o build/_output/testdevice.so.1.0.0 -buildmode=plugin ./modelplugin/TestDevice-1.0.0
@@ -55,9 +55,13 @@ protos: # @HELP compile the protobuf files (using protoc-go Docker)
 
 onos-config-docker: # @HELP build onos-config Docker image
 	docker build . -f build/onos-config/Dockerfile \
-		--build-arg ONOS_CONFIG_DEBUG=${DEBUG} \
     	--build-arg ONOS_BUILD_VERSION=${ONOS_BUILD_VERSION} \
 		-t onosproject/onos-config:${ONOS_CONFIG_VERSION}
+
+onos-config-docker-debug: # @HELP build onos-config Docker debug image
+	docker build . -f build/onos-config-debug/Dockerfile \
+		--build-arg ONOS_BUILD_VERSION=${ONOS_BUILD_VERSION} \
+		-t onosproject/onos-config:${ONOS_CONFIG_DEBUG_VERSION}
 
 onos-cli-docker: # @HELP build onos-cli Docker image
 	docker build . -f build/onos-cli/Dockerfile \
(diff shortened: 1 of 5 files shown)


kuujo:
@ray-milkey can you pull this and try to build the image? I still haven't been able to build the onos-config image locally, either on this PR or on master.

kuujo:
Okay, I got a little carried away with this tonight. I changed it a bit so there’s now a separate Dockerfile for the debut image which builds and installs the debugger binary and runs onos-config via the debugger. I also updated the Helm chart to support deploying the debug image with appropriate permissions. TBH it’s still a little messy but it does work.

kuujo:
I've removed with WIP tag. This is ready to be reviewed.

I'll add documentation for this change separately. But there are a few steps to deploying a debug image:

Run `make images` to build the debug image:
```
> make images
```

By default, `make images` will now build a `:debug` image in addition to `:latest` and the CLI. Then, ensure the debug image is available in Kubernetes (for Minikube, use the Minikube Docker daemon, for kind, run `kind load docker-image onosproject/onos-config:debug`):
```
> kind load docker-image onosproject/onos-config:debug
```

Then, deploy the `onos-config` Helm chart using the `debug` tag with the `debug` option enabled:
```
> helm install \
    -n onos-config \
    --set image.tag=debug \
    --set debug=true \
    --set devices='{device-1-device-simulator}' \
    deployments/helm/onos-config
```

The `--set debug=true` flag tells the Helm chart to deploy in debug mode. This enables certain security features needed for debugging, exposes the debug port, and disables probing to ensure the pod is not restarted when the application is paused by the debugger.

Once the pod is deployed, forward a local port to port `40000` on the pod with `kubectl`:
```
> kubectl port-forward onos-config-onos-config-54f49d6949-9nshq 40000:40000
```

Then, just connect the debugger in your IDE to port `40000` and the application will begin running.

Note that until the local debugger is connected to the debugger server inside the pod, the `onos-config` app will not run. Delve does not currently support a flag like `suspend=n` in Java. See go-delve/delve#245

Makefile
```diff
@@ -55,13 +60,18 @@ onos-config-docker: # @HELP build onos-config Docker image
...
+		-t onosproject/onos-config:${ONOS_CONFIG_DEBUG_VERSION}
+
 onos-cli-docker: # @HELP build onos-cli Docker image
 	docker build . -f build/onos-cli/Dockerfile \
         --build-arg ONOS_BUILD_VERSION=${ONOS_BUILD_VERSION} \
         -t onosproject/onos-cli:${ONOS_CONFIG_VERSION}
 
 images: # @HELP build all Docker images
-images: onos-config-docker onos-cli-docker
+images: onos-config-docker onos-config-docker-debug onos-cli-docker
```
SeanCondon: I wonder if it would be better to have an "images debug" other wise if I go "make run-docker" it will have to go off and build, vet, lint and test all 3 images everytime!

build/onos-config-debug/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+
+FROM onosproject/golang-build:$ONOS_BUILD_VERSION as onosBuilder
+COPY . /go/src/github.com/onosproject/onos-config
+RUN cd /go/src/github.com/onosproject/onos-config && DEBUG=true make build
+
+FROM golang:1.12.6-alpine3.9 as debugBuilder
+
+RUN apk upgrade --update --no-cache && apk add git && go get -u github.com/go-delve/delve/cmd/dlv
+
+FROM alpine:3.8
```
SeanCondon: is there a mismatch here 3.8 here, 3.9 above

Makefile
kuujo: Do we even need the CLI image for `run-docker`? Doesn't it make sense to just run `onos-config-docker` in `run-docker`?

SeanCondon: That's a question for Thomas - I presume he had some reason for putting it in there

PR #355, opened 2019-06-14
Adding in admin interface for Model plugins Fixes #334

### LINKED ISSUE #334: New gRPC service for model plugin registry to allow dynamically loading at runtime
Currently model plugins are loaded by setting them as command line arguments to **onos-config** on startup. It would be good to be able to additionally load and unload others dynamically at runtime

Add a gRPC service to add/remove/query the model plugins, and add a cli extension for these

Not bothering with this feature

**Additional context**
This can be added to the admin set of protos. The argument given would be the name of a named shared object **.so** library file. Should handle errors

SeanCondon:
Added to the

- admin.proto
- cli list-plugins and load-plugin

Also changed the protoc-go/Dockerfile - I import gnmi.ModelData into admin.proto so I needed this in docker.

Note: it's not possible to unload a plugin from a running go program, so no unload.


pkg/cli/command/list-plugins.go
```diff
@@ -0,0 +1,66 @@
+// Copyright 2019-present Open Networking Foundation.
```
tomikazi: This doesn't have to be done now, but I would suggest to put both commands into 'plugins.go' (or models.go) and then structure the commands for 'onos plugins {list|load} usage' or perhaps 'onos models {list|load}'

docs/run.md
```diff
@@ -12,9 +12,15 @@ go run github.com/onosproject/onos-config/cmd/onos-config \
...
     -networkStore=$HOME/go/src/github.com/onosproject/onos-config/configs/networkStore-sample.json \
     -modelPlugin=$HOME/go/src/github.com/onosproject/onos-config/modelplugin/TestDevice-1.0.0/testdevice.so.1.0.0 \
-    -modelPlugin=$HOME/go/src/github.com/onosproject/onos-config/modelplugin/TestDevice-2.0.0/testdevice.so.1.0.0 \
+    -modelPlugin=$HOME/go/src/github.com/onosproject/onos-config/modelplugin/TestDevice-2.0.0/testdevice.so.2.0.0 \
     -modelPlugin=$HOME/go/src/github.com/onosproject/onos-config/modelplugin/Devicesim-1.0.0/devicesim.so.1.0.0
 ```
+> This gives a list of model plugins to load at startup. Alternatively these can
+> be omitted here and loaded later with the **onos load-plugin** command. The
+> plugins here were built locally with a command like
+> **go build -o modelplugin/TestDevice-1.0.0/testdevice.so.1.0.0 -buildmode=plugin ./modelplugin/TestDevice-1.0.0**
```
tomikazi: I'd suggest that for commands we use the back-tick command mark-down instead.

docs/cli.md
```diff
@@ -119,6 +121,27 @@ device-3: localhost-3 (1.0.0)
...
+A model plugin is a shared object library that represents the YANG models of a
+particular Device Type and Version. The plugin allows user to create and load
+their own device models in to onos-config that can be used for validating that
+configuration changes observe the structure of the YANG models in use on the
+device. This improves usability by pushing information about the devices'
+model back up to the onos-config gNMI northbound interface.
+
+Model plugins can be loaded at the startup of onos-config by (repeated) --modelPlugin
+options, or they can be loaded at run time with the command:
+```bash
```
tomikazi: Text above implies that what's coming is a load command, not list command.

docs/run.md
SeanCondon: Done

docs/cli.md
SeanCondon: Good catch

PR #356, opened 2019-06-14
gNMI Northbound Error propagation

### LINKED ISSUE #331: Error propagation gNMI subscribe NB
An error propagation mechanism is needed to propagate different type of errors to the gRPC NB. 

doing some research the common patter is to have an Error Channel with routines that reads the values from it and handles accordingly (if it needs to do so). This would enable also to have a unique point of access to all errors for publishing them to other components if needed. 

A possibile alternative is to make all methods return errors but this is hard with go routines.

Andrea-Campanella:
Propagates error in the Northbound of gNMI on the gRPC channel through a result go channel and proper codes.

Fixes #331

### DIFF
--- a/pkg/northbound/gnmi/set.go
+++ b/pkg/northbound/gnmi/set.go
@@ -76,8 +76,8 @@ func (s *Server) Set(ctx context.Context, req *gnmi.SetRequest) (*gnmi.SetRespon
 		} else if ext.GetRegisteredExt().GetId() == GnmiExtensionDeviceType {
 			deviceType = string(ext.GetRegisteredExt().GetMsg())
 		} else {
-			return nil, fmt.Errorf("Unexpected extension %d = '%s' in Set()",
-				ext.GetRegisteredExt().GetId(), ext.GetRegisteredExt().GetMsg())
+			return nil, status.Error(codes.InvalidArgument, fmt.Errorf("Unexpected extension %d = '%s' in Set()",
+				ext.GetRegisteredExt().GetId(), ext.GetRegisteredExt().GetMsg()).Error())
 		}
 	}
 
@@ -89,7 +89,7 @@ func (s *Server) Set(ctx context.Context, req *gnmi.SetRequest) (*gnmi.SetRespon
 
 	if len(updateResults) == 0 {
 		log.Println("All target changes were duplicated - Set rejected")
-		return nil, status.Error(codes.Internal, fmt.Errorf("set change rejected as it is a "+
+		return nil, status.Error(codes.AlreadyExists, fmt.Errorf("set change rejected as it is a "+
 			"duplicate of the last change for all targets").Error())
 	}
 
@@ -100,7 +100,7 @@ func (s *Server) Set(ctx context.Context, req *gnmi.SetRequest) (*gnmi.SetRespon
 	// Look for use of this name already
 	for _, nwCfg := range manager.GetManager().NetworkStore.Store {
 		if nwCfg.Name == netcfgchangename {
-			return nil, status.Error(codes.Internal, fmt.Errorf(
+			return nil, status.Error(codes.InvalidArgument, fmt.Errorf(
 				"Name %s is already used for a Network Configuration", netcfgchangename).Error())
 		}
 	}
--- a/pkg/northbound/gnmi/set.go
+++ b/pkg/northbound/gnmi/set.go
@@ -17,10 +17,6 @@ package gnmi
 import (
 	"context"
 	"fmt"
-	"log"
-	"strings"
-	"time"
-
 	"github.com/docker/docker/pkg/namesgenerator"
 	"github.com/onosproject/onos-config/pkg/manager"
 	"github.com/onosproject/onos-config/pkg/store"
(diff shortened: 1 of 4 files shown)


pkg/northbound/gnmi/set.go
```diff
@@ -17,23 +17,119 @@ package gnmi
...
+
+	for _, ext := range req.GetExtension() {
+		if ext.GetRegisteredExt().GetId() == GnmiExtensionNetwkChangeID {
+			netcfgchangename = string(ext.GetRegisteredExt().GetMsg())
+		} else if ext.GetRegisteredExt().GetId() == GnmiExtensionVersion {
+			version = string(ext.GetRegisteredExt().GetMsg())
+		} else if ext.GetRegisteredExt().GetId() == GnmiExtensionDeviceType {
+			deviceType = string(ext.GetRegisteredExt().GetMsg())
+		} else {
+			return nil, fmt.Errorf("Unexpected extension %d = '%s' in Set()",
```
SeanCondon: Is there a way to set a code on this - I don't think it would be Internal though - more like InvalidArgument

pkg/northbound/gnmi/set.go
```diff
@@ -17,23 +17,119 @@ package gnmi
...
+
+	updateResults, networkChanges, err :=
+		s.buildUpdateResults(targetUpdates, targetRemoves, version, deviceType)
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	if len(updateResults) == 0 {
+		log.Println("All target changes were duplicated - Set rejected")
+		return nil, status.Error(codes.Internal, fmt.Errorf("set change rejected as it is a "+
```
SeanCondon: Maybe AlreadyExists for the error code

pkg/northbound/gnmi/set.go
```diff
@@ -17,23 +17,119 @@ package gnmi
...
+	}
+
+	if netcfgchangename == "" {
+		netcfgchangename = namesgenerator.GetRandomName(0)
+	}
+
+	// Look for use of this name already
+	for _, nwCfg := range manager.GetManager().NetworkStore.Store {
+		if nwCfg.Name == netcfgchangename {
+			return nil, status.Error(codes.Internal, fmt.Errorf(
```
SeanCondon: Again maybe AlreadyExists or invalidArgument

pkg/northbound/gnmi/subscribe.go
```diff
@@ -44,79 +52,105 @@ func (s *Server) Subscribe(stream gnmi.GNMI_SubscribeServer) error {
 	changesChan, err := mgr.Dispatcher.RegisterNbi(hash)
 	if err != nil {
 		log.Println("Subscription present: ", err)
-		return err
+		return status.Error(codes.Internal, err.Error())
```
SeanCondon: Maybe AlreadyExists

pkg/northbound/gnmi/subscribe.go
```diff
@@ -44,79 +52,105 @@ func (s *Server) Subscribe(stream gnmi.GNMI_SubscribeServer) error {
...
 	}
 	opStateChan, err := mgr.Dispatcher.RegisterOpState(hash)
 	if err != nil {
 		log.Println("Subscription present: ", err)
-		return err
+		return status.Error(codes.Internal, err.Error())
```
SeanCondon: same here

pkg/northbound/gnmi/subscribe.go
```diff
@@ -44,79 +52,105 @@ func (s *Server) Subscribe(stream gnmi.GNMI_SubscribeServer) error {
...
-	for {
-		in, err := stream.Recv()
-		if err == io.EOF {
-			log.Println("Subscription Terminated")
-			//Ignoring Errors during removal
-			mgr.Dispatcher.UnregisterNbi(hash)
-			mgr.Dispatcher.UnregisterOperationalState(hash)
-			return nil
-		}
+	go func() {
```
SeanCondon: why is this now a go routine - I thought the for was to block it from returning - also I don't like anon go functions like this - could you make it a named function

pkg/northbound/gnmi/subscribe.go
```diff
@@ -44,79 +52,105 @@ func (s *Server) Subscribe(stream gnmi.GNMI_SubscribeServer) error {
...
-		return err
+		return status.Error(codes.Internal, err.Error())
 	}
 	opStateChan, err := mgr.Dispatcher.RegisterOpState(hash)
 	if err != nil {
 		log.Println("Subscription present: ", err)
-		return err
+		return status.Error(codes.Internal, err.Error())
 	}
+	resChan := make(chan result)
```
SeanCondon: nice

pkg/northbound/gnmi/subscribe.go
```diff
@@ -44,79 +52,105 @@ func (s *Server) Subscribe(stream gnmi.GNMI_SubscribeServer) error {
...
-			//Ignoring Errors during removal
-			mgr.Dispatcher.UnregisterNbi(hash)
-			mgr.Dispatcher.UnregisterOperationalState(hash)
-			return err
-		}
+			if err != nil {
+				code, ok := status.FromError(err)
+				if ok && code.Code() == codes.Canceled {
+					log.Println("Subscription Terminated")
+					resChan <- result{success: true, err: nil}
```
SeanCondon: You don't exit here?

pkg/northbound/gnmi/subscribe.go
```diff
@@ -44,79 +52,105 @@ func (s *Server) Subscribe(stream gnmi.GNMI_SubscribeServer) error {
...
+				} else {
+					log.Println("Error in subscription", err)
+					//Ignoring Errors during removal
+					mgr.Dispatcher.UnregisterNbi(hash)
+					mgr.Dispatcher.UnregisterOperationalState(hash)
+					resChan <- result{success: false, err: err}
```
SeanCondon: shouldn't this exit here?

pkg/northbound/gnmi/subscribe.go
```diff
@@ -44,79 +52,105 @@ func (s *Server) Subscribe(stream gnmi.GNMI_SubscribeServer) error {
...
 			}
-			//Each subscription request spawns a go routing listening for related events for the target and the paths
-			go listenForUpdates(changesChan, stream, mgr, targets, subsStr)
-			go listenForOpStateUpdates(opStateChan, stream, targets, subsStr)
 		}
-	}
+	}()
 
+	for result := range resChan {
+		if !result.success {
```
SeanCondon: will this wait for its child threads to **join** before it can complete?

pkg/northbound/gnmi/subscribe.go
```diff
@@ -169,9 +212,13 @@ func buildAndSendUpdate(pathGnmi *gnmi.Path, target string, value string, stream
 		Val:  valueGnmi,
 	}
 	response := buildUpdateResponse(update)
-	sendResponse(response, stream)
+	err := sendResponse(response, stream)
+	if err != nil {
+		return err
```
SeanCondon: should this get its own code?

pkg/northbound/gnmi/set.go

pkg/northbound/gnmi/set.go

pkg/northbound/gnmi/set.go

pkg/northbound/gnmi/subscribe.go
Andrea-Campanella: done

pkg/northbound/gnmi/subscribe.go
Andrea-Campanella: done

pkg/northbound/gnmi/subscribe.go
Andrea-Campanella: no, the the result gets issued and the for loop picks it up. I break the for effectively ending and enabling garbage collection on the routine.

pkg/northbound/gnmi/subscribe.go
Andrea-Campanella: see before

pkg/northbound/gnmi/subscribe.go
Andrea-Campanella: there is just one go routine, synchronisation is handled but the "result" message. There will be one result message out of the resChan and will be handled here.

pkg/northbound/gnmi/subscribe.go
Andrea-Campanella: done underneath

pkg/northbound/gnmi/subscribe.go
Andrea-Campanella: the blocking is done in the for under this function that waits for this to return.
Moved to function.

PR #363, opened 2019-06-17
Enable CGO to allow plugins run in Docker

SeanCondon:


Makefile
```diff
@@ -65,12 +65,16 @@ all: build images
 
 run-docker: # @HELP run onos-config docker image
 run-docker: images
-	docker run -d -p 5150:5150 -v `pwd`/configs:/etc/onos-config \
+	docker stop onos-config
```
Andrea-Campanella: this probably out to be part of clean, or his own step, including stop of cli, thoughts ?

SeanCondon: The reason I added it is that it is not possible run run it twice in a row without deleting container.

I'd like to leave it here **and** add it to **clean** without having to force **clean** to be run.

SeanCondon:
I've temporarily marked as do not merge as Jordan's change should go first

Makefile
SeanCondon: changed to
```
docker stop onos-config || echo "onos-config was not running"
```

kuujo:
Embarrassingly, I spent a non-trivial amount of time trying to get this to work last week and failed. I guess I was using the wrong package. But this works!

PR #368, opened 2019-06-18
Moving project to go modules

Andrea-Campanella:
- Removed deps
- removing gopkg.lock, gopkg.tom
- adding go.sum, go.mod
- excluding modelplugins from the build since they have no main. Have to built manually or included in the build with ` -tags=modelplugin`
- updated docs


Makefile
```diff
@@ -13,13 +13,15 @@ ifndef DEBUG
 else
 	CGO_ENABLED=1 go build -gcflags "all=-N -l" -o build/_output/onos-config ./cmd/onos-config
 endif
+	export GO111MODULE=on
```
SeanCondon: but it is not needed for onos-config build?

docs/modelplugin.md
```diff
@@ -52,7 +52,7 @@ Once the files are created:
 1) Change directory back to **onos-config**
 2) Compile the plugin with (replacing the names as appropriate)
 ```bash
-> CGO_ENABLED=1 go build -o modelplugin/TestDevice-1.0.0/testdevice.so.1.0.0 -buildmode=plugin ./modelplugin/TestDevice-1.0.0
+> CGO_ENABLED=1 go build -o modelplugin/TestDevice-1.0.0/testdevice.so.1.0.0 -buildmode=plugin -tags=modelplugin ./modelplugin/TestDevice-1.0.0
```
SeanCondon: is this a new tag? this should probably be in the Makefile invocation too?

PR #374, opened 2019-06-20
Integration test framwork

kuujo:
This PR adds an integration test framework that runs tests on Kubernetes.

The only prerequisite to running tests is that the test image be built via `make images` or `make onos-it-docker`. The integration test image must be accessible within the Kubernetes cluster, e.g. by adding it in `kind`:

```bash
> kind load docker-image onosproject/onos-config-integration-tests:latest
```

The test framework is implemented in the `/test` directory as per the Go standard project layout. Within the test framework, there's an `onos-it` command which is used to run integration tests. The tests themselves are implemented in `/test/integration` using the `testing` package, and tests are registered with human-readable names in an `init()` function:

```go
func SomeTest(t *testing.T) {
    ...
}

func init() {
    Registry.Register("some-test", SomeTest)
}
```

This allows specific tests to be easily run through the `onos-it` command:

```bash
> go run github.com/onosproject/onos-config/test/cmd/onos-it run some-test
```

When a test is run, the Kubernetes test controller implemented in `/test/runner/controller.go` sets up a test cluster and runs the test inside a Kubernetes `Job`. The controller performs the following steps to setup the cluster:
* Create a namespace with a UUID based name for the test resources
* Deploy the [Atomix Kubernetes Controller](https://github.com/atomix/atomix-k8s-controller)
* Create a Raft `PartitionSet` for use by onos-config
* Deploy [device simulators](https://github.com/onosproject/simulators)
* Deploy onos-config nodes
* Create a `Job` that simply runs the `onosproject/onos-config-integration-tests:latest` image

While the test is running, the controller streams the test pod’s logs to stdout, and once the job is complete the controller gets the exit code from Kubernetes, cleans up k8s resources, and exits with the same exit code.

The integration tests image is simply a copy of the onos-config source. When run, the image calls `onos-it test` with the arguments passed to `onos-it run`. The `onos-it test` command runs the tests locally.

For onos-config, it's essential that we be able to run tests under different configurations. For that reason, the `onos-it run` command supports a few optional flags:
* `--nodes`/`-n` - the number of onos-config nodes to create for the test
* `--partitions`/`-p` - the number of Raft partitions to create for the test
* `--partitionSize`/`-s` - the size of Raft partitions to create for the test
* `--timeout`/`-t` - an optional test timeout (defaults to 10 minutes)

These flags allow the number of nodes to be configured in both onos-config and Atomix. But simulators are notably missing from the flags. In order to support simulator configurations and other custom onos-config configurations, the `onos-it` command supports custom JSON configurations specifying the onos-config stores and simulators. These configurations are defined in files in `/test/configs` and can be referenced by name with the `--config` or `-c` option, e.g. to run a test with three simulators:

```bash
onos-it run -c simulator-3 some-test
```

By default, the `default` configuration is used, and `default.json` defines all the stores using the default configurations currently used for stores. If the `simulators` key in the JSON configuration is defined, the k8s controller will deploy the simulators with the provided configurations and configure onos-config with the defined simulators in its device store.


kuujo:
FYI there’s still a lot of testing to do here. I’ve verified that it sets up the Atomix controller and Raft partitions correctly and at least runs the onos-config nodes, but I haven’t verified everything is correctly configured. We should add some basic tests that require/use simulators as well.

Also, there’s a question as to what should be configured outside of tests and what should be specified inside tests. The number of Raft partitions and onos-config nodes should never change the behavior of tests, but do we want the ability to start new simulators during tests? Maybe that’s a change to be addressed later if necessary.

Andrea-Campanella:
>  We should add some basic tests that require/use simulators as well.
I think rays smoke test is a perfect candidate here. 
> 
> Also, there’s a question as to what should be configured outside of tests and what should be specified inside tests. The number of Raft partitions and onos-config nodes should never change the behavior of tests, but do we want the ability to start new simulators during tests? Maybe that’s a change to be addressed later if necessary.

I believe adding a simulator would be useful to show runtime addition of it to the system, adding it's model and provision things to it.

kuujo:
I need someone to test this to make sure I’m not missing anything from my environment needed to test. The basic steps to verify you’re seeing what I’m seeing is:
• Install a test cluster (Minikube or kind) and ensure you’re building in that environment (`eval $(minikube docker-env)` for Minikube
• Pull the PR and build the image with `make images`
• For kind, add the image to the cluster with `kind load docker-image onosproject/onos-config-integration-tests:latest`
• Ensure your shell is configured to reach the rest k8s cluster (if `kubectl get nodes` works then the test framework should work)
• Run `go run github.com/onosproject/onos-config/test/cmd/onos-it run test-integration-test -c simulators-1`. This runs the single test case we have with a single simulator.

You should see the test controller setup the namespace, Atomix cluster, Raft partitions, onos-config cluster, and run the test. What I see currently is some test failures, but the tests do seem to be communicating with onos-config and the simulators are correctly connected.

kuujo:
There is one step that may be required. I haven't yet tested. If the above does not work, you may need to add the Atomix custom resources before running the test:
```bash
kubectl create -f https://raw.githubusercontent.com/atomix/atomix-k8s-controller/master/deploy/crds.yaml
```

I have to update the k8s Go client to be able to do this programmatically, but I ran into some dependency issues so it's currently on the TODO list.

kuujo:
@Andrea-Campanella that was my suspicion re: the simulators. There’s value to being able to manage them at runtime. But I suppose that’s something we need to tackle as stores are developed. It should be simple to move some code around to deploy them at runtime from inside a pod rather than outside when we’re ready.

Andrea-Campanella:
@kuujo Yes indeed, I think we can also tackle this step by step. Static number of simulators is fine for now, this is of incredible value as it is. We can move to dinamic later, also when we understand the relationship with enos-topology. @tomikazi when should we start thinking of something ?

Andrea-Campanella:
@kuujo 
```
I0621 12:59:05.913475   24840 controller.go:1208] Waiting for test job onos-test-29db31fa-9413-11e9-8e84-8c859015ae0f to become ready
go: finding github.com/onosproject/onos-config/test/cmd/onos-it latest
go: finding github.com/onosproject/onos-config/test/cmd latest
go: finding github.com/onosproject/onos-config/test latest
package github.com/onosproject/onos-config/test/cmd/onos-it: unknown import path "github.com/onosproject/onos-config/test/cmd/onos-it": cannot find module providing package github.com/onosproject/onos-config/test/cmd/onos-it
I0621 12:59:16.180752   24840 controller.go:1273] Tearing down test namespace onos-test-29db31fa-9413-11e9-8e84-8c859015ae0f
``` 
I was seeing this. It was not finding the onit command 
I made it work with this change 
`renamed:    test/cmd/onit/main.go -> test/cmd/onos-it/main.go`

Issue #376, opened 2019-06-20
Move Go build outside of Docker builds


kuujo:
Currently, Docker image builds require dependency pull steps, tests, etc. We use a build layer to build the artifacts and then copy them into the final layer. This is the way Docker/Moby wants us to build the image, but it results in a lot of unnecessary cycles when building images. We can instead simply build the executables on the host where all the dependencies are already present and then copy those artifacts into Docker images.

PR #382, opened 2019-06-21
Optimize docker build

kuujo:
#376 

Currently, the `make images` target builds four containers and pulls dependencies each time it builds one of those containers. This requires an unreasonable amount of time. This PR significantly optimizes the Docker build by adding a base image `onosproject/onos-config-base` which pulls dependencies and builds the binaries only once.

But optimally, the Docker build shouldn't even need to pull dependencies at all. The dependencies are normally available on the host. But because of the way the go module system works, it was difficult to figure out how to add the module's dependencies to the base image. So, instead, `make onos-config-base-docker` target and the base Docker image use Go modules' support for legacy vendoring to sneak the dependencies into the base image. When `make images` is run, `go mod vendor` is used to create the `vendor` cache, and the `vendor` is then added to the base image to ensure dependencies don't need to be pulled. After the Docker build is complete, the `vendor` folder is removed.

Obviously the vendoring approach is not ideal, but it saves a huge amount of time on image builds. I'd prefer if we could move the `vendor` folder to a hidden folder like `build/_output`, but alas I couldn't figure out how to make that happen.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,9 +54,11 @@ protos: # @HELP compile the protobuf files (using protoc-go Docker)
 		onosproject/protoc-go:stable
 
 onos-config-base-docker: # @HELP build onos-config base Docker image
+	go mod vendor
 	docker build . -f build/base/Dockerfile \
 		--build-arg ONOS_BUILD_VERSION=${ONOS_BUILD_VERSION} \
 		-t onosproject/onos-config-base:${ONOS_CONFIG_VERSION}
+	rm -rf vendor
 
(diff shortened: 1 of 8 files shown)


kuujo:
BTW the reason we’re not just adding binaries from the local `build` is because we can’t easily cross-compile for Linux/amd64 on Mac OS. After some brief investigation, I think it’s not worth the setup requirements needed to support cross-compilation. Using the base Docker image to compile the binary is almost as efficient.

PR #389, opened 2019-06-21
use go mod download to avoid vendoring for building the images

adibrastegarnia:
After discussing it with Jordan, I think we can use go mod download to avoid vendoring and speed up the build process for the images.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,11 +54,9 @@ protos: # @HELP compile the protobuf files (using protoc-go Docker)
 		onosproject/protoc-go:stable
 
 onos-config-base-docker: # @HELP build onos-config base Docker image
-	@go mod vendor
 	docker build . -f build/base/Dockerfile \
 		--build-arg ONOS_BUILD_VERSION=${ONOS_BUILD_VERSION} \
 		-t onosproject/onos-config-base:${ONOS_CONFIG_VERSION}
-	@rm -rf vendor
 
 onos-config-docker: onos-config-base-docker # @HELP build onos-config Docker image
 	docker build . -f build/onos-config/Dockerfile \
--- a/build/base/Dockerfile
+++ b/build/base/Dockerfile
@@ -2,5 +2,7 @@ ARG ONOS_BUILD_VERSION=stable
 
 FROM onosproject/golang-build:$ONOS_BUILD_VERSION
 ENV GO111MODULE=on
+COPY go.mod go.sum /go/src/github.com/onosproject/onos-config/
+RUN  go mod download
 COPY . /go/src/github.com/onosproject/onos-config
-RUN cd /go/src/github.com/onosproject/onos-config && GOFLAGS=-mod=vendor make build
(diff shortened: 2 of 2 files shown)


kuujo:
@adibrastegarnia this actually doesn't solve the problem, and it's because of the same issue I was running into when I tried to do this. The problem is, whenever the files change (e.g. the files added via `COPY`), Docker will always rebuild starting at the step at which the files changed, and from that point on all later steps have to be run as if they've never been run before. So because the entire source tree is being copied prior to `go mod download`, any time the source tree is changed the `go mod download` step will have to be run again, and the layer that exists prior to the `go mod download` step does not contain any of the dependencies so they'll all have to be downloaded again.

To demonstrate this, simply run `make onos-config-base-docker` once, then modify basically any file in the source tree, and then run `make onos-config-base-docker` again. Because the files copied in `COPY . /go/src/github.com/onosproject/onos-config` have changed, the build will be recreated from that point forward, including `go mod download`.

What we would need to eliminate the use of `go mod vendor` is to be able to copy _only_ `go.mod` and `go.sum` into the image and then run `go build ./...` or `go mod download`, then move the `COPY .` step _after_ `go mod download`:

```
COPY go.mod go.sum /go/src/github.com/onosproject/onos-config/
RUN  go mod download
COPY . /go/src/github.com/onosproject/onos-config
RUN cd /go/src/github.com/onosproject/onos-config && make build
```

That would ensure that `COPY go.mod go.sum ...` and `RUN go mod download` only ever have to be run again when `go.mod` or `go.sum` change. So, when we modify the source tree, as long as dependencies aren't changed the build would start at `COPY .` and be able to run `make build` on dependencies that have already been built in an earlier cached step.

adibrastegarnia:
@kuujo 
Thanks for checking that. Yes you are correct. The source should be copied after the dependencies copies. I will make the changes.

kuujo:
I think this still has the same issue. The `go mod download` step fails to actually download any dependencies AFAICT:

```
Step 5/7 : RUN  go mod download
 ---> Running in 059f90476707
warning: pattern "all" matched no module dependencies
```

The dependencies in my test are still being downloaded during `make build` and after the `COPY .` step.

kuujo:
Relevant:
https://github.com/golang/go/issues/27719

kuujo:
I don't know... there are a few solutions in that issue but TBH none of them even seem as elegant as just using `go mod vendor`. It's better than either committing a dummy main or maintaining and committing a list of dependencies separate from `go.mod` and using them to download dependencies.  Using `go mod vendor` is a pretty succinct way to get and copy in only the dependencies.

adibrastegarnia:
@kuujo 

Those solutions are not solid solutions as you said and some of them behave like vendoring  to me which is dependent to something else other than go.mod and go.sum files. It looks like current go build system doesn't support that very well. Until we find a better solution, I will close this pull request.

kuujo:
Yeah I think something needs to come of the issue I linked above before this can be improved any more

PR #392, opened 2019-06-24
Setting pre-existing config to device. Fixes #379

### LINKED ISSUE #379: Sync current config in controller to device at connection time
Currently onos-config allows to set a configuration for a non connected device. if the device then shows up the configuration needs to be checked against the device itself and propagated to it. 

At device connection the synchroniser should fetch any existing config change for the device in the config and change store and then apply it.

Andrea-Campanella:
This patch adds capability when a device connects to check against the config store and to set any pre-specified configuration to that device.

### DIFF
--- a/pkg/southbound/synchronizer/synchronizer.go
+++ b/pkg/southbound/synchronizer/synchronizer.go
@@ -23,16 +23,20 @@ import (
 	"github.com/onosproject/onos-config/pkg/southbound"
 	"github.com/onosproject/onos-config/pkg/southbound/topocache"
 	"github.com/onosproject/onos-config/pkg/store"
+	"github.com/onosproject/onos-config/pkg/store/change"
 	"github.com/onosproject/onos-config/pkg/utils"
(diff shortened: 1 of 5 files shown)


pkg/southbound/synchronizer/synchronizer.go
```diff
@@ -42,34 +46,70 @@ type Synchronizer struct {
...
 
-	log.Info(sync.Device.Addr, "Capabilities", capResponse)
+	log.Info(sync.Device.Addr, " Capabilities ", capResponse)
+
+	config, err := getNetworkConfig(sync, string(sync.Device.ID), "", "/*", 0)
+
+	//Device has initial configuration saved in onos-config, trying to apply
+	if err == nil {
+
+		initalConfig, err := change.CreateChangeValuesNoRemoval(config, "Initial set to device")
```
ray-milkey: typo - initial

Andrea-Campanella: done

PR #394, opened 2019-06-24
Update bug_report.md

Andrea-Campanella:
adding example text writing


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -19,8 +19,8 @@ Steps to reproduce the behavior:
 **Expected behavior**
 A clear and concise description of what you expected to happen.
 
-**Screenshots**
-If applicable, add screenshots to help explain your problem.
+**Example text or Screenshots**
+If applicable, add example text and/or screenshots to help explain your problem.
```
ray-milkey: Perhaps add something here like:

If a stack trace is generated, or there is an ERROR displayed in the log file, you can capture that here also.

Issue #383, opened 2019-06-21
Add support for cluster setup in onit


kuujo:
One of the common use cases for `stc` was simply setting up a cluster and debugging it by hand to reproduce issues. The `onit` command should similarly support a cluster setup command to allow users to manually interact with test clusters.

PR #397, opened 2019-06-25
Add separate cluster management commits to onit

kuujo:
#383 

This PR adds separate commands for managing test clusters via `onit`. The command now stores configurations for clusters, supports multiple clusters, allows simulators to be added to/removed from clusters, etc.

To create a cluster:
```
onit create cluster
```

Once the cluster has been created, you will see it in the clusters list:
```
onit get clusters
```

It will also be set as the default cluster:
```
onit get cluster
```

To add a simulator to the cluster:
```
onit add simulator
```

When the simulator is added, the onos-config stores will be updated and the pods will be redeployed with the updated configuration.

Similarly, simulators can be removed:
```
onit remove simulator
```

Once the cluster has been setup, tests can be run as normal:
```
onit run test-integration-test
```


docs/testing.md
```diff
@@ -20,13 +20,11 @@ For Minikube, ensure you build the image in the Minikube Docker context:
 ```
 
 For kind, you must manually `load` the `onos-config` image and the integration tests image 
-into the Kubernetes cluster each time the tests are modified:
+into the Kubernetes cluster each time the tests are modified. The `make kind` target will
```
Andrea-Campanella: you might want to add that if you make changes to the simulators or cli or debug images you need to load those manually and provide the `kind load docker-image onosproject/<image>:<tag>` example

test/runner/config.go
```diff
@@ -0,0 +1,326 @@
...
+	configs := []string{}
+	filepath.Walk(deviceConfigsPath, func(path string, info os.FileInfo, err error) error {
+		if info.IsDir() {
+			return nil
+		}
+
+		ext := filepath.Ext(info.Name())
+		if ext == ".json" {
+			configs = append(configs, info.Name()[:len(info.Name())-len(ext)])
+		}
```
Andrea-Campanella: shoudl we drop an error if we don't have `.json` here ?

docs/testing.md
kuujo: @Andrea-Campanella the `make kind` target runs `kind load ...`

test/runner/config.go
kuujo: I think we should allow for something like `.md` files to be in the config directories

tomikazi:
This is awesome Jordan!

docs/testing.md
Andrea-Campanella: @kuujo +1

kuujo:
Added additional commands for viewing the state of clusters and tests. Most commands are now documented in the updated `testing.md` documentation.

PR #403, opened 2019-06-25
Cleanup onit cluster controller

kuujo:
This PR does not change any onit code, but just refactors the `controller.go` file into many files, each with setup/teardown receivers for a different layer of the cluster:
* `atomix.go` manages the Atomix controller
* `partitions.go` manages Raft partitions
* `simulator.go` manages device simulators
* `onos-config.go` manages the onos-config nodes
* `test.go` runs tests


kuujo:
Tests hanged. Restarting build

PR #405, opened 2019-06-25
Namespace onit clusters from cluster ID

kuujo:
This PR modifies the use of a generated name for the cluster namespace in onit to allow users to pass a cluster ID which will be used directly to create the test namespace. This allows `kubectl` commands to more easily be used when a cluster is explicitly named.


tomikazi:
Requires rebase.

kuujo:
Ugh rebase is not right...

kuujo:
That should be good

Issue #408, opened 2019-06-26
Read onit cluster configuration from Kubernetes


kuujo:
Currently, the configuration for all onit clusters are stored in the onit configuration file. When users modify the Kubernetes cluster directly, the onit configuration file can reflect resources that have changed or may not actually be running at all. The same can happen when a user switches k8s clusters.

onit should read its configuration from Kubernetes directly, e.g. cluster list from namespaces, nodes/simulators/partitions from pods, etc to ensure the client always accurately reflects what is running in the k8s cluster.

PR #409, opened 2019-06-26
Refactor onit configuration to read all configuration data from Kubernetes resources

kuujo:
#408 

This PR modifies the onit configuration to move the majority of it to Kubernetes. Clusters are read from k8s, as are nodes, partitions, and tests. This ensures the cli is consistent with whatever k8s cluster it's connected to.


kuujo:
No changes. Only the configuration is way better :-)

Issue #402, opened 2019-06-25
Support debugging in onit clusters


kuujo:
I expect a very common use case for onit clusters will be debugging. The work on supporting remote debugging in onos-config should be merged with onit. Onit should deploy debug images by default, and we should be able to make convenience commands to make remote debugging more straightforward via `onit`, e.g. to open a debut port on the host:
```
onit debug node 1
```

PR #410, opened 2019-06-26
Support remote debugging in onit clusters

kuujo:
#402

This PR implements support for remote debugging in onit clusters. Because delve currently [has no elegant way to avoid pausing at startup](https://github.com/go-delve/delve/issues/245) to wait for a debugger to be attached, a hack has to be implemented to unpause the debugger when onos-config nodes are bootstrapped. Because this is a questionable process, I'm adding the WIP tag to test it a bit more before this is merged.

### DIFF
--- a/test/runner/onos-config.go
+++ b/test/runner/onos-config.go
@@ -82,10 +82,10 @@ func (c *ClusterController) setupOnosConfig() error {
 	if err := c.createOnosConfigConfigMap(); err != nil {
 		return err
 	}
-	if err := c.createOnosConfigDeployment(); err != nil {
+	if err := c.createOnosConfigService(); err != nil {
 		return err
 	}
-	if err := c.createOnosConfigService(); err != nil {
+	if err := c.createOnosConfigDeployment(); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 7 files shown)


kuujo:
Probably should also extend this a bit to support port forwarding in onit:
```
> onit debug 1
```

test/runner/onos-config.go
```diff
@@ -334,12 +349,34 @@ func (c *ClusterController) createOnosConfigService() error {
...
+			LabelSelector: "app=onos-config",
+		})
+		if err != nil {
+			return err
+		}
+
+		// Iterate through the pods in the deployment and unblock the debugger
+		for _, pod := range pods.Items {
+			if _, ok := unblocked[pod.Name]; !ok && pod.Status.ContainerStatuses[0].State.Running != nil {
+				_, _, err := c.execute(pod, []string{"/bin/bash", "-c", "dlv --init <(echo \"exit -c\") connect 127.0.0.1:40000"})
```
tomikazi: Hack or diabolically clever?! That's a tough call.

kuujo:
Added an `onit debug` command to forward a local port to the debugger on an onos-config node:
```bash
> onit get nodes
ID                             STATUS
onos-config-65988654df-gkb6b   RUNNING
> onit debug onos-config-65988654df-gkb6b
Forwarding from 127.0.0.1:40000 -> 40000
Forwarding from [::1]:40000 -> 40000
```

PR #421, opened 2019-06-26
Fix problems reported by gofmt and golint in the integration tests

ray-milkey:


kuujo:
This is great... but we're going to have some merge conflicts here from #409 for sure. They're unavoidable either way. Going to have to merge that one and then rebase.

PR #436, opened 2019-06-27
Southbound error channel

### LINKED ISSUE #135: Extend synchronizer to handle get/set responses

Andrea-Campanella:
WIP: Adding an error channel for the southbound and listening on the manager. Fixes #135

The Idea behind this work is that the Manager will then act upon the error accordingly, bringing the system back in an appropriate state. 
E.g. The device rejects a change the Manager needs to rollback that change in the store for that device and also rollback the network change it was part of.

### DIFF
--- a/pkg/events/errorevent.go
+++ b/pkg/events/errorevent.go
@@ -1,0 +1,53 @@
+// Copyright 2019-present Open Networking Foundation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package events
+
+import (
+	"encoding/base64"
+	"fmt"
+	"github.com/onosproject/onos-config/pkg/store/change"
+	"time"
+)
+
+// ErrorEvent a error event
+type ErrorEvent Event
+
+// ChangeID returns the changeId of the error event
+func (errEvent *ErrorEvent) ChangeID() string {
+	return errEvent.values[ChangeID]
+}
+
+// EventType returns the EventType of the error event
+func (errEvent *ErrorEvent) EventType() EventType {
+	return errEvent.eventtype
+}
+
+// Error returns the error of the error event
+func (errEvent *ErrorEvent) Error() error {
+	return fmt.Errorf(errEvent.values[Error])
+}
+
+// CreateErrorEvent creates a new error event object
+func CreateErrorEvent(eventType EventType, subject string, changeID change.ID, err error) ErrorEvent {
+	values := make(map[string]string)
+	values[ChangeID] = base64.StdEncoding.EncodeToString(changeID)
+	values[Error] = string(err.Error())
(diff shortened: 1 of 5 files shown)


pkg/southbound/synchronizer/synchronizer.go
```diff
@@ -86,14 +90,14 @@ func New(context context.Context, changeStore *store.ChangeStore, configStore *s
 
 		if err != nil {
 			log.Error("Can't translate the initial config for ", sync.Device.Addr, err)
-			return nil, err
+			return sync, nil
```
SeanCondon: Would we ignore the error in every case?

pkg/southbound/synchronizer/synchronizer.go
```diff
@@ -86,14 +90,14 @@ func New(context context.Context, changeStore *store.ChangeStore, configStore *s
```
SeanCondon: At 2 lines up - should this be **if config == nil** and the **err != nil** needs to be handled

pkg/southbound/synchronizer/synchronizer.go
```diff
@@ -118,45 +123,55 @@ func New(context context.Context, changeStore *store.ChangeStore, configStore *s
...
+func (sync Synchronizer) syncOperationalState(errChan chan<- events.ErrorEvent) {
 
 	target, err := southbound.GetTarget(sync.key)
 
 	if err != nil {
 		log.Error("Can't find target for key ", sync.key)
-		return err
+		errChan <- events.CreateErrorEvent(events.EventTypeErrorDeviceConnect,
+			sync.key.DeviceID, make([]byte, 0), err)
+		return
```
SeanCondon: this means that this go routine is dies at this point? Probably OK.

pkg/southbound/synchronizer/synchronizer.go
```diff
@@ -223,7 +238,7 @@ func (sync Synchronizer) syncOperationalState() error {
 	if len(subscribePaths) == 0 {
 		noPathErr := fmt.Errorf("target %#v has no paths to subscribe to", sync.ID)
 		log.Warning(noPathErr)
-		return noPathErr
+		return
```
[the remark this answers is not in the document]
Andrea-Campanella: I don't think this is worth sending to the channel because we are doing no subscription to the device at this point in time. This means that nothing will be pulled from it in terms of STATE or OPERATIONAL values. Also I'm not sending up the error because I think there is really nothing we can do here. We are trying the get STATE and OPERATIONAL, if that fails we use the yang models (YET NOT DONE), if even that fails we have exhausted our options.

pkg/southbound/synchronizer/synchronizer.go
Andrea-Campanella: I expect this error never to happen, it's config we stored in the enos-config stores and should have been checked before, I'm still doing the check because I don't like ignoring error altogether thus the log is there but it's an error that we should never hit here.

pkg/events/errorevent.go
```diff
@@ -0,0 +1,53 @@
...
+	return errEvent.eventtype
+}
+
+// Error returns the error of the error event
+func (errEvent *ErrorEvent) Error() error {
+	return fmt.Errorf(errEvent.values[Error])
+}
+
+// CreateErrorEvent creates a new error event object
+func CreateErrorEvent(eventType EventType, subject string, changeID change.ID, err error) ErrorEvent {
```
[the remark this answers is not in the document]
Andrea-Campanella: done

pkg/southbound/synchronizer/synchronizer.go
Andrea-Campanella: yes

PR #439, opened 2019-06-27
Notification for device when administratively added

Andrea-Campanella:
When a device is added through admin tools we need to send out a notification for it on the topoChannel.


kuujo:
This has passed but the PR has not been updated for some reason. Going to merge it

Issue #395, opened 2019-06-24
Support grouping integration tests


kuujo:
Currently, the `onit` command supports either running a set of specific tests by listing the test names or all tests by excluding any specific test names. In vanilla ONOS, we were able to run a collection of tests to verify the system via commands like `stc smoke`. As we add more system tests, we’re going to want the ability to run tests in groups and should add support for grouping tests. `onit` should then support running groups of tears the same way it currently supports single tests:
```
onit run smoke
```

PR #450, opened 2019-06-27
Support grouping of integration tests

antonjlin:
https://github.com/onosproject/onos-config/issues/395

### DIFF
--- a/build/onos-it/Dockerfile
+++ b/build/onos-it/Dockerfile
@@ -4,8 +4,8 @@ FROM onosproject/onos-config-base:$ONOS_CONFIG_BASE_VERSION as base
 
 FROM alpine:3.8
 
-COPY --from=base /go/src/github.com/onosproject/onos-config/build/_output/onit /usr/local/bin/onit
+COPY --from=base /go/src/github.com/onosproject/onos-config/build/_output/onit-k8s /usr/local/bin/onit-k8s
 
 USER nobody
 
-ENTRYPOINT ["onit", "test"]
+ENTRYPOINT ["onit-k8s"]
--- a/test/cli/cli.go
+++ b/test/cli/cli.go
@@ -53,6 +53,7 @@ func GetOnitCommand(registry *runner.TestRegistry) *cobra.Command {
 	return cmd
 }
 
+// GetOnitK8sCommand returns a Cobra command for running tests on k8s
 func GetOnitK8sCommand(registry *runner.TestRegistry) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:                    "onit-k8s",
--- a/test/cli/cli.go
+++ b/test/cli/cli.go
@@ -286,7 +286,7 @@ func getRemoveSimulatorCommand() *cobra.Command {
 // getGetCommand returns a cobra "get" command to read test configurations
 func getGetCommand(registry *runner.TestRegistry) *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "get {cluster,clusters,simulators,device-presets,store-presets,tests,testsuites}",
+		Use:   "get {cluster,clusters,simulators,device-presets,store-presets,tests,suites}",
 		Short: "Get test configurations",
 	}
 	cmd.AddCommand(getGetClusterCommand())
@@ -593,14 +593,6 @@ func getGetTestSuitesCommand(registry *runner.TestRegistry) *cobra.Command {
 		Short: "Get a list of integration testing suites",
 		Run: func(cmd *cobra.Command, args []string) {
 			registry.PrintTestSuites()
-			//for _, name := range registry.GetTestSuiteNames() {
-			//	desc := name + ": "
-			//	suite := registry.TestSuites[name]
-			//	for _, name := range suite.GetTestNames() {
-			//		desc += "\n > " + name
-			//	}
-			//	fmt.Println(desc)
-			//}
 		},
 	}
 }
@@ -951,14 +943,14 @@ func getSetClusterCommand() *cobra.Command {
 	}
 }
 
-// getGetCommand returns a cobra "get" command to read test configurations
+// getRunCommand returns a cobra "get" command to read test configurations
 func getRunCommand(registry *runner.TestRegistry) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "run {test,suite}",
 		Short: "Run integration tests",
 	}
 	cmd.AddCommand(getRunTestSuiteCommand(registry))
-	cmd.AddCommand(getRunTestCommand())
+	cmd.AddCommand(getRetunTestCommand())
 	return cmd
 }
 
--- a/test/cli/cli.go
+++ b/test/cli/cli.go
@@ -32,8 +32,8 @@ import (
 
 )
 
-// GetCommand returns a Cobra command for tests in the given test registry
-func GetCommand(registry *runner.TestRegistry) *cobra.Command {
+// GetOnitCommand returns a Cobra command for tests in the given test registry
+func GetOnitCommand(registry *runner.TestRegistry) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:                    "onit",
 		Short:                  "Run onos-config integration tests on Kubernetes",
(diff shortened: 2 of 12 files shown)


antonjlin:
Will update documentation after merged into master

ray-milkey:
This is great! Minor point - other tools like junit call groups of tests 'suites'. That might be a more descriptive name to use.

test/runner/runner.go
```diff
@@ -26,22 +26,58 @@ import (
...
-// Register registers a named test
+//TestGroup to run multiple tests
+type TestGroup struct {
+	name string
+	suite.SetupTestSuite
+	tests map[string]Test
+}
+
+// AddToGroup registers a named test
+func (r *TestRegistry) AddToGroup(testName string, testGroup *TestGroup, test Test) {
```
kuujo: This seems like it would make more sense as a receiver on `TestGroup` rather than `TestRegistry` since it's actually modifying the state of the group.

kuujo: Also, there's a bit of an inconsistency with naming here. We should either use `Add` or `Register`, not both. 
It also may not make sense for `AddToGroup` to take a test name here when the test is already providing a name for `Register`. We need an API that ensures at least that:
* Tests that are added to a group are also added to the general registry so they can be run by name
* Tests cannot be added to the global registry and the group registry with different names

kuujo: Nevermind... I see there _is_ a `Register` receiver on `TestGroup`. We should just remove the `AddToGroup` version as it's redundant.

test/runner/runner.go
```diff
@@ -26,22 +26,58 @@ import (
...
 	suite.SetupTestSuite
 	tests map[string]Test
+	testGroups map[string]TestGroup
 }
 
-// Register registers a named test
+//TestGroup to run multiple tests
+type TestGroup struct {
+	name string
+	suite.SetupTestSuite
```
kuujo: I actually don't think the use of `suite` is needed. We should remove the usage of these interfaces as we're not even running them as test suites.

kuujo:
Need to be rebased sowwy :-(

antonjlin:
Rebased and fixed casing bug.

kuujo:
This doesn't seem to be rebasing cleanly for some reason. The code review shows a lot of code from commits that have been merged. Not entirely sure why that's happening, but we have to figure out how to clean that up to make sure we're only reviewing the changes.

kuujo:
I think the `onit get suites` command should just print a list of suite names, and the `onit get tests` command should print a table of tests/suites, e.g.:
```
TEST                 SUITE
single-path       alltests,sometests
subscribe          alltests,sometests
...
```
This is not only because it's consistent with the output of other commands, but because we actually need to be able to easily parse the output of `onit get suites` for completion.

kuujo:
BTW we do need to add tests and suites to autocompletion, but you should do that separately.

antonjlin:
> I think the `onit get suites` command should just print a list of suite names, and the `onit get tests` command should print a table of tests/suites, e.g.:
> 
> ```
> TEST               SUITE
> single-path        alltests,sometests
> subscribe          alltests,sometests
> ...
> ```
> 
> This is not only because it's consistent with the output of other commands, but because we actually need to be able to easily parse the output of `onit get suites` for completion.
> 
> Or, alternatively, they could both print inverse tables since the set of tests in the suite is also useful. We just need to preserve the columnar format for autocompletion.

Would it be more intuitive to have `onit get suites` have a table and `onit get tests` just list in the current format? Since suites contain tests but tests don't contain suites, it makes it harder both in the code to display them and also kind of confusing for the reader.

kuujo:
Whatever you think works!

test/cli/cli.go
```diff
@@ -282,7 +284,7 @@ func getRemoveSimulatorCommand() *cobra.Command {
 // getGetCommand returns a cobra "get" command to read test configurations
 func getGetCommand(registry *runner.TestRegistry) *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "get {cluster,clusters,simulators,device-presets,store-presets,tests}",
+		Use:   "get {cluster,clusters,simulators,device-presets,store-presets,tests,testsuites}",
```
kuujo: `testsuites`->`suites`

test/cli/cli.go
```diff
@@ -574,13 +577,32 @@ func getGetTestsCommand(registry *runner.TestRegistry) *cobra.Command {
...
 }
 
+// getGetTestsCommand returns a cobra command to get a list of available tests
+func getGetTestSuitesCommand(registry *runner.TestRegistry) *cobra.Command {
+	return &cobra.Command{
+		Use:   "suites",
+		Short: "Get a list of integration testing suites",
+		Run: func(cmd *cobra.Command, args []string) {
+			registry.PrintTestSuites()
+			//for _, name := range registry.GetTestSuiteNames() {
```
kuujo: Remove commented code

antonjlin:
Oops, need to rebase first.

test/cli/cli.go
```diff
@@ -16,23 +16,24 @@ package cli
...
+	"github.com/onosproject/onos-config/test/console"
+	"github.com/onosproject/onos-config/test/runner"
+	"github.com/spf13/cobra"
 	corev1 "k8s.io/api/core/v1"
 
 )
 
-// GetCommand returns a Cobra command for tests in the given test registry
-func GetCommand(registry *runner.TestRegistry) *cobra.Command {
+// GetOnitCommand returns a Cobra command for tests in the given test registry
```
Andrea-Campanella: maybe I'm missing the point, but what is the difference between onit and onit-k8s ?

test/cli/cli.go
```diff
@@ -47,8 +48,19 @@ func GetCommand(registry *runner.TestRegistry) *cobra.Command {
...
 	cmd.AddCommand(getCompletionCommand())
+
+	return cmd
+}
+
+// GetOnitK8sCommand returns a Cobra command for running tests on k8s
+func GetOnitK8sCommand(registry *runner.TestRegistry) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:                    "onit-k8s",
+		Short:                  "ONLY TO BE USED ON K8S INSTANCE",
```
Andrea-Campanella: please avoid all caps

test/cli/cli.go
```diff
@@ -929,53 +952,88 @@ func getSetClusterCommand() *cobra.Command {
 	}
 }
 
-// getRunCommand returns a cobra "run" command
+// getRunCommand returns a cobra "get" command to read test configurations
```
Andrea-Campanella: is this a run or a get ? the documentation is confusing against line 958

test/cli/cli.go
```diff
@@ -929,53 +952,88 @@ func getSetClusterCommand() *cobra.Command {
...
+	cmd.Flags().IntP("timeout", "t", 60*10, "test timeout in seconds")
+	return cmd
+}
 
+func getRunSuiteRemoteCommand(registry *runner.TestRegistry) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "suite [suite]",
+		Short: "Run integration tests",
+		Run: func(cmd *cobra.Command, args []string) {
+			//tests := registry.TestSuites[args[0]]
```
Andrea-Campanella: remove commented out code if not needed

test/integration/registry.go
```diff
@@ -14,9 +14,23 @@
...
+import (
+	"github.com/onosproject/onos-config/test/runner"
+)
 
 var (
 	// Registry holds the registered test names
 	Registry = runner.NewRegistry()
+
+	//AllTests is an example of a test suite
+	AllTests = runner.NewTestSuite("alltests")
```
Andrea-Campanella: we might want to define these groupings in a separate file.

test/integration/registry.go
```diff
@@ -14,9 +14,23 @@
...
+	//SomeTests is an example of a test suite
+	SomeTests = runner.NewTestSuite("sometests")
```
Andrea-Campanella: also we are not adding any test to these test suites, am I correct ?

build/onos-it/Dockerfile
```diff
@@ -4,8 +4,8 @@ FROM onosproject/onos-config-base:$ONOS_CONFIG_BASE_VERSION as base
 
 FROM alpine:3.8
 
-COPY --from=base /go/src/github.com/onosproject/onos-config/build/_output/onit /usr/local/bin/onit
+COPY --from=base /go/src/github.com/onosproject/onos-config/build/_output/onit-k8s /usr/local/bin/onit-k8s
 
 USER nobody
 
-ENTRYPOINT ["onit", "test"]
```
Andrea-Campanella: why remove "test"

test/cli/cli.go
antonjlin: I talked this over with @kuujo -- Tests are presumably written for integration and with the assumption devices are also running in the same cluster, so running locally doesn't really make sense. They have to be run in k8s, so we should really probably just add a separate command to run the actual test functions. We split up `onit` is user facing and `onit-k8s` should only be called by onit to run in k8s.

build/onos-it/Dockerfile
antonjlin: onit-k8s has two options "test" or "suite" that trigger either a single test or a suite. No need to put the command one more "level" down since it's its own command now and would not interfere with any onit commands

kuujo:
@antonjlin looks good! Just some whitespace to be removed... do we need to address some of @Andrea-Campanella's comments?

test/integration/registry.go
antonjlin: correct

PR #453, opened 2019-06-28
Run integration tests as part of Travis CI

ray-milkey:


Andrea-Campanella:
after rebasing this patch on #454 
as you can see in #456 the test runs have been flawless for 7 times in a raw with an average time between 12 and 13 mins.

Issue #431, opened 2019-06-27
Add options for controlling log output to "onit get logs"


kuujo:
The Kubernetes API supports options for reading specific portions of the application logs, the last n lines, logs starting at time t, etc. These options should be exposed in the `onit get logs` command as well.

PR #464, opened 2019-06-28
added since and tail log queries

antonjlin:


kuujo:
Needs to be rebased

Andrea-Campanella:
@antonjlin this needs rebasing. Also travis did not trigger for some reason, if the rebase does not trigger it we can look into it.

kuujo:
#431

test/cli/cli.go
```diff
@@ -699,6 +708,12 @@ To output the logs from a test, get the test ID from the test run or from 'onit
 		},
 	}
 
+	cmd.Flags().Int64P("since", "", -1, "Only return logs newer than a relative " +
```
kuujo: Cobra supports a `Duration` type that may actually parse string input like `5s` or `10m` like the comment suggests? Currently this only takes a numeric input right?

antonjlin: Added in the latest commit.

test/cli/cli.go
```diff
@@ -707,6 +722,26 @@ To output the logs from a test, get the test ID from the test run or from 'onit
...
+		options.SinceSeconds = &sinceSeconds
+	}
+	if err != nil {
+		exitError(err)
+	}
+	tail, err := cmd.Flags().GetInt64("tail")
+	if tail > 0 {
+		options.TailLines = &tail
+	}
+	fmt.Println(options)
```
kuujo: Need to remove the options before this is merged

Issue #461, opened 2019-06-28
Stratum Integration testing


Andrea-Campanella:
**Is your feature request related to a problem? Please describe.**
Currently Testing are done based on the simulator. 

**Describe the solution you'd like**
We would need the capability to add a Stratum device to onos-config cluster and run tests with one or more paths supported by stratum,

PR #465, opened 2019-07-01
Stratum integration testing (#461) 

adibrastegarnia:
This PR integrates stratum image with `onit`. For the implementation, we should keep `add`/`remove` `simulator` and add `add`/`remove` `network`. The two need to be handled pretty differently. The Mininet pod needs to be added with those ports exposed, and then we should add a _separate_ service per device so the devices each have a separate host. That will preserve the separate host:port pattern for devices. Then they all have to be added to the stores. So we should create a separate `network.go` and new subcommand for the Mininet stuff

This PR supports a single node and linear topologies for now. The reason for is that we need to know how many devices will be emulated that we can create a Kubernetes service for each node in the network and expose its port.

### DIFF
--- a/test/cli/cli.go
+++ b/test/cli/cli.go
@@ -23,14 +23,24 @@ import (
 	"strings"
 	"text/tabwriter"
 	"time"
+	"errors"
 
 	"github.com/google/uuid"
 	"github.com/onosproject/onos-config/test/console"
 	"github.com/onosproject/onos-config/test/runner"
 	"github.com/spf13/cobra"
 	corev1 "k8s.io/api/core/v1"
-
 )
+
+// Contains tells whether array contains x.
+func Contains(array []string, elem string) bool {
+	for _, n := range array {
+		if elem == n {
+			return true
+		}
+	}
+	return false
+}
 
 // GetOnitCommand returns a Cobra command for tests in the given test registry
 func GetOnitCommand(registry *runner.TestRegistry) *cobra.Command {
@@ -56,8 +66,8 @@ func GetOnitCommand(registry *runner.TestRegistry) *cobra.Command {
 // GetOnitK8sCommand returns a Cobra command for running tests on k8s
 func GetOnitK8sCommand(registry *runner.TestRegistry) *cobra.Command {
 	cmd := &cobra.Command{
-		Use:                    "onit-k8s",
-		Short:                  "This command is only intended to be used in a k8s instance for running integration tests.",
+		Use:   "onit-k8s",
+		Short: "This command is only intended to be used in a k8s instance for running integration tests.",
 	}
 	cmd.AddCommand(getTestTestLocalCommand(registry))
 	cmd.AddCommand(getTestSuiteLocalCommand(registry))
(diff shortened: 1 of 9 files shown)


test/cli/cli.go
```diff
@@ -131,16 +133,23 @@ func getAddCommand() *cobra.Command {
 // getAddSimulatorCommand returns a cobra command for deploying a device simulator
 func getAddSimulatorCommand() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "simulator [name]",
+		Use:   "simulator [type] [name]",
```
Andrea-Campanella: can this type be placed in the SimulatorConfig object ? thus removing many changes from the API ? If instead this breaks the point of SimulatorConfig I would move the type string to the end of the constructor arguments for consistency reasons and to leave the device name as first element. cosmetic, but enhances readability

test/runner/simulator.go
```diff
@@ -79,91 +82,187 @@ func (c *ClusterController) createSimulatorConfigMap(name string, config *Simula
...
-			Volumes: []corev1.Volume{
-				{
-					Name: "config",
-					VolumeSource: corev1.VolumeSource{
-						ConfigMap: &corev1.ConfigMapVolumeSource{
-							LocalObjectReference: corev1.LocalObjectReference{
-								Name: name,
+		}
+
+	} else if strings.Compare(simType, "stratum") == 0 {
```
Andrea-Campanella: define `"onos"` and `"stratum"` as constants

test/runner/simulator.go
```diff
@@ -79,91 +82,187 @@ func (c *ClusterController) createSimulatorConfigMap(name string, config *Simula
...
-			Name:      name,
-			Namespace: c.clusterID,
-		},
-		Spec: corev1.ServiceSpec{
-			Selector: map[string]string{
-				"simulator": name,
+func (c *ClusterController) createSimulatorService(simType string, name string) error {
+
+	service := &corev1.Service{}
+	if strings.Compare(simType, "onos") == 0 {
```
Andrea-Campanella: reuse constants here

test/cli/cli.go
tomikazi: Agreed.

adibrastegarnia: agreed. Will do.

test/runner/simulator.go
adibrastegarnia: will do.

test/runner/simulator.go
adibrastegarnia: will do.

test/cli/cli.go
kuujo: Should this be called `device` instead of `simulator`? Or we could do:
```
onit add simulator ...
onit add stratum ...
```

Or the type can be a flag with a default value:
```
onit add device --type=simulator my-device
onit add device --type=stratum my-device 
```

Just change the `simulator` command to `device`, add an alias to `simulator` to avoid breakage, and default `type` to `simulator` or `onos` or whatever it should be.

adibrastegarnia: OK. I think the second one is better (i.e. onit add device .....) . I am working on it.

kuujo: agree

Andrea-Campanella: I like the device better, indeed

kuujo:
We need to update all the commands:
```
onit delete device ...
onit get devices ...
```
As well as the autocompletion code for devices.

test/cli/cli.go
adibrastegarnia: > We need to update all the commands:
> 
> ```
> onit delete device ...
> onit get devices ...
> ```
> 
> As well as the autocompletion code for devices.

OK. I will take care of them.

test/cli/cli.go
```diff
@@ -127,6 +127,69 @@ func getAddCommand() *cobra.Command {
 		Short: "Add resources to the cluster",
 	}
 	cmd.AddCommand(getAddSimulatorCommand())
+	cmd.AddCommand(getAddNetworkCommand())
+	return cmd
+}
+
+// getAddNetworkCommand returns a cobra command for deploying a stratum network
+func getAddNetworkCommand() *cobra.Command {
```
Andrea-Campanella: I know the current image deploys a 3 node instance of Stratum and I'm ok with that. Could you work with Carmelo and see if we can have a docker image of a single stratum device with no mininet added ?

adibrastegarnia: @Andrea-Campanella 
The idea is to allow the user to configure the number of nodes in the network when it sends a request to create a network pod. I mean it can be a single node or multiple nodes. But if you are looking for a completely standalone version of stratum to be executed outside mininet, I will check with Carmelo to see can we have one.

Andrea-Campanella: @adibrastegarnia as per our discussion yesterday the network solution is fine. Parametrize the mininet launch script with the number of stratum nodes passed from the user.

adibrastegarnia:
> Can you add documentation to the .md file on how to use this wit two examples ?
> 
>     1. single node in mininet
> 
>     2. 3 node network in mininet

@Andrea-Campanella 
My plan is to do that but it still needs work (I will do that at the end). I need to add each of the devices in the network to the onos-config and also make sure when we remove a network all of the services that we created will be deleted.

Andrea-Campanella:
> > Can you add documentation to the .md file on how to use this wit two examples ?
> > ```
> > 1. single node in mininet
> > 
> > 2. 3 node network in mininet
> > ```
> 
> @Andrea-Campanella
> My plan is to do that but it still needs work (I will do that at the end). I need to add each of the devices in the network to the onos-config and also make sure when we remove a network all of the services that we created will be deleted.

@adibrastegarnia Indeed. I was wondering what services you mention, it's s self included docker image which hosts mininet. I do not think something else rather than killing the docker image and removing the devs from onos-config store is necessary.

adibrastegarnia:
> > > Can you add documentation to the .md file on how to use this wit two examples ?
> > > ```
> > > 1. single node in mininet
> > > 
> > > 2. 3 node network in mininet
> > > ```
> > 
> > 
> > @Andrea-Campanella
> > My plan is to do that but it still needs work (I will do that at the end). I need to add each of the devices in the network to the onos-config and also make sure when we remove a network all of the services that we created will be deleted.
> 
> @adibrastegarnia Indeed. I was wondering what services you mention, it's s self included docker image which hosts mininet. I do not think something else rather than killing the docker image and removing the devs from onos-config store is necessary.

@Andrea-Campanella 
I mean services for the POD. For each switch there is a unique gRPC port that we should expose. We create a service for each and expose the port. For the device simulator, there is one service and we expose one port.

adibrastegarnia:
@kuujo @Andrea-Campanella 

This PR is ready for review. I also added documents which shows the usage of commands for creating of a network. Adding tests should be in a separate PR since I should talk to Ray about a client without TLS for now.

Andrea-Campanella:
@adibrastegarnia 
works on my environment! Great work! 
Regarding the client with no TLS I'm maybe missing the point. `onos-config` does support connecting through stratum by setting the `Plain` flag to `true` on the device. Setting that flag tells `onos-config` not to establish the connection in a secure way. Once the connection is established the `stratum` device appears to `onos-config` as just yet another device  flag is set the normal `gnmi_cli` commands can be issue to the NB of `onos-config`. There is no need for any other special client.

adibrastegarnia:
> @adibrastegarnia
> works on my environment! Great work!
> Regarding the client with no TLS I'm maybe missing the point. `onos-config` does support connecting through stratum by setting the `Plain` flag to `true` on the device. Setting that flag tells `onos-config` not to establish the connection in a secure way. Once the connection is established the `stratum` device appears to `onos-config` as just yet another device flag is set the normal `gnmi_cli` commands can be issue to the NB of `onos-config`. There is no need for any other special client.

@Andrea-Campanella 
Thanks for checking the PR. I didn't know about that option. I will check with you offline.

PR #472, opened 2019-07-02
Test & suites bash completion

antonjlin:


test/runner/runner.go
```diff
@@ -114,10 +114,12 @@ func (r *TestRegistry) GetTestSuiteNames() []string {
 }
 
 //PrintTestSuites prints test suites in a table
-func (r *TestRegistry) PrintTestSuites() {
+func (r *TestRegistry) PrintTestSuites(noHeaders bool) {
```
kuujo: We should probably make the signature consistent with the other printers that support this option:
```
func printTestSuites(includeHeaders bool)
```
Also, I wonder if we should move this `print` function over to `cli.go`. I'd like to keep all the output there. It doesn't really belong in the runner. Printing the tests is exclusively a CLI function.

antonjlin: fixed in latest update

PR #477, opened 2019-07-03
Distributed device store

kuujo:
This PR changes the device store to use the Atomix client to store devices. This PR is only a test/demonstration of a distributed store working in k8s, and the device store was chosen only because it was simplest. 

We should consider what we want to do with the device store soon.


kuujo:
Abandoning this for onosproject/onos-topo

Issue #481, opened 2019-07-04
onit cannot create cluster on Ubuntu (microk8s) - onos-config-debug fails


SeanCondon:
**Describe the bug**
Running **onit create cluster** on Ubuntu with microk8s fails because running the docker image **onos-config-debug** without **--security-opt seccomp:unconfined** means that **onos-config-debug** can't start

**To Reproduce**
After running **make images** on Ubuntu try running the docker image **onosproject/onos-config:debug** like, and watch it fail:
```
$ docker run -it onosproject/onos-config:debug 
API server listening at: [::]:40000
could not launch process: fork/exec /usr/local/bin/onos-config: operation not permitted
$
```

**Expected behavior**
Now try running it with:
```
$ docker run -it --security-opt seccomp:unconfined onosproject/onos-config:debug 
API server listening at: [::]:40000
```
and see how it runs properly.

This should also be applied to how **onit** runs this Docker image

**Additional context**
In **createOnosConfigDeployment()** of **test/runner/onos-config.go**, the PodSecurityPolicy should have the annotation added like:
https://kubernetes.io/docs/concepts/policy/pod-security-policy/#seccomp

kuujo:
The k8s documentation suggests `seccomp:unconfined` is the default:
https://kubernetes.io/docs/concepts/policy/pod-security-policy/#seccomp

kuujo:
Maybe it's the default only if a `PodSecurityPolicy` is added... but that requires non-trivial setup to the cluster itself to enable support for `PodSecurityPolicy`s

kuujo:
The above PR should fix the issue. A `PodSecurityPolicy` doesn't seem the right way to specify security options on a pod because `PodSecurityPolicy` is designed as a filter to prevent insecure pods from being scheduled. The original proposal documents showed instead the `seccomp.security.alpha.kubernetes.io/pod` annotation which should allow the profile to be set for the pod.

@SeanCondon what k8s version are you using? `kubectl version`

SeanCondon:
1.15.0
```
scondon@Z420:/RaidData/scondon/onos$ kubectl version
Client Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.0", GitCommit:"e8462b5b5dc2584fdcd18e6bcfe9f1e4d970a529", GitTreeState:"clean", BuildDate:"2019-06-19T16:40:16Z", GoVersion:"go1.12.5", Compiler:"gc", Platform:"linux/amd64"}
[...]

PR #484, opened 2019-07-07
Apply unconfined seccomp profile to debug containers run by onit

kuujo:
#481 

This PR applies the `unconfined` seccomp profile to debug containers to ensure they can be run with microk8s. I'm not entirely sure this will work, but based on the original proposals for seccomp support it is the best attempt I could come up with.


kuujo:
Tried running the debug process as root ☝️

PR #489, opened 2019-07-08
Rollback method and usage

### LINKED ISSUE #460: Set config on device error handling
Currently we do not handle any error in setting the configuration.

A set error should be retrieved from the error channel #135 and handled accordingly. Probably with a rollback of both device config for failed device, other devices involved in network config (if any) and of network config.

Andrea-Campanella:
Introduce a new rollback method on the Manager to enable:
- remove of last change in a configuration for a given target
- delta creation against configuration previous to the removed change
- (re)set of configuration on device according to rollback states

fixes #460

### DIFF
--- a/pkg/manager/manager_test.go
+++ b/pkg/manager/manager_test.go
@@ -356,33 +356,6 @@ func TestManager_GetManager(t *testing.T) {
 	assert.Equal(t, mgrTest, GetManager())
 }
 
-func TestManager_ComputeRollback(t *testing.T) {
-	mgrTest, _, _ := setUp()
-
-	updates := make(map[string]string)
-	deletes := make([]string, 0)
-
-	updates[Test1Cont1ACont2ALeaf2B] = ValueLeaf2B159
-
-	_, _, err := mgrTest.SetNetworkConfig("Device1-1.0.0", updates, deletes)
-
-	assert.NilError(t, err, "Can't create change")
-
-	updates[Test1Cont1ACont2ALeaf2B] = ValueLeaf2B314
-
-	changeID, configName, err := mgrTest.SetNetworkConfig("Device1-1.0.0", updates, deletes)
-
-	assert.NilError(t, err, "Can't create change")
-
-	changes, deletes, errRoll := computeRollback(mgrTest, "Device1", string(configName))
-	assert.Check(t, len(deletes) == 0, "no path should be deleted")
-	assert.NilError(t, errRoll, "Can't ExecuteRollback")
-	config := mgrTest.ConfigStore.Store[configName]
-	assert.Check(t, !bytes.Equal(config.Changes[len(config.Changes)-1], changeID), "Did not remove last change")
-	assert.Check(t, changes[Test1Cont1ACont2ALeaf2B] == ValueLeaf2B159, "Wrong value to set after rollback")
-
-}
-
 func TestManager_ComputeRollbackDelete(t *testing.T) {
 	mgrTest, _, _ := setUp()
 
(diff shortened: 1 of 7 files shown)


pkg/manager/manager_test.go
```diff
@@ -351,3 +355,60 @@ func TestManager_GetManager(t *testing.T) {
...
+	changes, deletes, errRoll := computeRollback(mgrTest, "Device1", string(configName))
+	assert.Check(t, len(deletes) == 0, "no path should be deleted")
+	assert.NilError(t, errRoll, "Can't ExecuteRollback")
+	config := mgrTest.ConfigStore.Store[configName]
+	assert.Check(t, !bytes.Equal(config.Changes[len(config.Changes)-1], changeID), "Did not remove last change")
+	assert.Check(t, changes[Test1Cont1ACont2ALeaf2B] == ValueLeaf2B159, "Wrong value to set after rollback")
+
+}
+
+func TestManager_ComputeRollbackDelete(t *testing.T) {
```
SeanCondon: this is like a superset of the previous one - is the previous one needed at all?

pkg/manager/rollbackconfig.go
```diff
@@ -0,0 +1,64 @@
...
+	if err != nil {
+		return err
+	}
+	_, _, errSet := m.SetNetworkConfig(store.ConfigName(configname), updates, deletes)
+	//TODO if error we might want to take further action
+	return errSet
+}
+
+func computeRollback(m *Manager, target string, configname string) (map[string]string, []string, error) {
+	id, err := m.ConfigStore.RemoveLastChangeEntry(store.ConfigName(configname))
```
SeanCondon: I would like if the ID was returned at the end - just for confirmation (from client side) it would be good to know which change was rolled back

pkg/manager/rollbackconfig.go
```diff
@@ -0,0 +1,64 @@
...
+import (
+	"fmt"
+	"github.com/onosproject/onos-config/pkg/store"
+	"github.com/onosproject/onos-config/pkg/store/change"
+	log "k8s.io/klog"
+)
+
+// RollbackTargetConfig rollbacks the last change for a given configuration on the target. Only the last one is
+// restored to the previous state. Going back n changes in time requires n sequential calls of this method.
+func (m *Manager) RollbackTargetConfig(configname string) error {
```
SeanCondon: should pass back the change ID I think

pkg/northbound/admin/admin.go
```diff
@@ -137,18 +137,17 @@ func (s Server) RollbackNetworkChange(
...
+		if store.B64(configChangeIds[len(configChangeIds)-1]) != store.B64(changeID) {
 			return nil, fmt.Errorf(
 				"the last change on %s is not %s as expected. Was %s",
-				k, v, store.B64(configChangeIds[len(configChangeIds)-1]))
+				configName, changeID, store.B64(configChangeIds[len(configChangeIds)-1]))
+		}
+		err := manager.GetManager().RollbackTargetConfig(string(configName))
+		if err != nil {
+			return nil, err
 		}
```
SeanCondon: maybe log here the list of targets and their changes - if we could fit them in to the response that would be better (although would require proto change - can be done later)

pkg/manager/rollbackconfig.go

pkg/manager/manager_test.go
Andrea-Campanella: simplicity of read I guess, but strictly speaking no

pkg/manager/rollbackconfig.go

pkg/northbound/admin/admin.go
Andrea-Campanella: done, I've added them to the message of the response for now

PR #491, opened 2019-07-08
Changed core values to used strongly typed rather than string values

SeanCondon:


pkg/northbound/gnmi/set.go
```diff
@@ -55,13 +56,13 @@ func (s *Server) Set(ctx context.Context, req *gnmi.SetRequest) (*gnmi.SetRespon
 	//Update
 	for _, u := range req.GetUpdate() {
 		target := u.Path.GetTarget()
-		targetUpdates[target] = s.doUpdateOrReplace(u, targetUpdates)
+		targetUpdates[target], err = s.doUpdateOrReplace(u, targetUpdates)
```
Andrea-Campanella: the error here assigned bt not checked until line 87, effectively letting `onos-config` try to apply a partial state/config. I think we should check for errors at line 73. and if any return. 
```	
if err != nil {
		return nil, status.Error(codes.Internal, err.Error())
}
```

pkg/northbound/gnmi/subscribe.go
```diff
@@ -299,6 +308,9 @@ func sendResponse(response *gnmi.SubscribeResponse, stream gnmi.GNMI_SubscribeSe
 func getChangeFromEvent(update events.ConfigEvent, mgr *manager.Manager) (string, *change.Change) {
 	target := events.Event(update).Subject()
 	changeID := update.ChangeID()
-	changeInternal := mgr.ChangeStore.Store[changeID]
+	changeInternal, ok := mgr.ChangeStore.Store[changeID]
+	if !ok {
+		log.Warning("No change found for: ", changeID)
```
Andrea-Campanella: the change will be nil, and if nil we will get invalid pointer. should we check where this is called that if change == nil we skip the event ?

pkg/store/change/change.go
```diff
@@ -121,11 +85,11 @@ func CreateChange(config ValueCollections, desc string) (*Change, error) {
 	var pathList = make([]string, len(config))
 	// If a path is repeated then reject
 	for _, cv := range config {
-		for _, p := range pathList {
-			if cv.Path == p {
-				return nil, errors.New("Error Path " + p + " is repeated in change")
-			}
-		}
+		//for _, p := range pathList {
```
Andrea-Campanella: why is this commented out ?

pkg/store/tree.go
```diff
@@ -160,3 +151,29 @@ func addPathToTree(path, value string, nodeif *interface{}) error {
...
+	case change.ValueTypeFLOAT:
+		(nodemap)[pathelems[1]] = (*change.TypedFloat)(value).Float32()
+	case change.ValueTypeBOOL:
+		(nodemap)[pathelems[1]] = (*change.TypedBool)(value).Bool()
+	case change.ValueTypeBYTES:
+		(nodemap)[pathelems[1]] = (*change.TypedBytes)(value).Bytes()
+	case change.ValueTypeLeafListSTRING:
+		(nodemap)[pathelems[1]] = (*change.TypedLeafListString)(value).List()
+	default:
+		(nodemap)[pathelems[1]] = fmt.Sprintf("unexpected %d", value.Type)
```
Andrea-Campanella: I don't see the check for this error the calling method, is it fine or are we ignoring this ?

modelplugin/yang/test1@2018-02-20.yang
```diff
@@ -44,6 +44,21 @@ module test1 {
         config false;
         description "Read only leaf inside Container 2a";
       }
+      leaf leaf2d {
```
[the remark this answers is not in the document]
SeanCondon: Done - bool now moved to 2g. 2d is now another decimal

pkg/cli/command/devicetree.go
```diff
@@ -125,7 +125,18 @@ func runDeviceTreeCommand(cmd *cobra.Command, args []string) {
 				Config:      make([]*change.Value, 0),
 			}
 			for _, cv := range in.Changevalues {
-				value, _ := change.CreateChangeValue(cv.Path, cv.Value, cv.Removed)
+				var tv *change.TypedValue
+				typeOptInt32 := make([]int, len(cv.Typeopts))
+				for i, v := range cv.Typeopts {
+					typeOptInt32[i] = int(v)
```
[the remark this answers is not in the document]
SeanCondon: Here we're reconstructing the TypedValue inside the **onos-cli** after retrieving it over gRPC from the server. The type could be decimal or LeafList of bytes, so if there are any TypeOpts, we would need them to fully recreate the type

pkg/manager/manager_test.go
```diff
@@ -380,8 +380,11 @@ func TestManager_ComputeRollbackDelete(t *testing.T) {
...
-	assert.Check(t, changes[Test1Cont1ACont2ALeaf2B] == ValueLeaf2B159, "Wrong value to set after rollback")
-	assert.Check(t, changes[Test1Cont1ACont2ALeaf2A] == ValueLeaf2A13, "Wrong value to set after rollback")
-	assert.Check(t, deletesRoll[0] == Test1Cont1ACont2ALeaf2D, "Path should be deleted")
+	assert.Equal(t, changes[Test1Cont1ACont2ALeaf2B].Type, change.ValueTypeFLOAT, "Wrong value to set after rollback")
+	assert.Equal(t, (*change.TypedFloat)(changes[Test1Cont1ACont2ALeaf2B]).Float32(), float32(ValueLeaf2B159), "Wrong value to set after rollback")
+
+	assert.Equal(t, changes[Test1Cont1ACont2ALeaf2A].Type, change.ValueTypeFLOAT, "Wrong value to set after rollback")
+	assert.Equal(t, (*change.TypedFloat)(changes[Test1Cont1ACont2ALeaf2A]).Float32(), float32(ValueLeaf2B159), "Wrong value to set after rollback")
 
+	assert.Check(t, deletesRoll[0] == Test1Cont1ACont2ALeaf2D, "Path should be deleted")
```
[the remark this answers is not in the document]
SeanCondon: done

pkg/northbound/gnmi/get.go
```diff
@@ -119,13 +120,15 @@ func getUpdate(prefix *gnmi.Path, path *gnmi.Path) (*gnmi.Update, error) {
...
+	var err error
 	if len(configValues) == 0 {
 		value = nil
 	} else if len(configValues) == 1 {
-		typedValue := &gnmi.TypedValue_StringVal{StringVal: configValues[0].Value}
-		value = &gnmi.TypedValue{
-			Value: typedValue,
+		value, err = values.NativeTypeToGnmiTypedValue(&configValues[0].TypedValue)
+		if err != nil {
+			log.Warning("Unable to convert native value to gnmi", err)
```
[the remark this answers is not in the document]
SeanCondon: done

pkg/northbound/gnmi/set.go
SeanCondon: done

pkg/northbound/gnmi/subscribe.go
SeanCondon: We do - see line 176 above

pkg/store/change/change.go
SeanCondon: good - catch - I had commented out when testing something and forgot to put it back in

pkg/store/tree.go
SeanCondon: the output will be visible in the JSON produced. It should never happen anyway all enum values are taken care of

pkg/cli/command/devicetree.go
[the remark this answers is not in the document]
Andrea-Campanella: +1

pkg/northbound/gnmi/set.go
```diff
@@ -132,15 +141,22 @@ func (s *Server) Set(ctx context.Context, req *gnmi.SetRequest) (*gnmi.SetRespon
...
 	updates, ok := targetUpdates[target]
 	if !ok {
-		updates = make(map[string]string)
+		updates = make(map[string]*change.TypedValue)
 	}
 	path := utils.StrPath(u.Path)
-	updates[path] = utils.StrVal(u.Val)
-	return updates
+
+	var err error
```
Andrea-Campanella: I do not believe you need a var here, just declare it inline

pkg/northbound/gnmi/set.go
```diff
@@ -46,7 +47,7 @@ func (s *Server) Set(ctx context.Context, req *gnmi.SetRequest) (*gnmi.SetRespon
 		version          string // May be specified as 101 in extension
 		deviceType       string // May be specified as 102 in extension
 	)
-
+	var err error
```
Andrea-Campanella: I do not believe you need a var here, just declare it inline underneath

pkg/northbound/gnmi/subscribe.go
Andrea-Campanella: my bad

pkg/store/change/typedvalue.go
```diff
@@ -86,7 +77,12 @@ func (tv *TypedValue) String() string {
 type TypedEmpty TypedValue
 
 // CreateTypedValueEmpty decodes an empty object
-func CreateTypedValueEmpty() *TypedEmpty {
+func CreateTypedValueEmpty() *TypedValue {
```
Andrea-Campanella: nice change to return interface TypedValue

pkg/store/tree.go
Andrea-Campanella: +1

pkg/northbound/gnmi/set.go
SeanCondon: Done

pkg/northbound/gnmi/set.go
SeanCondon: Cannot do inline because I'm doing **=** below not **:=**
I've moved it down to where it's needed to avoid any confusion with it

pkg/store/change/typedvalue.go
SeanCondon: works better

pkg/manager/setconfig.go
```diff
@@ -66,20 +66,20 @@ func (m *Manager) SetNetworkConfig(configName store.ConfigName, updates map[stri
 		}
 	}
 
-	var newChange = make([]*change.Value, 0)
+	var newChanges = make([]*change.Value, 0)
 	//updates
 	for path, value := range updates {
-		changeValue, _ := change.CreateChangeValue(path, value, false)
-		newChange = append(newChange, changeValue)
+		changeValue, _ := change.CreateChangeValue(path, (*change.TypedValue)(value), false)
```
Andrea-Campanella: why is the cast needed here ? the updates map already is of type TypedValue

pkg/store/tree.go
```diff
@@ -160,3 +151,45 @@ func addPathToTree(path, value string, nodeif *interface{}) error {
...
+	case change.ValueTypeSTRING:
+		(nodemap)[pathelems[1]] = (*change.TypedString)(value).String()
+	case change.ValueTypeINT:
+		(nodemap)[pathelems[1]] = (*change.TypedInt64)(value).Int()
+	case change.ValueTypeUINT:
+		(nodemap)[pathelems[1]] = (*change.TypedUint64)(value).Uint()
+	case change.ValueTypeDECIMAL:
+		(nodemap)[pathelems[1]] = (*change.TypedDecimal64)(value).Float()
+	case change.ValueTypeFLOAT:
+		if floatAsStr {
```
Andrea-Campanella: maybe I'm missing the point but why is this needed ? not supported in gnmi ?

SeanCondon: not supported in YGOT - doesn't seem to want float as raw (in JSON) - wants it as string - whatever

PR #492, opened 2019-07-09
Add an integration test for stratum switches (WIP)

adibrastegarnia:
@kuujo @Andrea-Campanella @ray-milkey 
This is a WIP PR. The main goal is to add an integration test for stratum switches as a base that we can use that later for new integration tests. Please add WIP and DO_NOT_MERGE labels until I update this comment.  #461 

1- List of devices stored in an environment variable called `TestDevicesEnv`. The first question that we should answer is: do we want to keep stratum devices separate than device simulators by defining a new env variable or do we want to append stratum switches into `TestDevicesEnv` as well. The advantage of the first one is that we don't need to worry about the indexing when we want to get a device simulator or stratum switch in an integration test.

### DIFF
--- a/test/configs/store/default.json
+++ b/test/configs/store/default.json
@@ -2,12 +2,92 @@
   "changeStore": {
     "Version": "1.0.0",
     "Storetype": "change",
-    "Store": {}
+    "Store": {
+      "J+CUGSfjr0npwzDN29ftbzXSG3o=": {
+        "ID": "J+CUGSfjr0npwzDN29ftbzXSG3o=",
+        "Description": "Initial configuration of Stratum Simulator",
+        "Created": "2019-06-05T12:03:17Z",
+        "Config": [
+          {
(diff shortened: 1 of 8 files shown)


Andrea-Campanella:
@adibrastegarnia I would keep one variable IMHO since it's cleaner and it's similar to what we will do during deployment. We dont' really want to distinguish a "device" too much.

test/configs/store/default.json
```diff
@@ -2,12 +2,92 @@
...
+        ]
+      }
+    }
   },
   "configStore": {
     "Version": "1.0.0",
     "Storetype": "config",
-    "Store": {}
+    "Store": {
+      "stratum-1-s0-1.0.0": {
```
SeanCondon: I think this is your problem - the key does not match the **name** value

adibrastegarnia:
I will close this and sometime later open a new one. We need to think about to avoid doing something now and then completely change it later.

PR #503, opened 2019-07-10
Nil pointer fix and flag for model plugins

Andrea-Campanella:
- Fix for Nil pointer during rollback 
- Adding the go modules flag for model plugin build instructions

### DIFF
--- a/pkg/northbound/gnmi/set.go
+++ b/pkg/northbound/gnmi/set.go
@@ -92,7 +92,6 @@ func (s *Server) Set(ctx context.Context, req *gnmi.SetRequest) (*gnmi.SetRespon
 				ext.GetRegisteredExt().GetId(), ext.GetRegisteredExt().GetMsg()).Error())
 		}
 	}
-
 	updateResults, networkChanges, err :=
 		s.buildUpdateResults(targetUpdates, targetRemoves, version, deviceType)
 	if err != nil {
@@ -184,6 +183,7 @@ func (s *Server) buildUpdateResults(targetUpdates mapTargetUpdates,
 		changeID, configName, cont, err := setChange(target, version, updates, targetRemoves[target])
 		//if the error is not nil and we need to continue do so
 		if err != nil && !cont {
+			//TODO Rollback needs also to happen here.
 			return nil, nil, err
 		} else if err == nil && cont {
 			continue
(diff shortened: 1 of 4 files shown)


pkg/northbound/gnmi/set.go
```diff
@@ -184,6 +183,7 @@ func (s *Server) buildUpdateResults(targetUpdates mapTargetUpdates,
 		changeID, configName, cont, err := setChange(target, version, updates, targetRemoves[target])
 		//if the error is not nil and we need to continue do so
 		if err != nil && !cont {
+			//TODO Rollback needs also to happen here.
```
SeanCondon: you took this back out?

pkg/northbound/gnmi/set.go
```diff
@@ -280,6 +267,27 @@ func listenForDeviceResponse(changes mapNetworkChanges, target string, name stor
 	}
 }
 
+func doRollback(changes mapNetworkChanges, mgr *manager.Manager, target string,
+	name store.ConfigName, errResp error) error {
+	rolledbackIDs := make([]string, 0)
+	for configName := range changes {
+		changeID, err := mgr.RollbackTargetConfig(string(configName))
```
SeanCondon: maybe RollbackTargetConfig() should take a **ConfigName** rather than string

pkg/northbound/gnmi/set.go
```diff
@@ -280,6 +267,27 @@ func listenForDeviceResponse(changes mapNetworkChanges, target string, name stor
...
+		log.Infof("Rolling back %s on %s", store.B64(changeID), configName)
+		if err != nil {
+			log.Error("Can'configName remove last entry for ", target, err)
+		}
+		rolledbackIDs = append(rolledbackIDs, store.B64(changeID))
+	}
+	//Removing the failed target
+	changeID, errRoll := mgr.RollbackTargetConfig(string(name))
+	if errRoll != nil {
+		log.Error("Can'configName remove last entry for ", target, errRoll)
```
SeanCondon: typo in **can't**

pkg/northbound/gnmi/set.go
```diff
@@ -280,6 +267,27 @@ func listenForDeviceResponse(changes mapNetworkChanges, target string, name stor
 	}
 }
 
+func doRollback(changes mapNetworkChanges, mgr *manager.Manager, target string,
+	name store.ConfigName, errResp error) error {
+	rolledbackIDs := make([]string, 0)
+	for configName := range changes {
+		changeID, err := mgr.RollbackTargetConfig(string(configName))
+		log.Infof("Rolling back %s on %s", store.B64(changeID), configName)
```
SeanCondon: log should happen after error check or before calling RollbackTargetConfig()

pkg/northbound/gnmi/set.go
Andrea-Campanella: yes, needs further investigation, want to get these fixes in ASAP

pkg/northbound/gnmi/set.go
Andrea-Campanella: yeah, that's part fo the configName/string refactor issue I opened. don't want to do it here

pkg/northbound/gnmi/set.go

PR #508, opened 2019-07-11
No Duplicate changes in device config

Andrea-Campanella:
When doing rollback we do not set a copy of the previous change but instead leave the set of changes as it was prior to the last set and rollback operation. 
Refactored also a common method.
We also listen for errors during the set request for rollback


pkg/manager/manager.go
```diff
@@ -211,3 +213,30 @@ func listenOnResponseChannel(respChan chan events.DeviceResponse) {
 		//TODO handle device connection errors accordingly
 	}
 }
+
+func (m *Manager) computeAndStoreChange(updates map[string]*change.TypedValue,
```
SeanCondon: It would be enough to just return the configChange.ID - that's the only part of it that's used by the caller anyway

PR #509, opened 2019-07-11
Fixed hardcoding of devicetype in 'onos devices add ...'

SeanCondon:


adibrastegarnia:
@SeanCondon 
We also need to make changes in onit. I will take care of it when this patch merged

SeanCondon:
@adibrastegarnia I forgot to leave the default value for DeviceType as "DeviceSim", so I'll push again
Where is it called in **onit** anyway?

adibrastegarnia:
> @adibrastegarnia I forgot to leave the default value for DeviceType as "DeviceSim", so I'll push again
> Where is it called in **onit** anyway?

in onos-config.go there are two functions one used for simulator and the same pattern for stratum.

PR #514, opened 2019-07-12
Check for duplicate against just the last change ID. Fixes: #511

### LINKED ISSUE #511: transaction integration test fails when run after other tests
**Describe the bug**
The transaction test will fail if it is run after the single-path and subscribe tests.

**To Reproduce**
1. make kind
2. onit run test single-path
3. onit run test subscribe
4. onit run test transaction

**Expected behavior**
Tests should all report PASS:

**Logs, text or screenshots**
Rays-MacBook:onos-config ray$ onit run test transaction
 ✓ Starting test job: test-2226283700 
=== RUN   transaction
--- FAIL: transaction (0.15s)
    transactiontest.go:91: 
        	Error Trace:	transactiontest.go:91
        	Error:      	Received unexpected error:
        	            	rpc error: code = Unknown desc = Can't remove last entry for device-1036654090-1.0.0%!(EXTRA *errors.errorString=Duplicate change ID 'l^M�����R��[�k�4\' in config)
        	Test:       	transaction
        	Messages:   	Rollback returned an error
    transactiontest.go:92: 
        	Error Trace:	transactiontest.go:92
        	Error:      	Expected value not to be nil.
        	Test:       	transaction
        	Messages:   	Response for rollback is nil
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x11cdbbd]

goroutine 33 [running]:
testing.tRunner.func1(0xc0002be900)
	/usr/local/go/src/testing/testing.go:830 +0x392
panic(0x1316260, 0x21b6660)
	/usr/local/go/src/runtime/panic.go:522 +0x1b5
github.com/onosproject/onos-config/test/integration.TestTransaction(0xc0002be900)
	/go/src/github.com/onosproject/onos-config/test/integration/transactiontest.go:93 +0x8ad
testing.tRunner(0xc0002be900, 0x1523390)
	/usr/local/go/src/testing/testing.go:865 +0xc0
created by testing.(*T).Run
	/usr/local/go/src/testing/testing.go:916 +0x35a

Andrea-Campanella:
while Create a new configuration we were checking that there was no duplication of changes in the whole array, this is not a valid case as for example a set(v), delete(v), set(v), delete(v) is a valid operation.
Now checking that we are not setting the same value twice in a row. 

Fixes #511

### DIFF
--- a/pkg/manager/rollbackconfig.go
+++ b/pkg/manager/rollbackconfig.go
@@ -46,7 +46,8 @@ func (m *Manager) RollbackTargetConfig(configname string) (change.ID, error) {
 func computeRollback(m *Manager, target string, configname string) (change.ID, map[string]*change.TypedValue, []string, error) {
 	id, err := m.ConfigStore.RemoveLastChangeEntry(store.ConfigName(configname))
 	if err != nil {
-		return nil, nil, nil, fmt.Errorf("Can't remove last entry for %s, %s", configname, err.Error())
+		return nil, nil, nil, fmt.Errorf("Can't remove last entry on target %s in config %s, %s",
+			target, configname, err.Error())
 	}
 	previousValues := make([]*change.ConfigValue, 0)
 	deletes := make([]string, 0)
--- a/pkg/manager/rollbackconfig.go
+++ b/pkg/manager/rollbackconfig.go
@@ -31,6 +31,7 @@ func (m *Manager) RollbackTargetConfig(configname string) (change.ID, error) {
 	log.Infof("Rolling back last change on config %s for target %s", configname, targetID)
 	id, updates, deletes, err := computeRollback(m, targetID, configname)
 	if err != nil {
+		log.Errorf("Error on rollback: %s", err.Error())
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


pkg/manager/rollbackconfig.go
```diff
@@ -45,7 +46,7 @@ func (m *Manager) RollbackTargetConfig(configname string) (change.ID, error) {
 func computeRollback(m *Manager, target string, configname string) (change.ID, map[string]*change.TypedValue, []string, error) {
 	id, err := m.ConfigStore.RemoveLastChangeEntry(store.ConfigName(configname))
 	if err != nil {
-		return nil, nil, nil, fmt.Errorf(fmt.Sprintf("Can't remove last entry for %s", configname), err)
+		return nil, nil, nil, fmt.Errorf("Can't remove last entry for %s, %s", configname, err.Error())
```
SeanCondon: should put in the targetID too

PR #516, opened 2019-07-12
Added checking of readonly types during Set

SeanCondon:
Should cause a set of a readonly value (e.g. leaf2c) to fail:
```
gnmi_cli -address localhost:5150 -set \
-proto "update: <path: <target: 'Device1', elem: <name: 'test1:cont1a'> elem: <name: 'cont2a'> elem: <name: 'leaf2c'>> val: <string_val: 'not allowed'>>" \
-timeout 5s -alsologtostderr \
-client_crt pkg/southbound/testdata/client1.crt \
-client_key pkg/southbound/testdata/client1.key \
-ca_crt pkg/southbound/testdata/onfca.crt
```

### DIFF
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -42,6 +42,7 @@ type Manager struct {
 	SouthboundErrorChan     chan events.DeviceResponse
 	Dispatcher              dispatcher.Dispatcher
 	ModelRegistry           map[string]ModelPlugin
+	ModelReadOnlyPaths      map[string][]string
 }
 
 // NewManager initializes the network config manager subsystem.
@@ -59,6 +60,7 @@ func NewManager(configs *store.ConfigurationStore, changes *store.ChangeStore, d
 		SouthboundErrorChan:     make(chan events.DeviceResponse, 10),
 		Dispatcher:              dispatcher.NewDispatcher(),
 		ModelRegistry:           make(map[string]ModelPlugin),
+		ModelReadOnlyPaths:      make(map[string][]string),
 	}
 
 	changeIds := make([]string, 0)
--- a/pkg/manager/modelregistry.go
+++ b/pkg/manager/modelregistry.go
@@ -21,6 +21,8 @@ import (
 	"github.com/openconfig/ygot/ygot"
 	log "k8s.io/klog"
 	"plugin"
+	"regexp"
+	"strings"
 )
 
 // ModelPlugin is a set of methods that each model plugin should implement
@@ -48,12 +50,22 @@ func (m *Manager) RegisterModelPlugin(moduleName string) (string, string, error)
 	modelPlugin, ok := symbolMP.(ModelPlugin)
 	if !ok {
 		log.Warning("Unable to use ModelPlugin in ", moduleName)
-		return "", "", fmt.Errorf("Symbol loaded from module %s is not a ModelPlugin",
+		return "", "", fmt.Errorf("symbol loaded from module %s is not a ModelPlugin",
 			moduleName)
 	}
 	name, version, _, _ := modelPlugin.ModelData()
-	modelName := toModelName(name, version)
+	modelName := ToModelName(name, version)
 	m.ModelRegistry[modelName] = modelPlugin
+	modelschema, err := modelPlugin.Schema()
+	if err != nil {
+		log.Warning("Error loading schema from model plugin", modelName, err)
+		return "", "", err
+	}
+
+	m.ModelReadOnlyPaths[modelName] = extractReadOnlyPaths(modelschema["Device"],
+		yang.TSUnset, "", "")
+	log.Infof("Model %s %s loaded. %d read only paths", name, version,
+		len(m.ModelReadOnlyPaths[modelName]))
 	return name, version, nil
 }
 
@@ -65,7 +77,7 @@ func (m *Manager) Capabilities() []*gnmi.ModelData {
 	for _, model := range m.ModelRegistry {
 		_, _, modelItem, _ := model.ModelData()
 		for _, mi := range modelItem {
-			modelName := toModelName(mi.Name, mi.Version)
+			modelName := ToModelName(mi.Name, mi.Version)
 			modelMap[modelName] = mi
 		}
 	}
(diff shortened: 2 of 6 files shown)


pkg/manager/manager.go
```diff
@@ -42,6 +42,7 @@ type Manager struct {
 	SouthboundErrorChan     chan events.DeviceResponse
 	Dispatcher              dispatcher.Dispatcher
 	ModelRegistry           map[string]ModelPlugin
+	ModelReadOnlyPaths      map[string][]string
```
Andrea-Campanella: I'm in doubt this information is in the right place here, is it worth creating a model registry store that encompasses both the elements at line 44 and at line 45 ?

pkg/manager/modelregistry.go
```diff
@@ -48,12 +50,22 @@ func (m *Manager) RegisterModelPlugin(moduleName string) (string, string, error)
...
+		return "", "", fmt.Errorf("symbol loaded from module %s is not a ModelPlugin",
 			moduleName)
 	}
 	name, version, _, _ := modelPlugin.ModelData()
-	modelName := toModelName(name, version)
+	modelName := ToModelName(name, version)
 	m.ModelRegistry[modelName] = modelPlugin
+	modelschema, err := modelPlugin.Schema()
+	if err != nil {
+		log.Warning("Error loading model plugin", err)
```
Andrea-Campanella: this is actually more reading model plugin schema. the model plugin has already being loaded.

pkg/manager/modelregistry.go
```diff
@@ -79,6 +91,69 @@ func (m *Manager) Capabilities() []*gnmi.ModelData {
...
+	readOnlyPaths := make([]string, 0)
+
+	for _, dirEntry := range deviceEntry.Dir {
+		var itemPath string
+		if parentPath == "" {
+			itemPath = fmt.Sprintf("%s/%s:%s", parentPath, dirEntry.Prefix.Name, dirEntry.Name)
+		} else {
+			itemPath = fmt.Sprintf("%s/%s", parentPath, dirEntry.Name)
+		}
+		//fmt.Println("Processing", itemPath)
```
Andrea-Campanella: might want to remove this

pkg/manager/modelregistry.go
```diff
@@ -79,6 +91,69 @@ func (m *Manager) Capabilities() []*gnmi.ModelData {
...
 
-func toModelName(name string, version string) string {
+// extractReadOnlyPaths is a recursive function to extract a list of read only paths from a YGOT schema
+func extractReadOnlyPaths(deviceEntry *yang.Entry, parentState yang.TriState, parentPrefix string, parentPath string) []string {
+	readOnlyPaths := make([]string, 0)
+
+	for _, dirEntry := range deviceEntry.Dir {
+		var itemPath string
+		if parentPath == "" {
+			itemPath = fmt.Sprintf("%s/%s:%s", parentPath, dirEntry.Prefix.Name, dirEntry.Name)
```
Andrea-Campanella: what if the prefix is empty ?

pkg/manager/modelregistry.go
```diff
@@ -79,6 +91,69 @@ func (m *Manager) Capabilities() []*gnmi.ModelData {
...
+			itemPath = fmt.Sprintf("%s/%s", parentPath, dirEntry.Name)
+		}
+		//fmt.Println("Processing", itemPath)
+		if dirEntry.IsLeaf() {
+			// No need to recurse
+			if dirEntry.Config == yang.TSFalse || parentState == yang.TSFalse {
+				readOnlyPaths = append(readOnlyPaths, itemPath)
+			}
+		} else if dirEntry.IsContainer() {
+			var name string
```
Andrea-Campanella: I don't think this is needed, you can use := in the if else clause

pkg/manager/modelregistry.go
```diff
@@ -79,6 +91,69 @@ func (m *Manager) Capabilities() []*gnmi.ModelData {
...
+			}
+			if dirEntry.Config == yang.TSFalse || parentState == yang.TSFalse {
+				readOnlyPaths = append(readOnlyPaths, itemPath)
+			}
+			newPaths := extractReadOnlyPaths(dirEntry, dirEntry.Config, dirEntry.Prefix.Name, name)
+			for _, newPath := range newPaths {
+				readOnlyPaths = append(readOnlyPaths, newPath)
+			}
+		} else if dirEntry.IsList() {
+			var name string
```
Andrea-Campanella: same

pkg/manager/modelregistry.go
```diff
@@ -79,6 +91,69 @@ func (m *Manager) Capabilities() []*gnmi.ModelData {
...
+			if dirEntry.Prefix.Name == parentPrefix {
+				name = parentPath + "/" + dirEntry.Name + "[" + dirEntry.Key + "='*']"
```
Andrea-Campanella: can we use fmt.Sprintf as before. I'd actually make a function out of it for simplicity reason and also for avoiding to forget formatting changes here and there.

pkg/manager/setconfig.go
```diff
@@ -60,8 +60,7 @@ func (m *Manager) SetNetworkConfig(configName store.ConfigName, updates map[stri
 				configName, conv(similarDevices))
 		} else {
 			return nil, configName, fmt.Errorf("no configuration found matching '%s'."+
-				"Please specify version and device type in extensions 101 and 102 "+
-				"and model data in extensions 150-200 in name@ver@org format", configName)
```
Andrea-Campanella: why are you removing this line ?

pkg/northbound/gnmi/set.go
```diff
@@ -171,7 +176,62 @@ func (s *Server) doDelete(u *gnmi.Path, targetRemoves mapTargetRemoves) []string
...
 }
 
-func (s *Server) buildUpdateResults(targetUpdates mapTargetUpdates,
+func (s *Server) checkForReadOnly(deviceType string, version string, targetUpdates mapTargetUpdates,
+	targetRemoves mapTargetRemoves) error {
+	configs := manager.GetManager().ConfigStore.Store
+
+	targetModelTypes := make(map[string][]string)
+	for t := range targetUpdates { // map - just need the key
+		if _, ok := targetModelTypes[t]; ok {
```
Andrea-Campanella: I don't get the meaning of this for loop. if it's there we continue ?

pkg/northbound/gnmi/set.go
```diff
@@ -171,7 +176,62 @@ func (s *Server) doDelete(u *gnmi.Path, targetRemoves mapTargetRemoves) []string
...
+			continue
+		}
+		for _, config := range configs {
+			if config.Device == t {
+				m, ok := manager.GetManager().ModelReadOnlyPaths[manager.ToModelName(config.Type, config.Version)]
+				if !ok {
+					log.Warningf("Cannot check for Read Only paths for %s %s because "+
```
Andrea-Campanella: IMHO we shoudl check the existence of the model plugin before hand, otherwise for each path we will see this message. Or if this is intended to be a per path error message maybe specify it differently

pkg/manager/manager.go
SeanCondon: Good question: **manager** is the one and only global object - I wasn't inclined to create another - didn't see the point.
It doesn't require a store - it's just a cache.
We could do one object (struct) that encompasses both though if you think that would do.

pkg/manager/modelregistry.go
SeanCondon: done

pkg/manager/modelregistry.go
SeanCondon: done

pkg/manager/modelregistry.go
SeanCondon: Does not ever seem to be

pkg/manager/modelregistry.go
SeanCondon: but then it woun't be available outside the **if** statement

pkg/manager/modelregistry.go
SeanCondon: same

pkg/manager/modelregistry.go
SeanCondon: all functioned up now ;-)

pkg/manager/manager.go
Andrea-Campanella: I'm fine leaving as is for now.

pkg/manager/setconfig.go
SeanCondon: this is not showing the change properly - only the 
```
and model data in extensions 150-200 in name@ver@org format
```
part has been removed

pkg/northbound/gnmi/set.go
SeanCondon: It Iterates through all the updates - many may use the same target - here we create a map of the models for all of the targets. Same for the updates. It's just an optimization on the looping

Comment added to code

pkg/northbound/gnmi/set.go
SeanCondon: There's probably a balance in there - if we had only 30 targets on the system it might be worth while keeping a cache per target. On the other hand most requests will be only for a few targets, and it's only on the Set

Andrea-Campanella: ok

pkg/manager/modelregistry.go
```diff
@@ -79,6 +91,82 @@ func (m *Manager) Capabilities() []*gnmi.ModelData {
...
+
+	for _, dirEntry := range deviceEntry.Dir {
+		prefix := ""
+		if dirEntry.Prefix != nil {
+			prefix = dirEntry.Prefix.Name
+		}
+		var itemPath string
+		if parentPath == "" {
+
+			itemPath = fmt.Sprintf("%s/%s:%s", parentPath, prefix, dirEntry.Name)
```
Andrea-Campanella: should format name be use also at 106 and 108 ? also 105 is an empty useless line

SeanCondon: don't know what I was thinking of - all consolidated now

PR #526, opened 2019-07-15
Query devices directly to make sure that changes were applied

ray-milkey:
After setting paths through the GNMI onos-config interface, check that the devices were updated.

### DIFF
--- a/test/integration/transactiontest.go
+++ b/test/integration/transactiontest.go
@@ -19,6 +19,7 @@ import (
 	admin "github.com/onosproject/onos-config/pkg/northbound/proto"
 	"github.com/onosproject/onos-config/test/env"
 	"github.com/onosproject/onos-config/test/runner"
+	"github.com/openconfig/gnmi/client"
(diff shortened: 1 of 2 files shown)


test/integration/transactiontest.go
```diff
@@ -58,30 +60,46 @@ func getDevicePathsWithValues(device string, paths []string, values[]string) []D
...
+	// Get the configured devices from the environment.
+	device1 := env.GetDevices()[0]
+	device2 := env.GetDevices()[1]
+	devices := make([]string, 2)
+	devices[0] = device1
+	devices[1] = device2
 
 	// Make a GNMI client to use for requests
-	gnmiClient, gnmiClientError := env.NewGnmiClient(MakeContext(), "")
+	gnmiClient, gnmiClientError := env.NewGnmiClientForDevice(MakeContext(), "onos-config:5150", device1)
```
Andrea-Campanella: imho `device1` here needs to be `gnmi`

PR #527, opened 2019-07-16
State and Operational paths from model

Andrea-Campanella:
Leveraging models to obtain state and operational paths to subscribe to in case the device does not give a reply to STATE or OPERATIONAL get requests.

### DIFF
--- a/pkg/southbound/synchronizer/factory.go
+++ b/pkg/southbound/synchronizer/factory.go
@@ -20,6 +20,7 @@ import (
 	"github.com/onosproject/onos-config/pkg/events"
 	"github.com/onosproject/onos-config/pkg/southbound/topocache"
 	"github.com/onosproject/onos-config/pkg/store"
+	"github.com/onosproject/onos-config/pkg/utils"
 	log "k8s.io/klog"
 )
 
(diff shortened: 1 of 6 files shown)


pkg/southbound/synchronizer/synchronizer.go
```diff
@@ -44,12 +44,13 @@ type Synchronizer struct {
...
 	operationalCache     map[string]string
+	modelReadOnlyPaths   map[string][]string
 }
 
 // New Build a new Synchronizer given the parameters, starts the connection with the device and polls the capabilities
 func New(context context.Context, changeStore *store.ChangeStore, configStore *store.ConfigurationStore,
-	device *topocache.Device, deviceCfgChan <-chan events.ConfigEvent,
-	opStateChan chan<- events.OperationalStateEvent, errChan chan<- events.DeviceResponse) (*Synchronizer, error) {
+	device *topocache.Device, deviceCfgChan <-chan events.ConfigEvent, opStateChan chan<- events.OperationalStateEvent,
+	errChan chan<- events.DeviceResponse, mReadOnlyPaths map[string][]string) (*Synchronizer, error) {
```
SeanCondon: I think you only need to pass in an **[]string** - i.e. the one for the type of device you are creating

pkg/southbound/synchronizer/factory.go
```diff
@@ -38,8 +39,11 @@ func Factory(changeStore *store.ChangeStore, configStore *store.ConfigurationSto
 			}
 			device := deviceStore.Store[topocache.ID(deviceName)]
 			ctx := context.Background()
+			completeID := utils.ToConfigName(deviceName, device.SoftwareVersion)
+			cfg := configStore.Store[store.ConfigName(completeID)]
+			mReadOnlyPaths, _ := modelReadOnlyPaths[utils.ToModelName(cfg.Type, device.SoftwareVersion)]
```
SeanCondon: maybe should use **mReadOnlyPaths, ok :=** in the result

pkg/southbound/synchronizer/factory.go
```diff
@@ -38,8 +39,11 @@ func Factory(changeStore *store.ChangeStore, configStore *store.ConfigurationSto
 			}
 			device := deviceStore.Store[topocache.ID(deviceName)]
 			ctx := context.Background()
+			completeID := utils.ToConfigName(deviceName, device.SoftwareVersion)
+			cfg := configStore.Store[store.ConfigName(completeID)]
```
SeanCondon: Use cfg, ok to check there was actually a value

PR #530, opened 2019-07-16
Added in certs for onos-config - moved tls.crt to client1.crt, reflecting it's subject

SeanCondon:


kuujo:
@SeanCondon also, the Helm chart certs _should_ be generated with `onos-config.opennetworking.org`

Issue #535, opened 2019-07-16
More thorough onit command documentation


kuujo:
**Is your feature request related to a problem? Please describe.**
`onit` command help does not currently describe sufficiently the features of onit and their uses. Users should be able to learn the commands through the command help.

**Describe the solution you'd like**
Add thorough documentation with examples to all onit commands.

kuujo:
This still needs a second PR for the cobra command help.

Issue #531, opened 2019-07-16
Missing completion for onit networks


kuujo:
**Describe the bug**
onit does not currently support completion for networks

**To Reproduce**
1. Add a network
2. Run `onit remove network [tab]`
3. See no completions

**Expected behavior**
onit should complete network names

PR #536, opened 2019-07-16
Add Missing completion for onit networks- Update documents to include examples to all onit commands. (#531, #535) 

adibrastegarnia:
This PR addresses issues #531 #535

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -47,6 +47,25 @@ configuration is primarily managed through various onit commands like `onit set`
 file, but it's nevertheless important to note that the application must have write access to one
 of the above paths.
 
+
+### Onit Auto-Completion
+*Onit* supports shell auto-completion for its various commands, sub-commands and flags.
+You can enable this feature for *bash* or *zsh* as follows:
+#### Bash Auto-Completion
+To enable this for **bash**, run the following from the shell:
+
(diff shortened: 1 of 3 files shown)


docs/testing.md
```diff
@@ -47,6 +47,13 @@ configuration is primarily managed through various onit commands like `onit set`
 file, but it's nevertheless important to note that the application must have write access to one
 of the above paths.
 
+
+### Bash or Zsh Auto-Completion
+*Onit* supports shell auto-completion for its various commands, sub-commands and flags. To enable this, run the following from the shell:
+
+```bash
+> eval "$(onit completion bash)"
```
kuujo: This needs to be `eval $(onit completion bash)` for bash and `source <(onit completion zsh)` for zsh. This also needs to be added to the `onit completion x` documentation as examples. And we can probably separate the bash and zsh sections here.

We should also recommend adding the output to `.bashrc` and `.zshrc`

adibrastegarnia: Sure. I will take care of it.

adibrastegarnia: This comment is addressed but there are other things that I should add or update

tomikazi:
I'm leaving it to Jordan to complete review and merge.

PR #538, opened 2019-07-17
Extracting model registry as first entity object

Andrea-Campanella:
To avoid circular dependencies we are extracting the model registry as a first entity object.
The patch also adds a location list with the names of the loaded .so files for the model plugins.
This map will be replicated in `atomix` for other instances of `onos-config` to access it and keep state in check.

### DIFF
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"github.com/onosproject/onos-config/pkg/dispatcher"
 	"github.com/onosproject/onos-config/pkg/events"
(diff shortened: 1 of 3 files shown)


pkg/manager/manager.go
```diff
@@ -36,31 +37,41 @@ type Manager struct {
 	ChangeStore             *store.ChangeStore
 	DeviceStore             *topocache.DeviceStore
 	NetworkStore            *store.NetworkStore
+	ModelRegistryObj        *modelregistry.ModelRegistry
```
SeanCondon: I think you can't make it a child of **manager** - it has to be a peer of manager, started from Main, like NewModelRegistry or something and accessible as a global variable

PR #539, opened 2019-07-17
Optimized the calculation of ReadOnly paths from the model Plugins Fixes #525

### LINKED ISSUE #525: Improve performance of gathering readonly paths from Model Plugins
When extracting "readonly paths" from Model Plugins, the system analyses every leaf in the system. Instead it should stop when it finds a container or list marked "config false" because there can only be "config false" items below that. Currently for the devicesim model there are **175** paths - this should be something like 50

Do the task more efficiently.

Not doing it - leaving it as it is will an impact on future features that use this list

**Additional context**
The list of read only paths will be used to set up a subscribe to devices - for this reason we would like the list of paths to be as concise as possible

SeanCondon:


docs/modelplugin.md
```diff
@@ -215,3 +209,35 @@ was needed.
 > If the generator program reports that a dependency was required e.g.
 **openconfig-inet-types.yang** then the version of this file with a date equal
 to or before 2017-07-14 should be downloaded - it is **openconfig-inet-types@2017-07-14.yang**
+
+## Readonly paths in OpenConfig models
```
Andrea-Campanella: I think this is valid for all models not just openconfig.

pkg/northbound/admin/admin.go
```diff
@@ -76,13 +78,23 @@ func (s Server) ListRegisteredModels(req *proto.ListModelsRequest, stream proto.
 			schemaEntries = append(schemaEntries, &schemaEntry)
 		}
 
+		roPaths := make([]string, 0)
+		if req.Verbose {
+			var ok bool
```
Andrea-Campanella: why do you need to define the ok variable ?

docs/modelplugin.md
SeanCondon: This is specifically an OpenConfig section of the document, so I thought it was appropriate

pkg/northbound/admin/admin.go
SeanCondon: because I can't use ":=" in the next line because roPaths exists already

SeanCondon: because I can't use ":=" in the next line

Andrea-Campanella: you can still do := if there is at least one new variable to the left of the assignment.

SeanCondon: there are bigger issues than this

SeanCondon: doesn't work in this case

PR #544, opened 2019-07-18
Added in ModelPlugins to the onit tool Fixes #537

### LINKED ISSUE #537: Add the use of Model Plugins to the 'onit' tool
Currently the **onit** tool runs without loading the Model Plugins

The Model Plugins should be loaded by default to ensure better integration testing

**Additional context**
The Model Plugins extend onos-config with YANG models and have a benefit when attributes are being Set

SeanCondon:


build/onos-config-debug/Dockerfile
```diff
@@ -17,6 +17,9 @@ COPY --from=base /go/src/github.com/onosproject/onos-config/build/_output/onos-c
 COPY --from=base /go/src/github.com/onosproject/onos-config/build/_output/testdevice.so.1.0.0 /usr/local/lib/testdevice.so.1.0.0
```
Andrea-Campanella: do you need the base ones here since this is the debug image?

SeanCondon: I thought it would be safer to copy them over in case there was an option of running **onit** in non-debug mode in future

Andrea-Campanella: let's remove them for now. Also I think running `onit` in non debug mode would make use of `latest` for docker and not `debug`

Issue #543, opened 2019-07-18
Pre-check against model for all Devices


Andrea-Campanella:
**Describe the bug**
Currently ones-config does a sequential apply of the configuration, meaning that if we have a configuration for device 1 and 2 the elements for device 1 will be checked against the model and immediately applied. If a failure on the model check happens on device 2 we need to actively rollback on Device 1. 

**To Reproduce**
1. gnmi_cli -set -address localhost:5150 \
	-proto "update: <path: <target: 'localhost-1', elem: <name: 'openconfig-system:system'> elem: <name: 'clock' > elem: <name: 'config'> elem: <name: 'timezone-name'>> val: <string_val: 'Europe/Berlin'>>, update: <path: <target: 'localhost-1', elem: <name: 'openconfig-system:system'> elem:<name:'config'> elem: <name: 'motd-banner'>> val: <string_val: 'New this is a demo device 1'>>, update: <path: <target: 'localhost-2', elem: <name: 'openconfig-system:system'> elem: <name: 'clock' > elem: <name: 'config'> elem: <name: 'timezone-name'>> val: <string_val: 'Europe/Paris'>>, update: <path:<target: 'localhost-2', elem:<name:'openconfig-system:system'>elem:<name:'aaa'>elem:<name:'foobar' > elem:<name:'config'>elem:<name:'authentication-method'>>val:<string_val:'openconfig-aaa-types:TEST' >>" \
	-timeout 5s -alsologtostderr \
	-client_crt pkg/southbound/testdata/client1.crt \
	-client_key pkg/southbound/testdata/client1.key \
	-ca_crt pkg/southbound/testdata/onfca.crt
3. observe physical rollback

**Expected behavior**
There should be a pre-check of the validity of the configuration before any set requests get issued to *any* device.

PR #545, opened 2019-07-18
Config check before any set requests issued

Andrea-Campanella:
Checking configuration validity against models before applying any to any device. Fixes:#543

### DIFF
--- a/pkg/manager/setconfig.go
+++ b/pkg/manager/setconfig.go
@@ -27,10 +27,91 @@ import (
 // SetConfigAlreadyApplied is a string constant for "Already applied:"
 const SetConfigAlreadyApplied = "Already applied:"
 
-// SetNetworkConfig sets the given value, according to the path on the configuration for the specified target
+// ValidateNetworkConfig validates the given updates and deletes, according to the path on the configuration
+// for the specified target
+func (m *Manager) ValidateNetworkConfig(configName store.ConfigName, updates map[string]*change.TypedValue,
+	deletes []string) (change.ID, error) {
+
+	deviceConfig, _, err := m.getStoredConfig(configName)
+	if err != nil {
+		return nil, err
+	}
+	deviceConfigTemporary, _ := store.CreateConfiguration(deviceConfig.Device, deviceConfig.Version,
+		deviceConfig.Type, deviceConfig.Changes)
+	chg, err := m.computeChange(updates, deletes)
+	if err != nil {
+		return nil, err
+	}
+	changeID, err := m.storeChange(chg)
+	if err != nil {
+		return changeID, err
+	}
+
+	deviceConfigTemporary.Changes = append(deviceConfig.Changes, changeID)
+
+	modelName := fmt.Sprintf("%s-%s", deviceConfig.Type, deviceConfig.Version)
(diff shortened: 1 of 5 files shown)


pkg/northbound/gnmi/set.go
```diff
@@ -98,6 +98,23 @@ func (s *Server) Set(ctx context.Context, req *gnmi.SetRequest) (*gnmi.SetRespon
 		return nil, status.Error(codes.InvalidArgument, errRo.Error())
 	}
 
+	//Checking for wrong configuration against the device models for updates
+	for target, updates := range targetUpdates {
+		_, err := validateChange(target, version, updates, targetRemoves[target])
```
SeanCondon: you could parallelize this like I do in manager.go ValidateStores()

pkg/northbound/gnmi/set.go
```diff
@@ -98,6 +98,23 @@ func (s *Server) Set(ctx context.Context, req *gnmi.SetRequest) (*gnmi.SetRespon
...
+		if err != nil {
+			return nil, err
+		}
+		//TODO copy targetRemoves into a temporary set and delete from that and use it at line 112
+		//delete(targetRemoves, target)
+	}
+	//Checking for wrong configuration against the device models for deletes
+	for target, removes := range targetRemoves {
+		_, err := validateChange(target, version, make(map[string]*change.TypedValue), removes)
```
SeanCondon: same as above

pkg/manager/setconfig.go
```diff
@@ -115,3 +110,40 @@ func (m *Manager) SetNetworkConfig(configName store.ConfigName, updates map[stri
 
 	return changeID, configName, nil
 }
+
+func (m *Manager) getStoredConfig(configName store.ConfigName) (store.Configuration, store.ConfigName, error) {
+	// Look for an exact match or then a partial match
```
SeanCondon: maybe put the comment before the func

pkg/manager/setconfig.go
```diff
@@ -27,85 +27,80 @@ import (
...
-		for key := range m.ConfigStore.Store {
-			deviceNameFromKey := string(key)[:strings.LastIndex(string(key), "-")]
-			if deviceNameFromKey == string(configName) {
-				similarDevices = append(similarDevices, key)
-			}
-		}
+// ValidateNetworkConfig validates the given updates and deletes, according to the path on the configuration
+// for the specified target
+func (m *Manager) ValidateNetworkConfig(configName store.ConfigName, updates map[string]*change.TypedValue,
+	deletes []string) (change.ID, error) {
```
SeanCondon: now that you don't store it I think there's no point at all in returning this change.ID

pkg/northbound/gnmi/set.go
```diff
@@ -386,3 +403,18 @@ func setChange(target string, version string, targetUpdates map[string]*change.T
 	}
 	return changeID, configName, false, nil
 }
+
+func validateChange(target string, version string, targetUpdates map[string]*change.TypedValue, targetRemoves []string) (change.ID, error) {
```
SeanCondon: again, because you don't store the change I think there's no point in returning it

pkg/manager/setconfig.go
Andrea-Campanella: done

PR #546, opened 2019-07-18
Make debugging in onit easier

adibrastegarnia:
This PR is to address issue #547 . The goal is to make debugging of subsystems (i.e. onos-config, onos-topo, onos-control, etc) easier in onit. In addition, we would like to document what  steps will be needed for debugging of a subsystem (e.g. onos-config) and add some working examples. 

1- Enable dlv server logging 
2- Add an example to show the usage of dlv client  for debugging such as setting breakpoints, printing local variables, check goroutines, print stack trace, debugging the code line by line, and etc in one of the onos-config source code files.


tomikazi:
Great work Adib.

PR #548, opened 2019-07-19
Adding type to R/O paths

Andrea-Campanella:


pkg/modelregistry/modelregistry.go
```diff
@@ -26,10 +27,19 @@ import (
...
 )
 
+// ReadOnlySubPathMap abstracts the read only subpath
+type ReadOnlySubPathMap map[string]change.ValueType
+
+// ReadOnlyPathMap abstracts the read only path
+type ReadOnlyPathMap map[string]ReadOnlySubPathMap
+
+// GlobalReadOnlyPaths abstracts the map for the read only paths
+type GlobalReadOnlyPaths map[string]ReadOnlyPathMap
```
SeanCondon: I was wondering if this really needs its own type - it's only used once

pkg/modelregistry/modelregistry.go
```diff
@@ -200,3 +240,58 @@ func extractnamespace(dirEntry *yang.Entry, parentNs string) string {
 	}
 	return parentNs
 }
+
+//Paths extract the rad only path up to the first read only container
```
SeanCondon: typo in comment

pkg/modelregistry/modelregistry_test.go
```diff
@@ -134,6 +135,614 @@ func Test_Schema(t *testing.T) {
 	}
 }
 
+func Test_State(t *testing.T) {
```
SeanCondon: It would be good on at least some of them to check that the valuetype returned was correct

pkg/modelregistry/modelregistry.go
Andrea-Campanella: we do that here and there, I think for easiness of read it can be, and it might also be used in other places.

PR #549, opened 2019-07-19
Integration test for model checking

ray-milkey:


test/integration/protoutils.go
```diff
@@ -38,24 +54,37 @@ func MakeProtoTarget(target string, path string) string {
...
-	gnmiPath := MakeProtoTarget(target, path)
+	gnmiPath := MakeProtoTarget(target, path, stripNamespaces)
 	protoBuilder.WriteString(gnmiPath)
 	return protoBuilder.String()
 }
 
+func makeProtoValue(value string) string {
+	var protoBuilder strings.Builder
+
+	if strings.Contains(value, ":") {
```
SeanCondon: I think if you made the string value "123456" as "string_val:123456" above, you would not need this if statement at all. What if the string value you set had a MAC address or IPv6 address in it - you wouldn't be handling it right

test/integration/modelstest.go
```diff
@@ -0,0 +1,80 @@
...
+	device := env.GetDevices()[0]
+
+	// Data to run the test cases
+	testCases := []struct {
+		description   string
+		path          string
+		value         string
+		expectedError string
+	}{
+		{description: "Unknown path", path: unknownPath, value: "123456", expectedError: "JSON contains unexpected field no-such-path"},
```
SeanCondon: create as "string_val:123456"

test/integration/protoutils.go
ray-milkey: I split out the type and the value into separate entities. See if you like it.

PR #550, opened 2019-07-22
Conversion of Json document in to native format - part 1 intermediate

SeanCondon:


pkg/store/json2Values.go
```diff
@@ -0,0 +1,74 @@
+// Copyright 2019-present Open Networking Foundation.
```
Andrea-Campanella: I know I might be old fashioned, but I'd go for `jsonToValues.go`

pkg/store/json2Values.go
```diff
@@ -0,0 +1,74 @@
...
+				changes = append(changes, o)
+			}
+		}
+	case string:
+		newCv := change.ConfigValue{Path: parentPath, TypedValue: *change.CreateTypedValueString(value)}
+		changes = append(changes, &newCv)
+	case float64:
+		newCv := change.ConfigValue{Path: parentPath, TypedValue: *change.CreateTypedValueFloat(float32(value))}
+		changes = append(changes, &newCv)
+	default:
```
Andrea-Campanella: is there a boolean type ?

pkg/store/json2Values.go
SeanCondon: done

pkg/store/json2Values.go
SeanCondon: added - good catch

Issue #547, opened 2019-07-18
Improve debugging in onit 


adibrastegarnia:
**Is your feature request related to a problem? Please describe.**
The debugging of containers in onit needs improvements. In addition, we need to document the debugging procedure for the developers. 

**Describe the solution you'd like**
Improve the debugging in onit and add working examples of debugging onos-config-debug in documents.

kuujo:
Yes this should be in the next demo thanks :-)

adibrastegarnia:
I think we can close this since we addressed them at : #551 #546 . In case we need to add more examples, we will refer to it.

PR #551, opened 2019-07-22
Add a debugging example using Golang IDE

adibrastegarnia:
This PR adds an example of debugging of onos-config using Goland IDE to address issue #547


docs/debugging.md
```diff
@@ -1,7 +1,16 @@
 # Debug onos-config in Onit Using Delve
 
-## Debugging Example #1
-- In this tutorial, we assume you already started an onit cluster. If you haven't done that yet, refer to [testing](testing.md) for more details. 
+## Debugging Example #1: Use Goland IDE (Recommended):
+- In this example, we assume you already started an onit cluster. If you haven't done that yet, refer to [testing](testing.md) for more details.
+
+- First, you need to configure your [Goland](https://www.jetbrains.com/go/) IDE to connect to the remote Delve debugger which is running in the onos-config 
+pod. You can do that by going to *Run | Edit Configurations… | + | Go Remote* and configuring the host (in this example *localhost*) and port (in this example port **40000**) your remote debugger is listening on.
```
Andrea-Campanella: is this port fixed or does it change every time ? if so is there a way to get it ?

adibrastegarnia: This is the port that is used in the debug image to start the dlv server and it is fixed.

adibrastegarnia:
@kuujo @Andrea-Campanella 

This PR is ready for review. If you are looking for a specific debugging feature that should be documented, please let me know. I tried to cover the most important ones.

docs/debugging.md
```diff
@@ -1,239 +1,297 @@
...
+  ```
+
+- Suppose we want to add a simulator and debug [synchronizer](https://github.com/onosproject/onos-config/blob/master/pkg/southbound/synchronizer/synchronizer.go)
+code by setting a few breakpoints in different lines of its file as follows:
+
+  ![goland_ide_set_breakpoints](images/goland_debug_set_breakpoints.gif) 
+                 
+- As you can see above, after setting the breakpoints, you should be able to see some messages
+in the onos-config pod logs that confirm you set the breakpoints. When we add a simulator
+the Goland IDE points to the first breakpoint when is hit and you can go through the list
```
Andrea-Campanella: I'd reformulate with `the Goland IDE hits the first breakpoint that was set. At this point you can go....`

docs/debugging.md
```diff
@@ -1,239 +1,297 @@
...
+code by setting a few breakpoints in different lines of its file as follows:
+
+  ![goland_ide_set_breakpoints](images/goland_debug_set_breakpoints.gif) 
+                 
+- As you can see above, after setting the breakpoints, you should be able to see some messages
+in the onos-config pod logs that confirm you set the breakpoints. When we add a simulator
+the Goland IDE points to the first breakpoint when is hit and you can go through the list 
+of variables and goroutines using the debugger dashboard. 
+
+- Suppose we want to have tracepoints instead breakpoints. That means they don't
```
Andrea-Campanella: missing `of` before breakpoints

Andrea-Campanella: substitute `That means they` with `tracepoints don't cause.....`

adibrastegarnia:
> Awesome work! Minor changes requested.

Andrea, Thanks for catching the typos and mistakes. I think I addressed your comment. Please take a look and let me know if more changes needed.

adibrastegarnia:
@kuujo 

Whenever you get a chance take a look and then we merge.

Issue #553, opened 2019-07-22
Support multiple ports via onit debug


kuujo:
**Describe the bug**
The `onit debug` command only supports forwarding a single port to an onos-config node.

**To Reproduce**
1. Run `onit create cluster --nodes 2` to create a 2 node cluster
2. Run `onit debug` for each node in the cluster
3. See error

**Expected behavior**
`onit` should not specify the local port to forward to the node being debugged. Instead, we should forward to `:40000` and allow the k8s client to select the local port. We may even be able to forward multiple local ports to each pod in a multi-node cluster with the `onit debug` command when no node ID is provided.

adibrastegarnia:
I will work on it soon.

PR #556, opened 2019-07-23
Support multiple ports via onit debug

adibrastegarnia:
@kuujo 
This PR is to address issue #553. To support multiple ports via onit debug, there are different ways to tackle the problem:

1- k8s port forwarder code already handles random local port assignment ( [Port Forwarder ](https://github.com/kubernetes/client-go/blob/master/tools/portforward/portforward.go)) That means if we don't provide any local port (i.e. Use this format :40000, or 0:40000), it will assign an available random port for the local port. The good thing about this approach is that we don't need to handle it in our code but the downside is that the assigned ports can be in any order because it is out of our control. In this approach, we can also allow the user assigns a local port as well instead of the system assigns a random local port where a specific order of assigned local ports is needed. 

2- Keep the information about the local ports assigned for each resource in the onit code and check for an open port starting from 40000 and provide that local port to the k8s port forwarder. 

Another thing that we need to consider is this: In the orginal issue we have the following statement:
**We may even be able to forward multiple local ports to each pod in a multi-node cluster with the onit debug command when no node ID is provided.**

It is possible to have that but it makes implementation a little bit complex. The reason for that is when we want to call *portForward* function then we need to start a goroutine  for call and to handle the error that *portForward* function returns, we will need to use an error channel.  Should we go for that complexity?

### DIFF
--- a/test/cli/cli.go
+++ b/test/cli/cli.go
@@ -22,6 +22,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"sync"
 	"text/tabwriter"
 	"time"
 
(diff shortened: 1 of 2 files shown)


adibrastegarnia:
We take care of it in onit repo.

PR #574, opened 2019-07-26
Remove namespaces 

### LINKED ISSUE #573: Eliminate namespaces from paths
As per reply here https://github.com/openconfig/ygot/issues/300 the namespaces can be fully eliminated from paths and their usage is not needed.

Remove all namespace, leave the : stripping in strPathUtils

Andrea-Campanella:
This patch removes namespace and connected methods from onos-config.
This is due to ygot not supporting namespaces and for simplification of operations.
Please see more info https://github.com/openconfig/ygot/issues/300 and in #573 

Fixes #573

### DIFF
--- a/docs/gnmi.md
+++ b/docs/gnmi.md
@@ -25,7 +25,7 @@ Use `gnmi_cli -get` to get configuration for a particular device (target) from t
 > If config from several devices are required, several paths can be added
 ```bash
 gnmi_cli -get -address localhost:5150 \
-    -proto "path: <target: 'localhost-1', elem: <name: 'openconfig-system:system'> elem: <name: 'clock' > elem: <name: 'config'> elem: <name: 'timezone-name'>>" \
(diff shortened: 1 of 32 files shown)


docs/gnmi.md
```diff
@@ -25,7 +25,7 @@ Use `gnmi_cli -get` to get configuration for a particular device (target) from t
 > If config from several devices are required, several paths can be added
 ```bash
 gnmi_cli -get -address localhost:5150 \
-    -proto "path: <target: 'localhost-1', elem: <name: 'openconfig-system:system'> elem: <name: 'clock' > elem: <name: 'config'> elem: <name: 'timezone-name'>>" \
+    -proto "path: <target: 'localhost-1', elem: <name: 'system'> elem: <name: 'clock' > elem: <name: 'config'> elem: <name: 'timezone-name'>>" \
```
SeanCondon: Maybe a note is needed in this file that namespaces are not involved at all - but make it sound like a positive thing

PR #576, opened 2019-07-29
Added in wildcard matching on Get and Subscribe on Northbound

SeanCondon:
Support the '*' and '...' in paths

### DIFF
--- a/pkg/northbound/gnmi/subscribe.go
+++ b/pkg/northbound/gnmi/subscribe.go
@@ -28,6 +28,7 @@ import (
 	"google.golang.org/grpc/status"
 	"io"
 	log "k8s.io/klog"
+	"regexp"
 	"time"
 )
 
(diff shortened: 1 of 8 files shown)


pkg/northbound/gnmi/subscribe.go
```diff
@@ -126,11 +127,11 @@ func listenOnChannel(stream gnmi.GNMI_SubscribeServer, mgr *manager.Manager, has
 		} else {
 			subs := subscribe.Subscription
 			//FAST way to identify if target and subscription is present
-			subsStr := make(map[string]struct{})
+			subsStr := make([]*regexp.Regexp, 0)
```
Andrea-Campanella: I'd suggest we keep it to a map, looking on the web it seems the the map with the struct{} is faster to index against then the list for linear search. that's why I sudden it in the first place. also keeps code complexity lower

SeanCondon: Map gives no advantage as we don't ever lookup by RegExp. We have to test each path to see if it rexexp matches any one of the subs.

Refactored the iteration over regexps out in to its own function

PR #577, opened 2019-07-29
Fixed error in config and added doc for wildcard

SeanCondon:


docs/gnmi.md
```diff
@@ -63,6 +63,47 @@ Use a proto value like:
...
+        -timeout 5s -alsologtostderr \
+        -client_crt pkg/southbound/testdata/client1.crt \
+        -client_key pkg/southbound/testdata/client1.key \
+        -ca_crt pkg/southbound/testdata/onfca.crt
+```
+> If the device is connected and the OperationState cache is populated this returns
+> 4 values - **eth1** config and state enabled values and **admin** config and
+> state enabled values.
+
+To retrieve both the config and state values of both then additionally the use
```
SeanCondon: this is repeated

PR #583, opened 2019-07-29
Refactoring of AdminService to ConfigAdminService

BILLYBOBJOE2664:


pkg/northbound/proto/admin.proto
```diff
@@ -83,8 +83,8 @@ message RollbackResponse {
     string message = 1;
 }
 
-// AdminService provides means for enhanced interactions with the configuration subsystem.
-service AdminService {
+// COnfigAdminService provides means for enhanced interactions with the configuration subsystem.
```
SeanCondon: typo in **COnfigAdminService** not the same as **ConfigAdminService**

PR #591, opened 2019-07-30
Add the golangci-lint meta linter 

ray-milkey:
This change adds the golangci-lint meta linter to onos-config builds.

golangci-lint allows us to use a number of different linters, and configure how they run.

A standard set of linters are enabled, and the problems they found have been addressed.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,17 @@
+linters:
+  enable:
+    - gofmt
+    - gocyclo
+    - golint
+    - misspell
+    - typecheck
+    - errcheck
+run:
+  skip-dirs:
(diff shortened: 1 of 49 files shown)


ray-milkey:
> Looks good. I wonder why they insist on the _, _ := foo() explicit throw-away assignment. I understand _, v := foo(), but not _, _ := foo(). There must be a reason though.

I think the idea is that a dangling error could be a bug; if the implementer says "Hey I am ignoring this" it's probably not. Sometimes the errors are given in two ways, like a nil return value and an error condition, and we ignore the error in favor of the nil pointer check.

PR #602, opened 2019-08-02
Bump timeout on subscription tests - these often fail on Travis builds

ray-milkey:
Trying to fix flakiness in the NB subscribe tests

### DIFF
--- a/pkg/northbound/gnmi/subscribe_test.go
+++ b/pkg/northbound/gnmi/subscribe_test.go
@@ -468,7 +468,7 @@ func assertUpdateResponse(t *testing.T, responsesChan chan *gnmi.SubscribeRespon
 		assert.Equal(t, pathResponse.Elem[1].Name, path2)
 		assert.Equal(t, pathResponse.Elem[2].Name, path3)
 		assert.Equal(t, response.GetUpdate().GetUpdate()[0].Val.GetUintVal(), uint64(value))
-	case <-time.After(1 * time.Second):
+	case <-time.After(5 * time.Second):
 		log.Error("Expected Update Response")
 		t.FailNow()
 	}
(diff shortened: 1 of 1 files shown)


pkg/northbound/gnmi/subscribe_test.go
```diff
@@ -140,7 +140,7 @@ func Test_SubscribeLeafStream(t *testing.T) {
 	}()
 
 	//FIXME Waiting for subscribe to finish properly --> when event is issued assuring state consistency we can remove
-	time.Sleep(100000)
+	time.Sleep(15 * time.Second)
```
SeanCondon: I think 15 seconds is a bit excessive - because this means every time the test is run it will stop here for 15 seconds

in my own pull request I changes to 100*time.Millisecond and it worked

The original value above is in nano seconds, so 100ms is a 1000 fold increase

pkg/northbound/gnmi/subscribe_test.go
```diff
@@ -294,7 +294,7 @@ func Test_ErrorDoubleSubscription(t *testing.T) {
 		}
 	}()
 	//FIXME Waiting for subscribe to finish properly --> when event is issued assuring state consistency we can remove
-	time.Sleep(100000)
+	time.Sleep(15 * time.Second)
```
SeanCondon: same

pkg/northbound/gnmi/subscribe_test.go
```diff
@@ -335,7 +335,7 @@ func Test_Poll(t *testing.T) {
 	}()
 
 	serverFake.Signal <- struct{}{}
-	time.Sleep(100000) //waiting before sending fake poll
+	time.Sleep(15 * time.Second) //waiting before sending fake poll
```
SeanCondon: same

pkg/northbound/gnmi/subscribe_test.go
```diff
@@ -387,7 +387,7 @@ func Test_SubscribeLeafStreamDelete(t *testing.T) {
 	}()
 
 	//FIXME Waiting for subscribe to finish properly --> when event is issued assuring state consistency we can remove
-	time.Sleep(100000)
+	time.Sleep(15 * time.Second)
```
SeanCondon: same

PR #615, opened 2019-08-07
Refactor topocache with DeviceService client

kuujo:
This PR refactors the topocache implementation to use onos-topo's `DeviceService`. Note that this is a very early iteration. The implementation replaces most of the abstractions from topocache with Protobuf generated messages. The topocache no longer exposes functions for modifying the topology, but instead only streams topology events from the `DeviceService` to the topo event channel. Just looking opening this PR to allow for comment if anyone dislikes the direction.


kuujo:
BTW this will fail miserably until the other topo PRs are reviewed and merged

kuujo:
A couple notes on code refactoring. I sense there’s some refactoring that needs to be done for the `Event` abstraction. The patterns we use for a local, in-memory store are often not suitable to distributed stores for performance reasons. The existing code is reading the topocache when a device event is received, but that creates a consistency _and_ performance issues. The topocache can cache devices and still expose a read method, but even in that case the channel reader may read a different device state than the one for which it received an event. The `TopoEvent` really needs to include the device’s state as it was when the event occurred. This is a cheap change and guarantees consistency.

But I suspect there’s some general refactoring that needs to be done for the event abstraction. The struct is designed like an interface, so it could just be an interface. I’m not really sure the abstraction is even needed at this point. But at least, it seems there’s no mechanism to differentiate creates/updates/deletes, and presumably those need specific handling that I don’t yet see. We can iterate on this so this can be addressed in another PR, but just pointing it out.

pkg/dispatcher/dispatcher.go
```diff
@@ -25,25 +25,24 @@ package dispatcher
 import (
 	"fmt"
 	"github.com/onosproject/onos-config/pkg/events"
-	"github.com/onosproject/onos-config/pkg/southbound/topocache"
 	log "k8s.io/klog"
 )
 
 // Dispatcher manages SB and NB configuration event listeners
 type Dispatcher struct {
-	deviceListeners        map[topocache.ID]chan events.ConfigEvent
```
SeanCondon: I kind of liked using topocache.ID - I know it was only an alias for string, but it was a useful reassurance from the compiler that the right type was being used. In my opinion it improved readability

pkg/events/events.go
```diff
@@ -75,6 +75,7 @@ type Event struct {
 	time      time.Time
 	eventtype EventType
 	values    map[string]string
+	object    interface{}
```
SeanCondon: we had not considered passing an object in the event previously - but it looks logical

pkg/southbound/topocache/topocache.go
```diff
@@ -22,114 +22,69 @@ Until then this simple cache will load a set of Device definitions from file
...
-	Store       map[ID]Device
-	topoChannel chan<- events.TopoEvent
+	client deviceproto.DeviceServiceClient
 }
 
-// LoadDeviceStore loads a device store from a file - will eventually be from onos-topology
-func LoadDeviceStore(file string, topoChannel chan<- events.TopoEvent) (*DeviceStore, error) {
-	storeFile, err := os.Open(file)
+// LoadDeviceStore loads a device store
+func LoadDeviceStore(topoChannel chan<- events.TopoEvent, opts ...grpc.DialOption) (*DeviceStore, error) {
```
SeanCondon: maybe this could be renamed to CacheDeviceStore - it's basically a GET on onos-topo I'm guessing

pkg/dispatcher/dispatcher.go
kuujo: Oh I agree. It’s nice to have. But since the `Device`s are Protobuf generated structs, wed have to provide for the conversion between a type that supports `ID` and the Protobuf generated type. That will introduce some overhead to get the type we want. We could still wrap the `string` `Id` in `ID` when it’s used outside `Device`, but that may affect readability as well by making the two equal values appear different:

kuujo: I think I actually found a way to get the Protobuf generator to generate the `ID` type for Go code. I will experiment with it and see if we can keep it.

kuujo:
Closing this to try again

PR #620, opened 2019-08-08
Added OpStateDiags

BILLYBOBJOE2664:


pkg/northbound/proto/diags.proto
```diff
@@ -77,6 +81,15 @@ message Configuration {
     repeated string changeIDs = 7;
 }
 
+// Change is a descriptor of a submitted configuration change targeted at a single device.
```
tomikazi: Probably want to update the comment here.

pkg/northbound/proto/diags.proto
```diff
@@ -87,3 +100,14 @@ service ConfigDiags {
 }
 
 
+service OpStateDiags {
```
tomikazi: We don't want to create a whole new interface/service.

pkg/northbound/proto/diags.proto
```diff
@@ -87,3 +100,14 @@ service ConfigDiags {
...
+    // GetChanges returns a stream of submitted changes objects.
+    rpc GetChanges (ChangesRequest) returns (stream Change);
+
+    // GetConfigurations returns a stream of submitted configurations aimed at individual devices.
+    rpc GetConfigurations (ConfigRequest) returns (stream Configuration);
+
+    // GetConfigurations returns a stream of submitted configurations aimed at individual devices.
+    rpc GetOpState (OpStateRequest) returns (stream OpState);
```
tomikazi: Just add this method to the existing ConfigDiags service above.

PR #628, opened 2019-08-11
Updated with added rpc to configdiags

BILLYBOBJOE2664:


SeanCondon:
these changes are in already

PR #630, opened 2019-08-12
Replace topo cache implementation with topo service

kuujo:
This PR is a second attempt at replacing the topocache implementation with the topo service client. Needs to be tested in onit before we can merge it.

The code is just about finished, just one of the tests fails, but it's the result of a bug in the test that was fixed by my changes to the test. Previously, the test data set `User` when the field was named `Usr`, and this resulted in the test passing when it should have failed had the data been correctly set.


cmd/onos-config/onos-config.go
```diff
@@ -145,6 +156,27 @@ func main() {
 	}
 }
 
+func newDialOptions(clientCrtPath, clientKeyPath *string) ([]grpc.DialOption, error) {
```
SeanCondon: I think you can reuse **HandleCertArgs()** from pkg/certs/clientcerts.go instead. And it has the advantage of loading a default client1.crt and client1.key internally if no argument is given

PR #643, opened 2019-08-20
Enable slash char in the key name 

Andrea-Campanella:
Allows the key to have a `/` in it, e.g. eth1/0.
Remove two usages of `strings.split(path, "/")` in favour of utils.StrPath, which takes into account the `/`

### DIFF
--- a/configs/changeStore-sample.json
+++ b/configs/changeStore-sample.json
@@ -106,8 +106,8 @@
         }
       ]
     },
-    "m9jEA7cfJXhYyFZEyKfnqs9zQho=": {
-      "ID": "m9jEA7cfJXhYyFZEyKfnqs9zQho=",
+    "Ic0FLA5vVnTqL6893OtyNvzMpss=": {
+      "ID": "Ic0FLA5vVnTqL6893OtyNvzMpss=",
       "Description": "Original configuration for devicesim",
       "Created": "2019-05-21T08:43:01Z",
       "Config": [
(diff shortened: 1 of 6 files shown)


configs/changeStore-sample.json
```diff
@@ -131,6 +131,11 @@
           "Value": "AQ==",
           "Type": 4
         },
+        {
```
SeanCondon: you should have an additional config/name entry
```
    {
          "Path": "/interfaces/interface[name=eth1/1]/config/name",
          "Value": "ZXRoMS8x",
          "Type": 1
        },
```

pkg/northbound/gnmi/set_test.go
```diff
@@ -110,6 +110,86 @@ func Test_doSingleSet(t *testing.T) {
 	assert.Equal(t, (*change.TypedUint64)(&newChange.Config[0].TypedValue).Uint(), uint(16))
 }
 
+// Test_doSingleSet shows how a value of 1 list can be set on a target
+func Test_doSingleSetList(t *testing.T) {
+	server, _ := setUp()
+	var deletePaths = make([]*gnmi.Path, 0)
+	var replacedPaths = make([]*gnmi.Update, 0)
+	var updatedPaths = make([]*gnmi.Update, 0)
+	pathElemsRefs, _ := utils.ParseGNMIElements(utils.SplitPath("/cont1a/cont2a[lista/b=c]/leafList"))
```
SeanCondon: there's never going to be a slash in the key name like "lista/b" it can only ever be in the key value
It should be /cont1a/list2a[name=a/b]/tx-power to align with the test1@2018-02-20.yang

PR #645, opened 2019-08-22
Added in gRPC request to upload model plugin over gRPC

SeanCondon:
I created this to be able to **upload** a model plugin to the onos-config through gRPC (rather than having to load the file by sftp it on to the onos-config server first and then **load** it).

I had to use streaming of the file for this as the Max message size is 4MB and the Model Plugin is >20MB.

Using streaming from client -> server like this is not supported by [grpc-web](https://www.freecodecamp.org/news/how-to-use-grpc-web-with-react-1c93feb691b5/) so this means it's not possible to run this from onos-gui.

The question is then is this useful? It allows us to upload the model plugin (potentially) from the onos-cli, but we would still have to sftp it to there inside the cluster, as we don't have external access at the moment. It is a slightly better story though.


pkg/northbound/admin/admin.go
```diff
@@ -54,6 +57,64 @@ func (s Server) RegisterModel(ctx context.Context, req *RegisterRequest) (*Regis
...
+	}
+	defer f.Close()
+
+	// while there are messages coming
+	i := 0
+	for {
+		chunk, err := stream.Recv()
+		if err != nil {
+			if err == io.EOF {
+				goto END
```
Andrea-Campanella: would break be enough here ? I don't see the point fo the goto

SeanCondon: I copied this from an example on the web - I'll use break instead

Issue #647, opened 2019-08-22
CLI hangs after a gNMI Set request from onos-ztp


tomikazi:
**Describe the bug**
The CLI to onos-config hangs after onos-ztp makes a Set request.

**To Reproduce**
1. Run 'onit create cluster'
2. Run 'onit add app --image onosproject/onos-ztp:latest onos-ztp'
3. From onos-ztp project dir, run 'kubectl -n $(onit get cluster) cp test/samplejson/stratum.json onos-cli-...:/tmp/stratum.json
4. Run 'onit onos-cli'
5. From onos-cli shell run 'onos ztp add role /tmp/stratum.json'
6. From onos-cli shell run 'onos config get configs'
7. From onos-cli shell run 'onos topo add device foo --target foo --type Stratum --version 1.0.0 --role leaf'
8. From onos-cli shell run 'onos config get configs'

Now you see hang.

tomikazi:
I'm seeing crashes (though at a slow rate) after the request is issued:
```onos-config-7dd9f44b7c-qkvrx         0/1     CrashLoopBackOff   21         102m```

kuujo:
that should be `kubectl -n $(onit get cluster) cp test/samplejson/stratum.json onos-cli-...:/tmp/stratum.json` (the onos-cli pod)

kuujo:
I _think_ this should be fixed but I'll let @tomikazi verify that and close this

SeanCondon:
I verified this works with a name **foo1234** instead of foo. It also works with a 3 letter long device such as "abc"

SeanCondon:
While the CLI does not hang for me it is also clear that the change is not actually accepted by the onos-config either.
This is because the change is made in that 30 second window from when onos-config is first told the device exists and the when it times out on connecting to that device. The change is rejected because the change can't be sent to the device.

A design change is necessary to allow changes to always be taken regardless of the state of the device. Keeping this issue open for the moment

Andrea-Campanella:
I believe this fixed by the new atomix based controllers workflow

PR #651, opened 2019-08-22
Ensure validation error is handled when creating configuration

kuujo:
This can result in a panic when the target is invalid.


kuujo:
#647

PR #677, opened 2019-09-19
Remove the redundant docs from onos-config since we moved them to ono…

adibrastegarnia:
Since we moved some of the common documents such as developer guidelines to onos-docs, we should remove them from this repo. Before merge this lets first the previous PR to update travis to trigger onos-docs job on merge commit on this repo.


adibrastegarnia:
@tomikazi 
Ready for your review. I already moved the general docs to onos-docs. I also addressed your comment.

PR #684, opened 2019-09-23
Convert CLI Run() methods to RunE() to allow for execution in tests

ray-milkey:
In this PR I changed the instances of Run() methods in the CLI to RunE() methods. Run() methods simply exit when they complete, making testing impossible. With this change, we can drive these methods via unit tests.

### DIFF
--- a/pkg/cli/changes.go
+++ b/pkg/cli/changes.go
@@ -43,12 +43,12 @@ func getGetChangesCommand() *cobra.Command {
 		Use:   "changes [<changeId>]",
 		Short: "Lists records of configuration changes",
 		Args:  cobra.MaximumNArgs(1),
-		Run:   runChangesCommand,
+		RunE:  runChangesCommand,
 	}
 	return cmd
 }
 
-func runChangesCommand(cmd *cobra.Command, args []string) {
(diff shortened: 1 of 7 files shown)


pkg/cli/changes.go
```diff
@@ -43,12 +43,12 @@ func getGetChangesCommand() *cobra.Command {
...
 		Short: "Lists records of configuration changes",
 		Args:  cobra.MaximumNArgs(1),
-		Run:   runChangesCommand,
+		RunE:  runChangesCommand,
 	}
 	return cmd
 }
 
-func runChangesCommand(cmd *cobra.Command, args []string) {
+func runChangesCommand(cmd *cobra.Command, args []string) error {
```
kuujo: It looks to me like this function can be simplified. We don't even need to read this stream in a separate goroutine. Since it's a unary request and streaming response, we can just read the stream in the calling goroutine:
```go
stream, err := client.GetChanges(context.Background(), changesReq)
...
for {
    response, err := stream.Recv()
    if err == io.EOF {
        return
    }
    if err != nil {
        return err
    }
    ...
}
```

PR #686, opened 2019-09-24
Refactoring of events to be defined as interfaces not structs Fixes #680

### LINKED ISSUE #680: GetOpState on NBI gRPC diags should support deletion of items from opState cache
**Describe the bug**
Currently the OpState cache is retrieved by calling GetOpState on the NBI gRPC interface. This returns the complete list of items. It has a subscribe option to choose whether further updates should be sent after the initial batch. This all works fine as long as the further updates modify or add existing items in the initial batch. If an update indicating an item was deleted however (as might happen if a stratum interface is deleted) then this is not handled.

**To Reproduce**
```
onos config get opstate devicesim-1 --subscribe
```
Then delete an interface from Stratum and observe how no message is sent for the deletion

**Expected behavior**

**Logs, text or screenshots**

**Additional context**
This should follow the pattern that onos-topo uses where every result has an added, updated or deleted flag on it

SeanCondon:
Created this Pull request again after I accidentally deleted it (doh)

### DIFF
--- a/pkg/dispatcher/dispatcher.go
+++ b/pkg/dispatcher/dispatcher.go
@@ -57,7 +57,7 @@ func (d *Dispatcher) Listen(changeChannel <-chan events.ConfigEvent) {
 
 	for configEvent := range changeChannel {
 		log.Info("Listener: Event ", configEvent)
-		deviceChan, ok := d.deviceListeners[device.ID(events.Event(configEvent).Subject())]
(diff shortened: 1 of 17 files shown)


SeanCondon:
> I think it looks great. Much cleaner. But I have one suggestion: we have a lot of `Create*` functions for the events. The convention in Go is to use `New*`, and I think we should be doing that here.
> 
> https://golang.org/doc/effective_go.html

Good idea - will do that

PR #689, opened 2019-09-24
Remove all explicit exit conditions and os.Stdout references from the onos-config CLI

ray-milkey:
This change removes some explicit references to os.Exit() and os.Stdout that make the onos-config CLI untestable.

### DIFF
--- a/pkg/cli/changes.go
+++ b/pkg/cli/changes.go
@@ -49,7 +49,12 @@ func getGetChangesCommand() *cobra.Command {
 }
 
 func runChangesCommand(cmd *cobra.Command, args []string) error {
-	client := diags.NewConfigDiagsClient(getConnection())
+	clientConnection, clientConnectionError := getConnection()
+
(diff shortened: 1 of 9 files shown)


pkg/cli/changes.go
```diff
@@ -49,7 +49,12 @@ func getGetChangesCommand() *cobra.Command {
 }
 
 func runChangesCommand(cmd *cobra.Command, args []string) error {
-	client := diags.NewConfigDiagsClient(getConnection())
+	clientConnection, clientConnectionError := getConnection()
```
kuujo: FWIW the idiomatic Go naming pattern would have this line as:
```go
conn, err := getConnection()
if err != nil {
    return err
}
```

kuujo: Also FWIW this is contrary to the classic literature on writing clean and maintainable code. 🤷‍♂

PR #693, opened 2019-09-25
Fix links to onos-docs markdown files

adibrastegarnia:
I just noticed that I forgot to include /developers in the path of onos-docs links

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,12 +61,12 @@ provided as additional layers:
 
 
 ## Additional Documentation
-* [How to contribute](https://github.com/onosproject/onos-docs/docs/contributing.md) to onos-config project
+* [How to contribute](https://github.com/onosproject/onos-docs/blob/master/docs/developers/contributing.md) to onos-config project
 * [How to build](docs/build.md) onos-config server, related commands and Docker image
 * [How to run](docs/run.md) onos-config server and related commands
 * [How to deploy](docs/deployment.md) onos-config in a Kubernetes cluster
 * [How to test](https://github.com/onosproject/onos-test/blob/master/docs/testing.md) onos-config integration tests with onit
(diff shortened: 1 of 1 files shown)


adibrastegarnia:
> I thought these were going to be relative links to the new docs.onosproject.org?

@SeanCondon 
This page is not in the docs. Tom suggested to keep these links in onos-config for now but point to the docs in onos-docs in case someone visit the github repo first without checking the  website. We should decide about what we we would like to have in the first page of each subsystem. We can have a pointer in the first page of each repo to point to a readme file under /docs of each repo that explains the main objectives of each subsystem (like what you did for  onos-gui) or I can include the README file of each subsystem in the build process.

SeanCondon:
It's tempting to keep useful information in the README at the root of the project, because that's where a lot of people will see it first.

That's what I was trying to achieve in onos-gui but I should also have said in this top level README that the docs are readable at https//docs.onosproject.org too

adibrastegarnia:
> It's tempting to keep useful information in the README at the root of the project, because that's where a lot of people will see it first.
> 
> That's what I was trying to achieve in onos-gui but I should also have said in this top level README that the docs are readable at https//docs.onosproject.org too

What we can do is to exclude the first page in building the docs.onosproject.org as we do now and keep it just for the purpose if someone visits the github page first but have an intro doc under docs as well to explain just the main objectives of the subsystem without pointing to extra links such developer guidelines, etc.

SeanCondon:
Sounds good - are you changing this PR then or do you want to go ahead with it as is?

adibrastegarnia:
@SeanCondon Lets merge this to make the links in the first page valid for now then we should decide what we would like to do with the first pages.

PR #694, opened 2019-09-25
Unit tests for onos-config CLI plugin commands

ray-milkey:


pkg/cli/rollback_test.go
```diff
@@ -26,7 +26,7 @@ func Test_rollback(t *testing.T) {
 	outputBuffer := bytes.NewBufferString("")
 	CaptureOutput(outputBuffer)
 
-	setUpMockClients()
+	setUpMockClients(nil)
 	rollback := getRollbackCommand()
```
SeanCondon: the `nil` here is just temporary until you do the next part - right?

ray-milkey: The nil is an annoyance that I would love to get rid of, but have not figured out how yet. The List Registered Models Client has to be initialized before the Config Admin Service Client is made. This is because each command in the CLI makes the admin client it wants, and we don't want to mess with the command code for testing. The way it works now the test decides how it wants to mock the List client before calling the command. There are other sub clients too so this will get uglier. I hope to come up with a way to make it better.

pkg/cli/models_test.go
```diff
@@ -0,0 +1,109 @@
...
+			ReadOnlyPath: roPaths,
+			ModelData:    modelData,
+		}
+	}
+}
+
+var nextInfoIndex = 0
+
+func recvMock() (*admin.ModelInfo, error) {
+	if nextInfoIndex < len(modelInfo) {
```
SeanCondon: this is the part i was not understanding - nice

PR #695, opened 2019-09-26
Framework for a CLI opstate command unit test

ray-milkey:


pkg/cli/opstate_test.go
```diff
@@ -0,0 +1,96 @@
...
+	assert.NilError(t, err, "Error fetching opstate command")
+	outputString := outputBuffer.String()
+	output := strings.Split(strings.TrimSuffix(outputString, "\n"), "\n")
+
+	testCases := []struct {
+		description string
+		index       int
+		regexp      string
+	}{
+		{description: "Path 0", index: 2, regexp: `/root/system/path0\s+\|\(STRING\) value0 +\|`},
```
SeanCondon: nice

ray-milkey:
I looked at the subscription based API. It doesn't seem to change the CLI code much, which is what these tests are trying to test. I am going to leave it out for now. I think this is more appropriately tested inside the opstate client tests.

PR #701, opened 2019-09-27
Hold for device reply before notifying subscribers of a change.

### LINKED ISSUE #437: Event to changeChannel only when the device accepts the change.
Currently and event is sent out to the changeChannel when the config is stored in ONOS config. 
This does not reflect if the change has been applied to the device or not. If the device does not apply the change we are still sending out notifications about that change as if it was accepted, for example to subscriptions for that path. 
We should send out the event to the changeChannel only if the device accepted that change. IF the device is not present is ok sending it out as soon as it's stored.

Andrea-Campanella:
Leveraging the double listener mechanism (ChangesChannel and subscribe response from device channel) in NB subscribe.
If the device is there we wait for the reply from it. 
If the device is not replying we do not report any notification for the subscriptions.
If the device for which the subscription and the change arrive is not connected we immediately return.
Fixes #437

### DIFF
--- a/pkg/northbound/gnmi/subscribe.go
+++ b/pkg/northbound/gnmi/subscribe.go
@@ -23,6 +23,7 @@ import (
 	"github.com/onosproject/onos-config/pkg/store/change"
 	"github.com/onosproject/onos-config/pkg/utils"
 	"github.com/onosproject/onos-config/pkg/utils/values"
+	"github.com/onosproject/onos-topo/pkg/northbound/device"
 	"github.com/openconfig/gnmi/proto/gnmi"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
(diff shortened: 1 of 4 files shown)


pkg/northbound/gnmi/subscribe.go
```diff
@@ -174,19 +175,53 @@ func listenForUpdates(changeChan chan events.ConfigEvent, stream gnmi.GNMI_Subsc
...
-					}
-					err = buildAndSendUpdate(pathGnmi, target, &changeValue.TypedValue, stream)
-					if err != nil {
-						log.Error("Error in sending update path ", err)
-						resChan <- result{success: false, err: err}
-					}
-				}
+			//if the device is registered it has a listener, if not we assume the device is not in the system and
+			// send an immediate response
+			respChan, ok := mgr.Dispatcher.GetResponseListener(device.ID(target))
```
SeanCondon: Is there anyway we can do a unit test or integration test for this?

pkg/northbound/gnmi/subscribe_test.go
```diff
@@ -421,6 +423,82 @@ func Test_SubscribeLeafStreamDelete(t *testing.T) {
...
+	go func() {
+		cfg := <-configChan
+		assert.Assert(t, cfg.Applied())
+		go func() {
+			respChan <- events.NewResponseEvent(events.EventTypeAchievedSetConfig, targetStr, []byte(cfg.ChangeID()), "")
+		}()
+	}()
+	//Sending set request
+	go func() {
+		_, err = server.Set(context.Background(), setRequest)
```
SeanCondon: could this be a race? It might not be done by the time you get to statements below

pkg/northbound/gnmi/subscribe_test.go
```diff
@@ -421,6 +423,82 @@ func Test_SubscribeLeafStreamDelete(t *testing.T) {
 
 }
 
+// Test_SubscribeLeafStreamWithDeviceLoaded tests subscribing with mode STREAM for an existing device
+// and then issuing a set request with updates for that path
+func Test_SubscribeLeafStreamWithDeviceLoaded(t *testing.T) {
+	server, mgr := setUp()
+	mgr.DeviceStore = &topocache.DeviceStore{
```
SeanCondon: maybe should ask Ray about gomock to mock this up - I know he has been working with it lately

pkg/northbound/gnmi/subscribe_test.go
```diff
@@ -421,6 +423,82 @@ func Test_SubscribeLeafStreamDelete(t *testing.T) {
...
+	setRequest := &gnmi.SetRequest{
+		Delete:  deletePaths,
+		Replace: replacedPaths,
+		Update:  updatedPaths,
+	}
+
+	go func() {
+		cfg := <-configChan
+		assert.Assert(t, cfg.Applied())
+		go func() {
```
SeanCondon: A go routine inside a go routine - is this because you're concerned there's no listener?

PR #704, opened 2019-09-30
Device state propagation

### LINKED ISSUE #700: Handle Device ConnectionState
Currently there is no clear state of device connection, meaning that apps coudl get notified of a device presence even if the device is not actually there/can receive changes. 
Also onos-config could be caught in a corner when receiving configuration during device connection. 

IMG_3930
onos-topo and onos-config need to have a stronger interaction.
The proposed system is the following
0) onos-topo receives address of a device, saves it with "connecting" state
1) onos-topo bases on the device type ask all the subsystem to connect to the device with their channel, config shoudl be doing gNMI.
2) onos-config tries to open the connection
3) if connect is successful the device returns ok
4) onos-config notifies onos-topo about device being there
5) onos-topo notifies any listener that the device is actually there and change the state to available .

during 2-3 configuration for that device is blocked. 
onos-config will also check with onos-topo the status of the device every time some config is set.
- if the device is present in topo and state is connected for gNMI it will accept the change
- if the device is present in topo but state is connecting it will reject the change
- if the device is not present it will accept the change

onos-topo should keep information of state per each protocol the device can speak. 

Comments are welcome. 
Related to #518 and #459

Andrea-Campanella:
Propagates device state from config to topo, introducing notion of state.
Needs probably porting to protobuf with enum for state. 

Partially fixes #700

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,7 +13,7 @@ require (
 	github.com/golang/protobuf v1.3.2
 	github.com/mitchellh/go-homedir v1.1.0
 	github.com/onosproject/onos-control v0.0.0-20190715190020-706a2ee0d37b // indirect
-	github.com/onosproject/onos-topo v0.0.0-20191003071657-8bcd1fd812c5
+	github.com/onosproject/onos-topo v0.0.0-20191007162302-b2f95b144981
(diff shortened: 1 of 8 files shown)


Andrea-Campanella:
Pending on https://github.com/onosproject/onos-topo/pull/44

PR #705, opened 2019-09-30
Added in unit tests for Synchronizer

SeanCondon:


pkg/southbound/synchronizer/factory_test.go
```diff
@@ -0,0 +1,140 @@
...
+		assert.Error(t, resp.Error(),
+			"could not create a gNMI client: Dialer(1.2.3.4:11161, 500ms): context deadline exceeded", "after gRPC timeout")
+		break
+	}
+
+	/*****************************************************************
+	 * Now it should have cleaned up after itself
+	 *****************************************************************/
+
+	//topoEventDelete := events.NewTopoEvent(device1.ID, events.EventItemDeleted, &device1)
```
Andrea-Campanella: was this left out on purpose ? or is it useless ?

pkg/southbound/mocks/target_mock.go
```diff
@@ -0,0 +1,112 @@
+// Code generated by protoc-gen-gogo. DO NOT EDIT.
```
ray-milkey: Can we rename this file to target_mock_test.go? If we do that, the coverage tools will not count the mocked code in here against us.

pkg/southbound/synchronizer/factory_test.go
ray-milkey: Do we need this commented out portion?

SeanCondon: leaving it out

SeanCondon: Leaving it out

pkg/southbound/mocks/target_mock.go
SeanCondon: Actually I can't - if I make this _test.go then the main code can't find it.
Also the linter complains I have a test file but nothing to test it on. I'd have to create a source file with dummy code.

PR #709, opened 2019-10-02
Updated synchronizer tests further

SeanCondon:
Brought coverage up to 54% for synchronizer - don't want to test the Stratum specific stuff too much - will be changing in next patch, and will bring coverage up there


Andrea-Campanella:
LGTM

PR #712, opened 2019-10-02
Add distributed stores and controllers

kuujo:
This PR refactors the onos-config subsystem to add distributed stores and use a level-triggered architecture for processing changes. It is incomplete. I'm opening it now and will push changes as they become ready. Once all the components needed for distributing configurations are in place, we will merge this PR and then begin refactoring the existing code to integrate the stores.


kuujo:
The first commit adds the protobuf files and generated code. I believe this can be implemented without the request store, which served primarily to provide ordering of events for consistency. This change instead uses a sequential index for `NetworkConfig`s which can be used by the config controller to deduce order.

This change also adds a `Snapshot` object. The snapshot object is a request and status of a snapshot. This portion is a work in progress.

kuujo:
Clicked the wrong button 😜

kuujo:
This PR now depends on https://github.com/onosproject/onos-topo/pull/42

That PR updates the Atomix client and moves some Protobuf code around. Since it’s not merged yet, this PR has a `replace` directive that will have to be removed before we can merge. We need to merge both the PRs and push a new Atomix image at the same time since they’ll all break each other.

pkg/types/change/change.proto
```diff
@@ -0,0 +1,119 @@
...
+    LEAFLIST_INT = 9;
+    LEAFLIST_UINT = 10;
+    LEAFLIST_BOOL = 11;
+    LEAFLIST_DECIMAL = 12;
+    LEAFLIST_FLOAT = 13;
+    LEAFLIST_BYTES = 14;
+}
+
+// Value is an individual Path/Value combination in a Change
+message Value {
```
SeanCondon: cool - you brought  this over

pkg/types/change/change.proto
```diff
@@ -0,0 +1,119 @@
...
+    ERROR = 0;
+
+    // UNAVAILABLE indicates the device was unavailable
+    UNAVAILABLE = 1;
+}
+
+// Change represents a configuration change to a single device
+message Change {
+    // The identifier of the change
+    string id = 1 [(gogoproto.customname) = "ID", (gogoproto.casttype) = "ID"];
```
SeanCondon: Are you thinking of an auto-generated hash for this? I don't think the user will be able to set it, as it's only created by the Network service.

Andrea-Campanella:
> This PR now depends on [onosproject/onos-topo#42](https://github.com/onosproject/onos-topo/pull/42)
> 
> That PR updates the Atomix client and moves some Protobuf code around. Since it’s not merged yet, this PR has a `replace` directive that will have to be removed before we can merge. We need to merge both the PRs and push a new Atomix image at the same time since they’ll all break each other.

@SeanCondon merged the Pr on topo I think you can remove the `replace` and push the atomix image

pkg/types/snapshot/snapshot.proto
```diff
@@ -0,0 +1,119 @@
...
+    // timestamp is the time at which to take the snapshot
+    google.protobuf.Timestamp timestamp = 7 [(gogoproto.stdtime) = true, (gogoproto.nullable) = false];
+
+    // created is the time at which the configuration was created
+    google.protobuf.Timestamp created = 8 [(gogoproto.stdtime) = true, (gogoproto.nullable) = false];
+
+    // updated is the time at which the configuration was last updated
+    google.protobuf.Timestamp updated = 9 [(gogoproto.stdtime) = true, (gogoproto.nullable) = false];
+}
+
```
SeanCondon: I was expecting a NetworkSnapshot (for the result)

pkg/types/change/change.proto
```diff
@@ -0,0 +1,119 @@
...
+    string device_version = 5;
+
+    // The current status of the configuration
+    Status status = 6;
+
+    // The reason the change failed
+    Reason reason = 7;
+
+    // An optional status message
+    string message = 8;
```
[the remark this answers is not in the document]
Andrea-Campanella: same holds true for any other optional value in any of the .proto files

kuujo:
All the ref storing is done and the final packages and interfaces are in place. Some notes on the final package structure:
• The `types` package contains all the Protobuf files and generated code used in stores and controllers
• The `store` package contains the stores for each layer
• The `controller` package contains the controllers for each layer

Within each of these packages the state is divided into two layers:
• The `network` layer is network wide state that can be handled by any node (typically by the leader of the cluster)
• The `device` layer is state that’s specific to a device and is handled by the master for that device

Within each of the `network` and `device` layers are packages for the three different request types:
• `change` is a configuration change
• `rollback` is a rollback if an existing configuration change
• `snapshot` is a snapshot of the set of changes

A type, store, and controllers package exists for each of these types of objects.

kuujo:
I also refactored the the controller logic itself so that is partitions requests onto per-device channels for concurrency. So each master will have a goroutine for changes, rollbacks, and snapshots, and optimistic concurrency control is used to avoid conflicts between those concurrent goroutines (though changes will be much more common than rollbacks and snapshots).

kuujo:
Closing this PR now and I'm beginning to open separate PRs for each of the interfaces in this PR. Will address review comments as I go along. Thanks!

PR #715, opened 2019-10-03
Updating ConvertJSON to handle non-openconfig yang model for opstate

SeanCondon:


pkg/modelregistry/jsonvalues/convertjson.go
```diff
@@ -109,9 +109,9 @@ func CorrectJSONPaths(jsonBase string, jsonPathValues []*change.ConfigValue,
 				break
 
 			} else if hasPrefixMultipleIdx(jsonPathWildIndex, modelPathWildIndex) {
-				jsonSubPath := jsonPathWildIndex[len(modelPathWildIndex):]
+				//jsonSubPath := jsonPathWildIndex[len(modelPathWildIndex):]
```
Andrea-Campanella: why was this commented out ? if it's not needed please remove

pkg/modelregistry/jsonvalues/convertjson.go
```diff
@@ -109,9 +109,9 @@ func CorrectJSONPaths(jsonBase string, jsonPathValues []*change.ConfigValue,
...
 
-				newTypeValue, err := assignModelType(paths, modelPath, jsonSubPath, jsonPathValue)
+				newTypeValue, err := assignModelType(paths, modelPath, jsonPathValue)
```
Andrea-Campanella: I see no difference in the usage of the method from before but the following lines are different, is it worth consolidating by extracting this line and the error evaluation or am I missing some logic.

pkg/modelregistry/jsonvalues/convertjson.go
SeanCondon: removed

pkg/modelregistry/jsonvalues/convertjson.go
SeanCondon: updated

PR #717, opened 2019-10-03
Unit tests for CLI devicetree command

ray-milkey:


pkg/cli/devicetree_test.go
```diff
@@ -0,0 +1,110 @@
...
+
+func diagChangesRecv() (*admin.Change, error) {
+	if changesRecvCount < 3 {
+		now := time.Now()
+		result := &admin.Change{
+			Time: &now,
+			Id:   "2Lo0ZC0wqhuwbnSF9px6kDvEEWI=",
+			Desc: "Change1",
+			ChangeValues: []*admin.ChangeValue{
+				{Path: "/a/b/c",
```
SeanCondon: To create a ChangeValue you could use change.NewChangeValue(path, value, isRemove) here, and create value with change.NewTypedValueString("VALUE")

Then with the ChangeValue(s) you could use change.NewChange and put that straight in to `result` - it will generate a hash - but that will stay consistent

ray-milkey: These changes would suffer from an impedance mismatch between admin.Change and store.Change. I would have to convert from the store based change to an admin based change, creating two different versions of the change. Additionally, this test is not testing store functionality, and actually has no store in it. All it is testing is that the formatting of the data is correct. That data itself has no meaning. I don't think the extra code to generate the data using the utility routines is necessary. We'd be replacing 10 lines of code to create the object directly with at least 25 lines of code to invoke the store routines. I would prefer not to do this.

PR #718, opened 2019-10-04
Improve manager package coverage

ray-milkey:


pkg/manager/manager_test.go
```diff
@@ -431,3 +435,83 @@ func TestManager_ComputeRollbackDelete(t *testing.T) {
...
+	//  Create an op state map with test data
+	device1ValueMap := make(map[string]*change.TypedValue)
+	change1 := change.NewTypedValueString(value1)
+	change2 := change.NewTypedValueString(value2)
+	device1ValueMap[path1] = change1
+	device1ValueMap[path2] = change2
+	mgrTest.OperationalStateCache[device1] = device1ValueMap
+
+	// Test fetching a known path from the cache
+	state1 := mgrTest.GetTargetState(device1, path1)
```
SeanCondon: This supports wildcards too - probably worth testing e.g. GetTargetState with /a/*/c

pkg/manager/manager_test.go
```diff
@@ -431,3 +435,83 @@ func TestManager_ComputeRollbackDelete(t *testing.T) {
...
+func TestManager_ValidateStoresReadOnlyFailure(t *testing.T) {
+	mgrTest, _, _ := setUp()
+
+	plugin := MockModelPlugin{}
+	mgrTest.ModelRegistry.ModelPlugins["TestDevice-1.0.0"] = plugin
+
+	roPathMap := make(modelregistry.ReadOnlyPathMap)
+	roSubPath1 := make(modelregistry.ReadOnlySubPathMap)
+	roSubPath1["/"] = change.ValueTypeSTRING
+	roPathMap["/cont1a"] = roSubPath1
```
SeanCondon: This should be `/cont1a/cont1b/leaf2c` because that's what the `config=false` is on in the YANG file. See 
https://github.com/onosproject/onos-config/blob/b00fcea7730ddec1b81171d3aabba64ef89b0f18/pkg/southbound/synchronizer/synchronizer_test.go#L61
for an example

PR #719, opened 2019-10-04
Add configuration change Protobuf files and related types

kuujo:
This PR adds the primary `change` types required to store and replicate configuration changes. The types are separated into two layers: `network` changes and `device` changes, and are stored in the `pkg/types` package which primarily contains generated code. Types are implemented as Protobuf files and code is generated with custom types using gogo. These types will be used in stores and controllers to apply changes.

Note that this change duplicates the `Change` type from the northbound API. This is intentional, as we don't want to affect existing code quite yet. Once these changes and their stores get in, we will evolve similar existing types out of the codebase.


pkg/types/network/change/types.proto
```diff
@@ -0,0 +1,82 @@
...
+syntax = "proto3";
+
+package onos.config.network.change;
+
+import "google/protobuf/timestamp.proto";
+import "gogoproto/gogo.proto";
+import "github.com/onosproject/onos-config/pkg/types/device/change/types.proto";
+
+// Status is the status of a network change
+enum Status {
```
SeanCondon: These appear in all your proto files - could they be reused from one file?

pkg/types/network/change/types.proto
```diff
@@ -0,0 +1,82 @@
...
+
+    // SUCCEEDED indicates the change was applied successfully
+    SUCCEEDED = 2;
+
+    // FAILED indicates the change failed
+    FAILED = 3;
+}
+
+// Reason is a failure reason
+enum Reason {
```
SeanCondon: Same here

pkg/types/device/change/types.proto
```diff
@@ -0,0 +1,122 @@
...
+    UNAVAILABLE = 1;
+}
+
+// Change represents a configuration change to a single device
+message Change {
+    // The identifier of the change
+    string id = 1 [(gogoproto.customname) = "ID", (gogoproto.casttype) = "ID"];
+
+    // index is the index of the configuration
+    uint64 index = 2 [(gogoproto.casttype) = "Index"];
```
SeanCondon: Does this `index` have the same meaning as in NetworkChange i.e.
`    // index is a monotonically increasing, globally unique index of the change`?

pkg/types/device/change/types.proto
```diff
@@ -0,0 +1,122 @@
...
+
+    // The configuration revision number
+    uint64 revision = 3 [(gogoproto.casttype) = "Revision"];
```
SeanCondon: Does this `revision` have the same meaning as below? If so add comment
```
    // revision is the change revision number
    // Each unique state of the change will contain a unique revision number which can be
    // used for optimistic concurrency control when updating or deleting the change state.
```

pkg/types/device/change/types.go
```diff
@@ -0,0 +1,66 @@
...
+func (i ID) GetDeviceID() device.ID {
+	return device.ID(string(i)[:strings.Index(string(i), separator)])
+}
+
+// GetDeviceVersion returns the device version to which the change is targeted
+func (i ID) GetDeviceVersion() string {
+	if !i.containsVersion() {
+		return ""
+	}
+	return string(i)[strings.Index(string(i), separator)+1 : strings.LastIndex(string(i), separator)]
```
SeanCondon: I'm not clear on how the string is to be constructed here? Maybe a few unit tests would illustrate it
From what I can see it's 
`devicename:version:index`?

pkg/types/network/change/types.proto
kuujo: I don't have a problem with using a single pair of enums. The reason they're implemented this way is because creating a separate status/reason enum for each type allows them to evolve separately if needed. If one pair of enums applies to all types then any change to the enums also needs to be valid for all types.

pkg/types/device/change/types.proto
kuujo: yes it does thanks

pkg/types/device/change/types.proto
kuujo: yes indeed thanks

pkg/types/device/change/types.go
kuujo: Yeah... the reason I didn't add unit tests for this is because adding a test to this package will test all the generated code. Instead, I will unit test it in the store code to be added after this is merged.

PR #720, opened 2019-10-04
Add configuration rollback Protobuf files and related types

kuujo:
This PR adds types needed to store and propagate rollback requests. Rollback types are implemented in Protobuf and code is generated with gogo to enable the usage of custom types. The types are separated into two layers: `network` rollbacks represent the rollback of a network change, and `device` rollbacks are per-device rollback notifications to be propagated to device masters.


kuujo:
@SeanCondon the `NetworkRollback` relates to a `NetworkChange` by its `change_id` field. 

Rollback requests do need to be stored for fault tolerance. This is just one way to do it. Without storing rollback requests, failures can lead to partial rollbacks. Using the stores also helps distribute the rollback to device masters on other nodes using the same pattern used for changes.

But this approach may be a relic of the use of the change log. The other approach can be to simply add statuses to for handling rollbacks to `NetworkChange` and `Change`. Then the `NetworkChange` and `Change` controllers can handle the rollbacks instead of implementing rollbacks separately. That may have some benefits since we would have better control over concurrency in a single controller. But ultimately both approaches provide the same feature using the same patterns, just different data models.

I can’t find a reason this could not be implemented with additional status values. I believe the use of a separate type is just a relic of the change log that was removed. But if we need a reliable change log we can add distinct requests for changes and rollbacks there.

I will close this and re-implement it in the change objects.

PR #721, opened 2019-10-05
Add leadership store implementation

kuujo:
This PR adds a simple leadership store plus tests. The changes here will be the basis for mastership election as well.


pkg/store/cluster/cluster.go
```diff
@@ -0,0 +1,31 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cluster
+
+import "os"
+
+const nodeIDEnv = "NODE_ID"
```
kuujo: @adibrastegarnia FYI we need to add this `NODE_ID` environment variable to onos-test

adibrastegarnia: @kuujo OK I will take care of it. But what is the value then?  In NodeInfo data structure we assign pod.Name to its ID but when the deployment is not created yet there is no ID unless we should generate a unique ID ourselves and assign it to the pod when it is creating (if it is possible not sure yet). Perhaps I am overthinking and it should be just "onos-config".

kuujo: @adibrastegarnia the value just needs to be a unique name string, but we should use the pod name. k8s provides features for passing this into the containers via environment variables, e.g.:
https://github.com/atomix/atomix-k8s-controller/blob/master/deploy/atomix-controller.yaml#L141-L144
Just use `NODE_ID` rather than `POD_NAME` and that should be it

PR #723, opened 2019-10-07
Add controller abstraction

kuujo:
This PR adds a controller abstraction for managing different types of resources in a control loop. 

A `Controller` receives events from a `Watcher` (e.g. the change store) and applies them to a `Reconciler`, which contains the application logic for handling a specific type of request (e.g. a change). `Controller`s provide fault tolerance to reconcilers by retrying requests/events in a loop until reconciliation is complete, and retry patterns may differ for different types of reconciler failures. If the reconciler returns an error, the controller will retry the request after an exponential backoff period. If the reconciler returns a boolean indicating that reconciliation failed, the controller will requeue the request to be processed after pending requests are processed to allow the cluster's state to progress before the reconciler retries the failed attempt.

`Controller`s can be configured to process events only on a single node in a cluster or partition events among many nodes. Additionally, each controller instance can be configured to partition individual requests among many goroutines for concurrency. This flexibility allows the same abstraction to be used by both cluster-wide controllers and device-specific controllers. In the case of cluster-wide controllers (e.g. a network change), the leadership store can be used to configure the controller to process events only on a single node. In the case of device-specific controllers (e.g. a device change), the mastership store can be used to distributed requests to the appropriate nodes and partition requests on each node into a goroutine-per-device.

The most important component of a controller is the configured `Reconciler`. A reconciler will be created for each type of object handled by onos-config. Even though the controller is often triggered by events, reconcilers are level triggered: they receive only an object identifier and use it to reconcile the current state of the cluster rather than depending on an event object for state. This allows reconciler requests to be retried and reordered without concern for consistency, and so it simplifies the failure recovery mechanisms required at lower levels of the system (e.g. Raft client retries, topo service retries, etc).


pkg/controller/activator_test.go
```diff
@@ -0,0 +1,63 @@
...
+	ch := make(chan leadership.Leadership)
+	err = store2.Watch(ch)
+	assert.NoError(t, err)
+
+	err = store1.Close()
+	assert.NoError(t, err)
+
+	<-ch
+
+	assert.True(t, <-ch2)
```
Andrea-Campanella: if for any reason the channels are blocked and no element is received this will hang the tests without failing them. I've found that 
```
	select {
	case response := <-ch:
		// assertions
	case <-time.After(5 * time.Second):
		log.Error("Expected Response")
		t.FailNow()
	}
```
enables to avoid such lock occasion

pkg/controller/filter_test.go
```diff
@@ -0,0 +1,88 @@
...
+	assert.False(t, filter1.Accept(types.ID(device2)))
+
+	ch := make(chan mastership.Mastership)
+	err = store2.Watch(device1, ch)
+	assert.NoError(t, err)
+
+	err = store1.Close()
+	assert.NoError(t, err)
+
+	<-ch
```
Andrea-Campanella: same comment about channels here

PR #725, opened 2019-10-07
Add support for tombstones to changes for rollback implementation

kuujo:
This PR modifies the data model for changes to support rollbacks. Rollbacks are implemented in changes by simply using a much simpler `Status` for each change. The `Status` of a change is no longer a simple enum, but is a message which provides:
* `state` - the previous `status` constants
* `reason` - a reason for the state
* `message` - an optional message
* `deleted` - a tombstone flag

The most important component of the data model is the `deleted` flag. This is a tombstone which indicates to the controller that the changes need to be rolled back. When the flag is set, the `state` represents the state of the rollback as opposed to the state of the change in the process of being applied. Once the `state` has progressed to `SUCCEEDED`, the change can be safely deleted from the store.


kuujo:
@Andrea-Campanella that’s actually why I added the `UNKNOWN` value to `Reason`. In the other enum `State` I left the `0` value as `PENDING` intentionally. All changes should start in the `PENDING` state whether the client sets it or not.

PR #727, opened 2019-10-08
Add device change store

kuujo:
This PR adds the change store for device changes. The store exposes a simple CRUD interface. Changes are stored with unique, monotonically increasing, per-device indexes, allowing changes to be replayed for a specific device. Optimistic locking is used for concurrency control.


pkg/store/change/device/store_test.go
```diff
@@ -0,0 +1,224 @@
...
+
+	// Verify events were received for the changes
+	changeEvent := <-ch
+	assert.Equal(t, devicechange.ID("device-1:1"), changeEvent.ID)
+	changeEvent = <-ch
+	assert.Equal(t, devicechange.ID("device-1:2"), changeEvent.ID)
+	changeEvent = <-ch
+	assert.Equal(t, devicechange.ID("device-1:3"), changeEvent.ID)
+	changeEvent = <-ch
+	assert.Equal(t, devicechange.ID("device-2:1"), changeEvent.ID)
```
Andrea-Campanella: Same comment on channel, please use switch statement to avoid blocking if something goes wrong. Same applies for all other instances of the test where you listen on channels

pkg/types/change/device/types.go
```diff
@@ -27,24 +27,11 @@ const separator = ":"
...
 }
 
-// GetDeviceVersion returns the device version to which the change is targeted
-func (i ID) GetDeviceVersion() string {
-	if !i.containsVersion() {
-		return ""
-	}
-	return string(i)[strings.Index(string(i), separator)+1 : strings.LastIndex(string(i), separator)]
-}
-
```
Andrea-Campanella: Could we leave these methods in at least until we rip-out the existing channel based mechanism ?

kuujo: This is one of the changes that I added specifically for the stores. It doesn’t affect other code. I just removed the device version from device change IDs after some discussion with @SeanCondon about how versions need to be handled.

Andrea-Campanella: @kuujo I see, my bad. Thanks.

PR #734, opened 2019-10-09
Network change controller

kuujo:
This PR adds a controller for network changes and tests for the logic therein. The controller seems fairly complete, but more tests still need to be added for certain scenarios. I'm opening this PR for early review.

The expected behavior of the controller can be deduced from tests. But I will add more detailed information on the implementation soon...


kuujo:
Okay, I want to elaborate on the codes a bit.

This this architecture is a level triggered architecture, as opposed to edge triggered. What that means is rather than reacting to events as they occur, the controller periodically checks the state of the system. When it checks the state of the system, if there is any discrepancy between the current state and the desired state, it modifies the current state to reach the desired state.

For example, when a network change is found but no device changes are found for the network change, the device changes are created. When a network change in the APPLYING state is found, any device changes that are PENDING are changed to APPLYING. When a network change is APPLYING and all its device changes are complete, the network change is completed. This model depends much less on order, reliability, and timing as no matter the ordering of events, the current state of the system is all that’s relevant.

With that said, the network controller is simply responsible for managing the lifecycle of `NetworkChange`s and the device `Change`s that depend on them. The network controller checks the state of a `NetworkChange` and device `Change`s to progress a network change through the a sequence of states that constitute the asynchronous handling of a change request.

The way the controller reacts to a `NetworkChange` in each state is as follows:
* `PENDING`
   * If necessary, create a device `Change`s for each change in the network change
   * If prior changes for all devices in the change have been completed, set the `NetworkChange` state to `APPLYING`
* `APPLYING`
   * Change device `Change` States from `PENDING` to `APPLYING`
   * If all device `Change` states have `SUCCEEDED`, mark the `NetworkChange` `SUCCEEDED`
   * If a device `Change` state is `FAILED` with a reason other than `UNAVAILABLE`, mark the `NetworkChange` `FAILED`
   * (More on the additional case where a device `Change` state is `FAILED` with the `UNAVAILABLE` reason below)
* `SUCCEEDED`
   * No action taken
* `FAILED`
   * If any child device `Change` is `APPLYING` or `SUCCEEDED`, mark it `APPLYING` with the `deleted` flag set to `true`

These states when applied to device `Change`s can be thought of as switches to indicate to device controllers (potentially running on other nodes) how to handle the changes. In general, the `PENDING` state indicates the change should not be applied, the `APPLYING` state indicates it should be sent to the device, and the `SUCCEEDED` and `FAILED` statuses are indicators of whether the change was successful.

So I think all of this is pretty simple and sensible handling of the changes. But some questions do remain regarding the handling of `UNAVAILABLE` failure reasons. When a device change cannot be applied because a device is unavailable, it must fail the change with the `UNAVAILABLE` reason. In principle, AIUI we intend for changes to unavailable devices to be enqueued until they become available. That’s simple enough for single device changes. But for multi-device changes it becomes a little more complex. Here are a few concerns:
* What happens when a change can only be partially applied because one device was available and the other wasn’t? I would suggest a partial success due to device availability should result in the change reverting to the `PENDING` state - essentially requeueing it until all the devices become available. @SeanCondon suggested the change should fail if only a subset of the devices is available, but that will create a likely very frequent race condition when provisioning multiple devices. There will always be a period of time while booting the devices during which a subset of the devices are online, meaning provisioning changes for multiple devices would fail with this logic. We can mitigate the chances of that happening, but it will never be totally avoidable and the behavior of the mitigating controller would ultimately be the same as if we requeue those failures.
* The race condition mentioned above even exists in single device changes. Even if we only attempt to apply a change once a device is available, it can become unavailable before the change can be applied, in which case we have to decide whether to revert back to `PENDING` or fail the change.
* In my opinion we can benefit from a consistent behavior here: once a change is received, we will eventually attempt to apply it. This is certainly desirable in a distributed system. If the change is hanging due to partial availability of the devices, it can be rolled back
* Of course, we could also accept a configurable policy to define the behavior of the controller when devices are unavailable, but that’s obviously beyond the scope of this PR

There’s one more component of this PR that hasn’t yet been added. That is the use of the `deleted` flag. The `deleted` flag was added as a tombstone for rollbacks, but it’s also needed here to revert changes in multi-device transactions. By the time the controller learns a device change has failed due to any error, other changes may have already succeeded or may succeed by the time the master can be notified of the failure. In that case, setting the `deleted` flag on device changes notifies the device controller that changes need to be undone.

So that’s about all the scenarios I can think of, and I intend to add tests for each of them. But I want to verify that the behavior is correct.

Andrea-Campanella:
> So I think all of this is pretty simple and sensible handling of the changes. But some questions do remain regarding the handling of `UNAVAILABLE` failure reasons. When a device change cannot be applied because a device is unavailable, it must fail the change with the `UNAVAILABLE` reason. In principle, AIUI we intend for changes to unavailable devices to be enqueued until they become available. That’s simple enough for single device changes. But for multi-device changes it becomes a little more complex. Here are a few concerns:
> 
> * What happens when a change can only be partially applied because one device was available and the other wasn’t? I would suggest a partial success due to device availability should result in the change reverting to the `PENDING` state - essentially requeueing it until all the devices become available. Or it can even remain in the `APPLYING` state and retry until it’s successful or cancelled by the network change controller. @SeanCondon suggested the change should fail if only a subset of the devices is available, but that will create a likely very frequent race condition when provisioning multiple devices. There will always be a period of time while booting the devices during which a subset of the devices are online, meaning provisioning changes for multiple devices would fail with this logic. We can mitigate the chances of that happening, but it will never be totally avoidable and the behavior of the mitigating controller would ultimately be the same as if we requeue those failures.
> * The race condition mentioned above even exists in single device changes. Even if we only attempt to apply a change once a device is available, it can become unavailable before the change can be applied, in which case we have to decide whether to revert back to `PENDING` or fail the change.
> * In my opinion we can benefit from a consistent behavior here: once a change is received, we will eventually attempt to apply it. This is certainly desirable in a distributed system. If the change is hanging due to partial availability of the devices, it can be rolled back
> * Of course, we could also accept a configurable policy to define the behavior of the controller when devices are unavailable, but that’s obviously beyond the scope of this PR

@kuujo I like the PENDING approach. I would implement it with an exponential backoff for retry, otherwise we might wasting a lot of computation time for a device that never shows up. 
The other thing to be aware to me is that a PENDING change needs to be the last of the stack for all the devices involved, both available and un-available. This means that if a change for a,b,c is pending because b is un-available and another change arrives for a,c this second change is applied but then a,b,c needs to be rebased on top to allow rollback. @SeanCondon thoughts ?

SeanCondon:
First of all I think we have to take in to account the situation where we are pre-provisioning a device that is not physically present. We still want to be able to push changes down to its device controller and for it to move to `SUCCEEDED`. Technically you might say it's `UNAVAILABLE` (but what we mean is `OFFLINE`). (In the previous implementation the Configuration was so named to highlight that you're not dealing with a device here per se - just a set of configured attributes).

Here's my idea of what `OFFLINE` means:
`OFFLINE` means the device is deliberately disconnected for gNMI purposes (or it was never made `ONLINE`) and that it's OK to push changes down to its device controller. There is no synchronzer instance present for it at this stage. This is a state per device (within the device controller) - not per device change.

If the real device becomes `AVAILABLE` at a later stage (through a Topo event) a synchronizer will be created which has its own state machine. The synchronizer will tell the Device controller to pause (change its state to `SYNCING`) - keeping any future changes in a `PENDING` state.

The Synchronizer will continue the synchronization downwards. If it does not succeed, it tries and tries again with exponential backoff. The effect will be that changes will stay in `PENDING` with no further network changes possible - all because of this one contrary device. Well that's how it's meant to be. Either take the device offline or fix it (maybe it's software needs to be upgraded or it needs to be reset some how). Taking it `OFFLINE` will allow network configuration to continue again.

So I think the Device Controller needs to have an attribute per device with `OFFLINE`, `SYNCING` or `ONLINE` (there might be a case for a 4th `FROZEN` or `CLOSED` to say no more changes are wanted).

I would explain the effect of `ONLINE` like this: device changes should save to the store **AND** to the synchronizer and if either one fails then the change should go back to `PENDING` and this should block the network change. It should propagate error message to the Reason field to allow the user understand what the problem is. To fix it the device might have to be taken `OFFLINE` or fixed by some other means (maybe a gNOI command needs to be run).

For this reason I think the device Change state machine should not have an `UNAVAILABLE` state to it - it should be able to handle any synchronizer errors by marking keeping the change in `PENDING` or using `FAILED`.

@Andrea Regarding your last scenario (and @kuujo mentioned it too) - I think we should keep strict ordering of network changes - regardless of how it might be possible to fit them side by side. e.g. if one change affects a&b and the next change affects c, then we **should not** apply c until a&b are successful. Otherwise all sorts of chaos could be preent and you would hear operators complain that they don't know what state their network is in.

kuujo:
@Andrea-Campanella

Exponential backoff is built in to the controller abstraction but is not always used ... I’ll explain: 

In the controller, there are two reasons a change can fail:
1. The reconciler returns an `error`
2. The reconciler returns `false`

When the reconciler attempts to interact with stores or devices and an error occurs, it should return the error. In that case, the controller will use exponential backoff to retry the reconciliation to avoid overloading the cluster.

When the reconciler returns `false`, that is used to indicate that the change could not be reconciled, not because any error occurred but because the state of the cluster did not allow for it. When the reconciler returns `false` with no error, the controller requeues the change to be processed after pending events. This is used to account for race conditions in the stores. The network controller uses it to wait for prior device changes to be processed. By processing pending events before retrying the change, the controller allows the state of the cluster to advance before attempting to retry the change.

But the controller is also triggered by store events, so there’s a third case where we can avoid backoffs and retries altogether. For example, if the reconciler cannot apply a change due to device availability, it can safely discard the change. If the controller is configured with a watcher on the device store, it will be re-triggered once the device comes back online.

I think there are specific rules that need to be followed for efficiency in the reconciler. Errors should only be returned when they come from the network. Changes should only be requeued when an inconsistency can be resolved by pending events. Changes should be discarded when they’re successful or a failure is due to an event that can take an arbitrary amount of time to resolve, such as a device availability change, in which case it should simply wait for the device store to notify the controller.

As for the `PENDING` states, changes should certainly never be reordered. That would be a severe violation of consistency IMO. Reordering changes to the same paths would mean the controller is deciding which value the user desires. If the user submitted change 1 before change 2 then change 1 should always be applied before change 2.

What I’m suggesting - to leave a change in `PENDING` as long as it takes for all devices to come online - would in the scenario you described simply result in the change for a,c remaining in `PENDING` as well. Without doing this we will end up creating either consistency or reliability problems: consistency problems if we start unilaterally reordering changes, and reliability problems if we fail changes for the types of partial availability states that will inevitably arrive when bringing multiple devices online simultaneously.

kuujo:
I do acknowledge the potential for multi-device transactions to block changes on some devices when one is simply offline. There may be a couple of solutions:
1. Provide an administrative function to `unblock` a device
   * Find all pending changes to unavailable devices that intersect the target device
   * Cancel those changes
2. Support a configurable timeout
   * Record when a change to an available device is blocked by changes to unavailable devices
    * After a timeout, do #1

I think the way to go about it is to allow blocking indefinitely, then implement an administrative function for unblocking devices, then allow it to be done automatically. In that case, we’re effectively still maintaining the same behavior but accounting for race conditions during provisioning and can leave it up to the user which they prefer.

kuujo:
I merged the network controller and device controller together in a single PR since the conversation affects both of them.

I also refactored the proto files and both controllers to implement some of the ideas I’ve mentioned. This is now pretty close to the implementation I proposed.

The change `State` was split into two separate enums:
* `Phase` indicates the type of operation requested for the change:
   * `CHANGE` is a change
   * `ROLLBACK` is a rollback
* `State` indicates the state within a phase
   * `PENDING` indicates the change is pending
   * `RUNNING` indicates the phase (change or rollback) is in progress
   * `COMPLETE` indicates successful completion
   * `FAILED` indicates a failure

This actually isn’t any different than the same four states plus a `deleted` boolean, but it makes more sense in the context of states being relevant to the phase (or `deleted` flag). It also allows for additional phases to be added if necessary.

This seems to have simplified the controller logic quite a bit as well. The network controller logic is a pretty simple tree:
* if phase is `CHANGE`
   * Ensure device changes are created 
   * if state is `PENDING`
      * Ensure device changes are `PENDING`
      * If no prior intersecting changes are `PENDING` or `RUNNING` in the `CHANGE` phase
         * Set network change state to `RUNNING`
   * If state is `RUNNING`
      * Ensure device changes are `RUNNING`
      * If a device change is `FAILED`
         * Ensure all other `PENDING` or `COMPLETE` device changes are in the `ROLLBACK` phase
         * If all changes are either `FAILED` in the `CHANGE` phase or `COMPLETE` in the `ROLLBACK` phase
            * Set the network change back to `PENDING` if unavailable or `FAILED` if an error occurred
      * If all device changes are `COMPLETE`
         * Set network change state to `COMPLETE`

This is basically the entire tree for handling a `CHANGE`, and the `ROLLBACK` phase is fairly similar. Of course, the reconciler is called in a loop (the reconciliation loop in k8s), so once one branch of the tree ends, the reconciler will often be called again and act on the updated state from the prior loop. For example, the reconciler sets the network change to `RUNNING`, then in the next loop updates the device changes to `RUNNING`. It’s simplest to work out the algorithm using simple rules like this, but we’ll probably want to optimize this code later to eliminate redundant updates to changes. That would be premature here.

The device change controller is much simpler. It just applies changes in the `CHANGE` phase and rolls back changes in the `ROLLBACK` phase. But the implementations of apply/rollback remain empty.

I’ll add more tests tonight which should demonstrate all the possible scenarios for the network controller.

kuujo:
@SeanCondon 

I think I agree with what you’re saying, but a couple of points and clarifications...

First, the rationale for the `UNAVAILABLE` reason was to indicate that a change could not be applied because a device was either offline or a connection failed. There’s an important distinction to be made that this cannot rely solely on device states but includes network failures in the synchronizer, timeouts, etc... any failure in which the change failed because we could not get back a response (“available” is a distributed systems term that means a system is not only running but can respond successfully to requests. Remember, we can always believe a device is online when it’s not). In that case, in this PR, the change is reverted back to `PENDING` and thus will be retried. 

Alternatively, the `ERROR` reason indicates the change could not be applied due to an error explicitly returned by the device. I assumed if the device was rejecting a change then the network change should be failed as no matter how many times we retry, the state of the device will remain the same (later changes are blocked) and so the error will remain the same.

If there’s never a case where we propagate an error returned by the device then we can indeed remove the `UNAVAILABLE` reason. In that case, all changes will simply be blocked indefinitely when any failure occurs. Of course, this does seem to imply there’s no failure that cannot be resolved by an administrator. But I suppose in the case where a failure cannot be resolved, the solution is for an administrator to simply roll back the change?

I’m not really opposed to any of that.

But it feels like there’s some confusion about the role of change states in the controllers. The states are used by the controller to determine which changes need to be applied to devices and to notify device controllers when a change can be applied. This is done by simply changing the state of a change. So the change states effectively represent whether or not onos-config has applied each change and what the result of the attempt was:
* When a change is received for an offline device (for provisioning), it remains in the `PENDING` state until the device becomes available
* When a device is online, the network controller can tell the device controller(s) it’s safe to apply the change(s) by changing them to the `APPLYING`/`RUNNING` state 
* When a device change is `APPLYING`/`RUNNING` it can attempt to apply the change through the synchronizer 
* When device change results are set (success/failure), the network change controller (which is the component that has the full view of all devices) determines whether the network change succeeded or failed, in which case it can retry

So, that’s all to say that we don’t want changes for offline devices to progress to `COMPLETED` because it’s the `APPLYING` state that should trigger the synchronizer to apply a change. A change should only be completed once it has been successful, otherwise we have no way to determine which changes need to be applied.

The fundamental guarantee the network and device controllers provide is:
* If change _a_ is stored before change _b_, _a_ will be applied before _b_

In other words, if the device controller sees change _b_ in the `APPLYING` state, it can safely assume change _a_ was already applied. The device controller only needs to concern itself with applying change _b_ and storing the result.

Effectively, a call to the device change reconciler can be thought of as a single event received on the `deviceConfigChan` in the current synchronizer implementation.

PR #735, opened 2019-10-09
Extension for Disconnected devices

### LINKED ISSUE #518: include an extension to report no yet connected device
**Describe the bug**
This issue is based on the experiment which is reported here: #513 . The problem is when there is no device running, the onos-config responds to the requests. For example, in that experiment, we added a device to the device store, configStore, and change Store  but we didn't add a real device using onit add network/simulator. In that scenario, onos-config responds to a get operation with a value which is in the onos-config memory but not coming from the initial configuration in the device. 

**To Reproduce**
See issue #513 

**Expected behavior**
onos-config shouldn't return some values in respond to the requests if there a real device does not exist. It should reply back with a message and inform the user the device doesn't exist.

Andrea-Campanella:
Provides capability for enos-config to report receiving request for no-connected devices.
In get and set the response had the disconnected device extension set.
For subscribe if the device is not connected and un update come the Subscribe response included the extension for device disconnected.

The devices for get and set are contained in the message as a comma separate list. 

Fixes #518

### DIFF
--- a/pkg/northbound/gnmi/get.go
+++ b/pkg/northbound/gnmi/get.go
@@ -76,14 +76,17 @@ func (s *Server) Get(ctx context.Context, req *gnmi.GetRequest) (*gnmi.GetRespon
 
 		notifications = append(notifications, notification)
 	}
-	disconnectedDevices := make([]string, 0)
-	for k := range disconnectedDevicesMap {
-		disconnectedDevices = append(disconnectedDevices, string(k))
-	}
-	disconnectedDeviceString := strings.Join(disconnectedDevices, ", ")
+
 	response := gnmi.GetResponse{
 		Notification: notifications,
-		Extension: []*gnmi_ext.Extension{
+	}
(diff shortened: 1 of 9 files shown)


pkg/northbound/gnmi/get.go
```diff
@@ -69,6 +80,23 @@ func (s *Server) Get(ctx context.Context, req *gnmi.GetRequest) (*gnmi.GetRespon
 	response := gnmi.GetResponse{
 		Notification: notifications,
 	}
+	if len(disconnectedDevicesMap) != 0 {
+		disconnectedDevices := make([]string, 0)
+		for k := range disconnectedDevicesMap {
+			disconnectedDevices = append(disconnectedDevices, string(k))
+		}
+		disconnectedDeviceString := strings.Join(disconnectedDevices, ", ")
```
SeanCondon: I don't know if I'd have the space in `", "` - suggest just `","`

PR #736, opened 2019-10-09
Device change controller

kuujo:
This PR implements the controller for device `Change`s. The PR implements all the configuration and features needed for distributing device changes to device masters, parallelizing changes to different devices, and handling device changes up to southbound requests.

There are three types of device change states the controller recognizes:
* `APPLYING`
   * Attempt to apply the change
   * If the change succeeds, mark it `SUCCEEDED`
   * If the device is unavailable, mark the change `FAILED` with `UNAVAILABLE` reason
   * If the change fails, mark it `FAILED` with the `ERROR` reason
* `APPLYING`+`deleted`
   * Attempt to roll back the change
   * If the rollback succeeds, mark it `SUCCEEDED`
   * If the device is unavailable, mark the change `FAILED` with `UNAVAILABLE` reason
   * If the rollback fails, mark it `FAILED` with the `ERROR` reason
* `PENDING`+`deleted`
   * Attempt to rollback the change
   * If the rollback succeeds, unset its `deleted` flag
   * If the rollback fails, re-enqueue the change for retry

The `APPLYING` and `APPLYING`+`deleted` states indicate a simple change and rollback respectively. 

The `PENDING`+`deleted` state is specifically used to rollback a change _temporarily_. If the change is successfully rolled back, the change remains in the `PENDING` state waiting for the network controller to apply it again.


kuujo:
I actually wonder if there’s another state needed here, or the `deleted` flag needs to be made into a richer enum. It feels to me like we should not be reverting a change back to `PENDING` once it has been moved to `APPLYING`. Changes should only progress from `PENDING` to `APPLYING` to `SUCCEEDED`/`FAILED`. But if we need the device controller to be able to temporarily rollback a change that may later need to be reapplied, it needs to be able to differentiate between a temporary and permanent rollback.

Maybe we should use the state enum to differentiate between `APPLYING` a change and `REVERTING` (which is a decision made by the network controller to maintain consistency in the network) or `ROLLING_BACK` the change (which is a decision made by the user). In that case, the `deleted` flag would remain as an indicator of whether a `SUCCEEDED` change actually applies to the state of the device.

There’s also another pattern for statuses I’ve seen used in k8s. They sometimes use separate status objects to represent different phases of a pod or container lifecycle. For example, the fields in the `Status` object could be:
* `Apply`
* `Revert`
* `Rollback`

Where each field is an object that can be in one of two states:
* Pending
* Complete

So the `Apply` status would look like:
* `Apply`
   * State: `Pending`/`Complete`

Additionally, each of the categories of actions can record timestamps so we can maintain a complete history of apply attempts, reverts (failed applies), and rollbacks.

In this case, a change where `Apply` is `PENDING` the device controller would apply the change. In a change where `Apply` is `COMPLETE` and `Revert` is `PENDING`, the device controller would revert the change. And in a change where `Rollback` is `PENDING` the device controller would roll back the change.

Of course, this could also be implemented as two different enums:
* `Phase` - the change phase
   * `APPLY`
   * `REVERT`
   * `ROLLBACK`
* `State` - the state within the phase
   * `PENDING`
   * `COMPLETE`
* `Result` - the result for a `COMPLETE` phase
   * `SUCCEEDED`
   * `FAILED`

Using distinct messages for each phase allows for more flexibility in the future, but I have to admit the latter enum solution seems fairly elegant.

kuujo:
Closing this to merge with #734 since the discussions are related

PR #740, opened 2019-10-11
Support compaction of change logs via snapshots

kuujo:
This PR adds support for reclaiming space in change stores without losing device state. This feature is implemented using snapshots. 

Three new types of objects are added to facilitate compaction:
* `NetworkSnapshot` is a request to take a snapshot of some or all devices
* `DeviceSnapshot` is a per-device snapshot request created as a child of a `NetworkSnapshot`
* `Snapshot` is the state of a device at a specific point in time

It also adds stores for the above object types, and controllers to handle `NetworkSnapshot` and `DeviceSnapshot` state changes.

A `NetworkSnapshot` represents the lifecycle of a snapshot across multiple devices. `NetworkSnapshot`s are created with a `timestamp` indicating the time before which to discard changes. Device changes are stored with consistent, monotonically increasing wall clock times, so when a `DeviceSnapshot` controller takes a snapshot of a device's state, it simply iterates through the `Change`s that have occurred since the last snapshot up to the requested snapshot time and records the latest state for each path in the device. Once the state up to the requested snapshot time is recorded, it's stored in a `Snapshot` and all `Change`s up to the snapshot time are removed from the `Change` store.

Once all `DeviceSnapshot`s for a `NetworkSnapshot` are complete, the `NetworkSnapshot` controller discards changes to the indicated devices up to the snapshot timestamp.


kuujo:
I'm rethinking how this is implemented. Going to close this PR for now and will reopen after some changes.

PR #741, opened 2019-10-11
Separated out the TypedValue from Value

SeanCondon:
This does 2 unrelated things:
* Marks the old infrastructure as `Deprecated` so it's clear what has to be replaced
* Updates types/change/device/types.proto Value attribute in to its own type like with the legacy structure


pkg/controller/change/device/controller_test.go
```diff
@@ -241,9 +241,12 @@ func newChange(device device.ID) *devicechange.Change {
...
 		Values: []*devicechange.Value{
 			{
-				Path:  "foo",
-				Value: []byte("Hello world!"),
-				Type:  devicechange.ValueType_STRING,
+				Path: "foo",
+				Value: &devicechange.TypedValue{
+					Bytes:    []byte("Hello world!"),
+					Type:     devicechange.ValueType_STRING,
+					TypeOpts: nil,
```
Andrea-Campanella: in some classes you add this nil while in other you leave it empty. Is there a difference ? I believe that leaving it empty is equivalent to putting it to nil

SeanCondon: it just came up when i did "fill all fields" in the IDE - should be left out

kuujo:
lgtm

PR #747, opened 2019-10-14
Support snapshots for changes

kuujo:
This PR adds support for reducing storage overhead by discarding sections of `NetworkChange` and `DeviceChange` history and replacing sequences of changes with snapshots.

To support snapshots, this PR adds three new data types:
* `NetworkSnapshot` represents the lifecycle of a single snapshot job for all devices
* `DeviceSnapshot` represents the lifecycle of a snapshot for a single device
* `Snapshot` stores the set of `Value`s for a single device at a certain point in time

Additionally, two new stores and controllers are added for `NetworkSnapshot` and `DeviceSnapshot`/`Snapshot`.

Snapshots can be configured with a `RetentionOption`s for both `NetworkChange`s and `DeviceChange`s that can specify:
* The period for which to retain changes (e.g. retain the last 1 week of changes)
* The minimum number of changes to retain (e.g. retain the last 100 changes)

The algorithm for taking a snapshot of the device logs works in two phases:
* `MARK` is used to determine which changes can be removed and prepare stores to remove changes
* `DELETE` is used to delete changes that are no longer needed

* `MARK` phase
   * `NetworkSnapshot`
      * The `NetworkSnapshot` controller iterates through the `NetworkChange`s in the store and determines the set of devices that have been changed and the highest change that can be discarded for each device (according to the configured `RetentionOptions`)
      * If the controller encounters a `NetworkChange` that is not `COMPLETE`, it sets the prior index as the maximum change that can be discarded for each device in the `NetworkChange`. This ensures the system does not discard `DeviceChange`s where the `NetworkChange` is not complete (e.g. blocked by another device)
      * For each `NetworkChange` that _is_ `COMPLETE`, the controller marks the change `deleted`
      * The controller creates a `DeviceSnapshot` for each device with changes, indicating the highest index that can be discarded for each device
   * `DeviceSnapshot`
      * The `DeviceSnapshot` controller loads the prior snapshot, if any
      * The `DeviceSnapshot` controller then iterates through changes for the device - starting at the index of the previous snapshot - up to the limit define by either the index provided by the `NetworkSnapshot` controller or the limit defined by the `RetentionOption`s
      * While iterating through `DeviceChange`s, the controller builds a map of the set of paths and values, building on the prior snapshot if available
      * Once the limit has been reached, the `Snapshot` for the device is overwritten and the `MARK` phase is completed
* `DELETE` phase
   * `NetworkSnapshot`
      * The `NetworkSnapshot` controller transitions to the `DELETE` phase once all device changes have `COMPLETE`d the `MARK` phase
      * The controller sets all `DeviceSnapshot`s for the `NetworkSnapshot` to the `DELETE` phase
      * The controller then deletes all `NetworkChange`s that it previously marked `deleted`
      * Once all `DeviceSnapshot`s have `COMPLETE`d the `DELETE` phase, the `NetworkSnapshot` is `COMPLETE`d
   * `DeviceSnapshot`
      * Once in the `DELETE` phase, the `DeviceSnapshot` controller deletes all `DeviceChange`s up to the index of the snapshot
      * Once `DeviceChange`s are deleted, the `DeviceSnapshot` is `COMPLETE`d

There are a few guarantees and limitations of this algorithm:
* All changes that are pending/incomplete are retained
* A `DeviceChange` will never be deleted prior to its `NetworkChange` being deleted
* The history of `DeviceChange`s for a device will always be sequential: represented by a snapshot plus a sequence of changes following the snapshot
* But the `DeviceChange` store may contain changes for which no `NetworkChange` exists. This is to allow longer per-device histories to be retained. Retaining all `NetworkChange`s while a `DeviceChange` exists would also mean maintaining all `DeviceChange`s for those `NetworkChange`s, and because snapshotting requires a full sequential history of changes per-device, retention options on multi-device changes could severely limit the number of changes that can be removed, e.g. if one device almost never changes and a minimum history is set it could block devices with infinite histories.
* Because the `NetworkChange` store contains changes that affect all devices, it does not maintain a complete history for the entire network going back n days or m changes. If it did, a single change on one device would prevent snapshotting of all devices.


SeanCondon:
Hi Jordan - very good description. The only question I have is re:
> But the DeviceChange store may contain changes for which no NetworkChange exists. This is to allow longer per-device histories to be retained.

This is the only case in the complete implementation that I am aware of where we can end up with a `DeviceChange` with no corresponding `NetworkChange`. It's like these `DeviceChange`s have become orphaned (with their Network change reference being left blank). It wasn't clear to me from reading the text how this happens.

I wonder if there's the case of having a special type of 'NetworkChange' (one per device maybe) to become foster parent to these orphaned 'DeviceChange's. In that way the navigability from `NetworkChange` to `DeviceChange` might be maintained.

kuujo:
@SeanCondon I thought about this a lot for the past few days and tried to come up with a way to avoid this. I guess there are three options:
* Preserve all `NetworkChange`s where any `DeviceChange` cannot be removed, and preserve all `DeviceChange`s for any `NetworkChange` that cannot be removed. 
* Preserve all `NetworkChange`s where any `DeviceChange` cannot be removed, but allow `DeviceChange`s that can be removed to be removed (some `NetworkChange`s may point to `DeviceChange`s that have been removed)
* Allow `DeviceChange`s to be preserved even if their `NetworkChange` has been removed

This PR takes the last approach. The second approach is also feasible. But the first approach can create a lot of problems. Because `NetworkChange`s can change multiple devices, and because we need to preserve a complete history of changes since the snapshot, if we allow _n_ changes to be preserved then the device with the oldest history can block any other device touched by the same `NetworkChange`s. 

So, this is a conscious decision to allow `DeviceChange`s to be orphaned. The rationale is this still allows the history for a single device to be retained and viewed if desired, but it avoids severely limiting the ability to reduce histories in some cases.

The way this is implemented is the `NetworkSnapshot` takes both a `NetworkRetention` policy and a `DeviceRetention` policy. So, e.g. you can retain the last week of `NetworkChange`s and the last day of each set of `DeviceChange`s or a minimum of 100 device changes. `DeviceSnapshot` controllers can only snapshot `DeviceChange`s created from `NetworkChange`s that will be removed by the `NetworkSnapshot` controller.

Alternatively, it seems the second option could retain the information you’re seeking because:
* `DeviceChange`s can only be removed for `NetworkChange`s that have already been successful, thus...
* The state of each `DeviceChange` can be  deduced fairly reliably from the state of the `NetworkChange` since it still contains references to devices, and the `NetworkChange` state depends on its children’s states

It would also be fairly easy to implement. But either way, I don’t think we can practically keep all the `NetworkChange`s and `DeviceChange`s because devices’ changes can intersect, thus practically limiting the amount of data we can remove to the oldest history deemed removable.

There _is_ one way we could do it: one option for retention is retaining a minimum number of changes. The rationale for this is just to avoid completely discarding the recent history of a device that hasn’t changed in a while. We could retain those changes only for informational purposes, and only use guarantee we can retain complete `NetworkChange`s and `DeviceChange`s within the retention time frame. So, for example, if you take a snapshot configured to retain the last week of changes and at least 100 changes for each device, changes within the 100 change window but older than a week old will be snapshotted and thus cannot be rolled back (we no longer have the history of `NetworkChange`s for the `DeviceChange`s), but can still be viewed individually.

Maybe #2 is best?

kuujo:
Actually, I think there’s a way to retain the `NetworkChange`s and all `DeviceChange`s for them. It will slow down the snapshotting process and make it a bit more complicated, but it will be throttled anyways. Performance is not a concern. But there will be a caveat:
* Each device must still have a complete history going back to the last snapshot
• But prior to the last snapshot, changes for a device may remain but there can be holes in the history. This is inevitable because of intersecting changes. It’s impossible ensure all the device histories are complete and retain all device changes for each network change.

Does this sound desirable?

On a side note, we will need to add a new primitive to allow for efficiently iterating a sorted data structure with immutable offsets (a log) and support for deletes. But that should be relatively simple to do, and it’s already needed for the `NetworkChange` store (which currently uses a counter to create an ordered map).

SeanCondon:
I like the sound of that later option. But I have lost track of why
> changes for a device may remain but there can be holes in the history

I can't really think of a reason why the device would have a history greater/deeper/older than the the `NetworkChange` history. Can't we just use the snapshot of the `DeviceChange` as the consolidated list for that device. Is it that you think users might want to have different policies for different devices or device types? I think that level of control is probably not necessary - fine to hold it in a log, but probably would not expect to go back there in time.

Forgive me if I've got the wrong end of the stick here :-)

kuujo:
There doesn’t have to be holes in the history. We can certainly preserve entire histories, but it could be extremely costly. The problem is, histories across devices cannot be aligned since changes can span multiple devices. We can’t keep only e.g. the last 100 changes per device and also preserve the network changes that created them and still guarantee we can reclaim space. Keeping the last n changes for each device is simple, but if keeping a complete history for each device back to some point also implies histories for devices affected by the same changes must also be retained then there’s a potential we have to retain a huge amount of data. Take the following example:
* Devices _a_ and _b_ were changed a week ago. This was the last change to device _a_
* But device  _b_ was changed a million times since the a,b change

In this case, if we allow a snapshot to be configured to retain the last _n_ changes per device, since the a,b change was the last change to _a_, it would be retained. If we also require that device changes not be orphaned, we also have to preserve the network change that created it a week ago. That’s not too bad yet. But if we also require that every device maintain a complete history back to any change it has retained, device _b_ would have to retain the last million + 1 changes, and our ability to reclaim space for device _b_ becomes dependent on the changes being made to device _a_.

So, that’s why I say the only way we can retain at least _n_ changes per device is by allowing gaps in the histories of each device, at least for changes older than the snapshot. But IMO this isn’t that big of an issue. The device will still retain the complete history of changes it was instructed to retain, and changes older than that are only retained for the purpose of ensuring `NetworkChange`s still in the store are complete (all the `DeviceChange`s they created are retained. So if some `DeviceChange` is retained by some device, the `NetworkChange` that created it is retained, and all `DeviceChange`s created by that `NetworkChange` are retained (which leads to holes in device histories).

There is a way around this, but I’m not sure how desirable it is. This problem only exists if one option is to retain at least _n_ changes per device. If the only option for compaction is to discard changes more than x days old, then you can very reliably ensure you end up with a consistent history of changes for the network and per-device. But you will also end up with _no_ history for any device that hasn’t been changed in the last x days. The minimum history option is to guard against this, and the only case where it’s applicable (and can create incomplete histories in other devices modified by the same set of changes) is when the history would otherwise be minimal. So basically, we’re keeping the last x days of history for each device, but in cases where that constraint would wipe out the history for a device we’re retaining just enough information to be useful.

SeanCondon:
Jordan - thanks for the illustrative explanation

I think the way around it is acceptable as you describe it. I think being able to rollback by time is more important than to be able to rollback any absolute number of steps. I think it is a price worth paying to have the conceptual logic that every DeviceChange has a corresponding Network Change. 

I think regarding your device **a** above - I think it is acceptable that the history that was rolled up in snapshotting this would be preserved in a log **only** and not as a device history in the `DeviceChange` store because the fact that it's **logged** would be enough to recreate some arbitrary or all histories either manually or by code at a later date if they wanted to trace back. We could have a special log format for this if it helped.

This I think would help with the auditing of the system and would scare me a lot less than the possibility of having orphaned DeviceChanges and holes in the Device history

kuujo:
Fair enough. But in that case we need to think about a log, and that creates some problems of its own. For the log to be useful, the `NetworkChange`s _must_ be derived from the log. So, this goes back to the very first implementation I created which wrote first to an append-only log, then to the stores from the log.

For a log to be reliable for reproducing state changes it has to be consistent, which means it has to be done by a consistent system. We can’t simply write application logs and derive a total order from them because the logs can be written by different nodes at different times. The only way to derive a consistent order, in that case, would be total clock synchronization which is impossible. Without that, disparate application logs would not precisely represent the order in which onos-config applied changes to the network.

We’d also want the `NetworkChange` queue to be derived from the order determined by logs to ensure the changes can be precisely reproduced from the log, which requires a centralized logging solution as well. 

I’m not opposed to centralized logging. It was in my first design. But it doesn’t totally solve this problem. Even systems like Kafka don’t/can’t keep logs forever. The aging of logs out of any system will result in erratic per-device change logs, but we can ensure useful information for every device is still retained.

> Orphaned DeviceChanges and holes in the Device history

The alternative that I proposed avoids orphaned DeviceChanges, which is what creates holes in the Device histories. When a DeviceChange is retained by some device, causing the NetworkChange to be retained, if that NetworkChange also must retain all its DeviceChanges then that is what can cause holes in another device’s history. For example:
* Device _a_ retains the last 100 DeviceChanges
* 1 is the most recent change, 100 is the oldest change
* The 100th change is from _NetworkChange 1_ which also changed device _b_
* Device _b_ also retains 100 changes, but its change from _NetworkChange 1_ happened 200 changes ago
* Thus, device _b_ ends up with the last 100 changes + the _a,b_ change at 200

I don’t really think this is all that problematic. The histories of devices _a_ and _b_ should still be thought of as only the complete history of the last 100 changes and a snapshot, and the history should only be shown as those changes. Changes beyond 100 are only retained for completion of the NetworkChanges that reference them, and NetworkChanges only for completion of the DeviceChanges they created. Such NetworkChanges and DeviceChanges cannot be used to determine history and should be treated as if they have already been deleted (and should be marked as such) for all purposes aside from determining lineage for the DeviceChanges that prevent them from being removed from stores entirely.

In other words:
* If you look at the history of the network you should see the last _n_ count or _d_ days of changes plus all pending changes
* If you look at the history of a single device you should only see the last _n_ count or _d_ days of changes, including all pending changes, no gaps
* If you look at the lineage for any network change, it is always retained at both the network level and per device within a complete history of each, and a NetworkChange that is still visible can always be rolled back 
* If you want to see the lineage for a specific device change, the lineage is always retained at both the network level and device level but not always within a complete history of changes. Not all device changes can always be rolled back, e.g. if the change aged out of the NetworkChange log. But you can still see however much history for a device you wish to retain.

So, from a user’s perspective we should still be retaining full histories of changes. The gaps in the histories are merely the result of retaining additional information for device histories that are not aligned after compaction.

kuujo:
@SeanCondon I propose we go with your last suggestion for now and simplify the algorithm in this PR by only allowing snapshots to specify a `Duration` for which to retain all changes. It feels like that should suffice for now, and we can open a separate issue to track future work on snapshots and describe the proposals there.

Only supporting time-based retention can indeed result in device histories being lost in some cases, but for our immediate purposes it should be more than sufficient. We can think about that problem and the centralized logging problems I mentioned after we finish integrating and testing the stores.

Andrea-Campanella:
@kuujo @SeanCondon thanks for the discussion, was following along. I think that the solution proposed by @kuujo in the latest comment is the right way forward and would enable us to implement the feature without overthinking/over specifying it too much, leaving the door open for future enhancements.

SeanCondon:
Thanks @kuujo - I think this is the right approach - lets see what way specific user requirements and real world experience bring us in the future. 

Do you have more commits to make to bring it to that point? I'm OK to approve this whenever you're ready

kuujo:
Yep I need to clean this up a bit then it should be g2g

pkg/controller/snapshot/device/controller.go
```diff
@@ -0,0 +1,210 @@
...
+	snapstore "github.com/onosproject/onos-config/pkg/store/snapshot/device"
+	"github.com/onosproject/onos-config/pkg/types"
+	changetype "github.com/onosproject/onos-config/pkg/types/change"
+	devicechangetype "github.com/onosproject/onos-config/pkg/types/change/device"
+	snaptype "github.com/onosproject/onos-config/pkg/types/snapshot"
+	devicesnaptype "github.com/onosproject/onos-config/pkg/types/snapshot/device"
+	deviceservice "github.com/onosproject/onos-topo/pkg/northbound/device"
+)
+
+// NewController returns a new network controller
```
kuujo: Comment is incorrect

pkg/controller/snapshot/device/controller.go
```diff
@@ -0,0 +1,210 @@
...
+}
+
+// Resolver is a DeviceResolver that resolves device IDs from device snapshot IDs
+type Resolver struct {
+	snapshots snapstore.Store
+}
+
+// Resolve resolves a device ID from a device snapshot ID
+func (r *Resolver) Resolve(id types.ID) (deviceservice.ID, error) {
+	snapshot, err := r.snapshots.Get(devicesnaptype.ID(id))
```
kuujo: The device ID can be resolved from the snapshot ID without accessing the store

pkg/controller/snapshot/device/controller.go
```diff
@@ -0,0 +1,210 @@
...
+
+	// Complete the snapshot MARK phase
+	deviceSnapshot.Status.State = snaptype.State_COMPLETE
+	if err := r.snapshots.Update(deviceSnapshot); err != nil {
+		return false, err
+	}
+	return true, nil
+}
+
+// reconcileMark reconciles a snapshot in the DELETE phase
```
kuujo: Comment is incorrect

pkg/controller/snapshot/device/watcher.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+// Start starts the device snapshot watcher
+func (w *Watcher) Start(ch chan<- types.ID) error {
+	w.mu.Lock()
+	if w.ch != nil {
+		return nil
+	}
+
+	configCh := make(chan *devicesnaptype.DeviceSnapshot, queueSize)
```
kuujo: Wrong variable name

pkg/controller/snapshot/network/controller.go
```diff
@@ -0,0 +1,375 @@
...
+	snapshot.Status.State = snaptypes.State_RUNNING
+	if err := r.networkSnapshots.Update(snapshot); err != nil {
+		return false, err
+	}
+	return true, nil
+}
+
+// canApplySnapshot returns a bool indicating whether the change can be applied
+func (r *Reconciler) canApplySnapshot(snapshot *networksnaptypes.NetworkSnapshot) (bool, error) {
+	for index := networksnaptypes.Index(0); index < snapshot.Index; index++ {
```
kuujo: Sequential ordering in the `NetworkSnapshot` store is guaranteed, so we should be able to simply check whether the prior index is complete.

pkg/controller/snapshot/network/watcher.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+// Start starts the network snapshot watcher
+func (w *Watcher) Start(ch chan<- types.ID) error {
+	w.mu.Lock()
+	if w.ch != nil {
+		return nil
+	}
+
+	configCh := make(chan *networksnaptype.NetworkSnapshot, queueSize)
```
kuujo: Incorrect variable name

pkg/controller/snapshot/network/watcher.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+// Start starts the device snapshot watcher
+func (w *DeviceWatcher) Start(ch chan<- types.ID) error {
+	w.mu.Lock()
+	if w.ch != nil {
+		return nil
+	}
+
+	configCh := make(chan *devicesnaptype.DeviceSnapshot, queueSize)
```
kuujo: Incorrect variable name

kuujo:
This PR is ready. I just need to clean up some of the code.

So the algorithm used is this:
* When a `NetworkSnapshot` is created, it can be configured with a duration for which to retain changes
* The `NetworkSnapshot` controller iterates through `NetworkChange`s and marks changes for deletion
* For each device, the controller determines the highest index that can be snapshotted according to the snapshot configuration
* If a `NetworkChange` is inside the retention window, the snapshots for each device in the change are limited to the prior change
* If a `NetworkChange` is not complete, the snapshots for each device in the change are limited to the prior index
* A `DeviceSnapshot` is created for each device indicating the index up to which to take a snapshot
* The `DeviceSnapshot` controller initializes a snapshot with the state of any prior snapshot, then replays changes up to the provided index to build and store the new snapshot
* Once all device snapshots are complete, the `NetworkSnapshot` controller tells `DeviceSnapshot` controllers to delete changes
* Once all `DeviceChange`s have been deleted, the `NetworkChange`s that created them are deleted

This algorithm guarantees (even if a node crashes while taking a snapshot):
* A complete history of network changes and changes for each device is retained according to the configured retention period
* A change will never be deleted before it is `COMPLETE`
* A change will never be deleted before a snapshot has been taken of it
* If a `NetworkChange` exists and has not been marked `deleted`, all its `DeviceChange`s exist

Once I resolve the issues on my own review this can be merged.

PR #755, opened 2019-10-15
Removed the deprecated types ConfigValue, ChangeValue and ValueCollection

SeanCondon:
The main points are:
* Modified `pkg/types/change/device/types.proto`
   * Created new `PathValue` - like `Value` without removed flag e.g. for OpState values
   * Does not affect the `DeviceChange`
   * Renamed the existing `Value` to `ChangeValue`
* Removed `pkg/store/change/configvalue.go`
* Removed `pkg/store/change/changevalue.go`
* Created `pkg/types/change/device/value.go` to hold methods of old `ChangeValue`

All other changes are replacing old types with new


pkg/store/change/change_test.go
```diff
@@ -102,11 +100,11 @@ func Test_ConfigValue(t *testing.T) {
 	path := "/cont1a/cont2a/leaf2a"
 	value := 13
 
-	configValue2a, _ := NewChangeValue(path, types.NewTypedValueUint64(uint(value)), false)
+	configValue2a, _ := types.NewChangeValue(path, types.NewTypedValueUint64(uint(value)), false)
 
 	assert.Equal(t, configValue2a.Path, path)
 
-	assert.Equal(t, configValue2a.String(), "/cont1a/cont2a/leaf2a [13 0 0 0 0 0 0 0] false")
+	assert.Equal(t, configValue2a.String(), `path:"/cont1a/cont2a/leaf2a" value:<bytes:"\r\000\000\000\000\000\000\000" type:UINT > `)
```
Andrea-Campanella: I think it's better if we do not compare strings directly but go in depth and compare the path and the value as objects

pkg/store/change/change_test.go
```diff
@@ -141,50 +139,41 @@ func Test_changecreation(t *testing.T) {
...
 
 func Test_changeValueString(t *testing.T) {
-	cv1, _ := NewChangeValue(Test1Cont1ACont2ALeaf2A, types.NewTypedValueUint64(123), false)
-	var buf [8]byte
-	binary.LittleEndian.PutUint64(buf[:], 123)
-	byteArr123 := fmt.Sprintf("%v", buf)
+	cv1, _ := types.NewChangeValue(Test1Cont1ACont2ALeaf2A, types.NewTypedValueUint64(123), false)
 
-	assert.Equal(t, cv1.String(), Test1Cont1ACont2ALeaf2A+" "+byteArr123+" false",
+	assert.Equal(t, cv1.String(), `path:"/cont1a/cont2a/leaf2a" value:<bytes:"{\000\000\000\000\000\000\000" type:UINT > `,
```
Andrea-Campanella: same

pkg/types/change/device/types.proto
```diff
@@ -97,14 +97,25 @@ message TypedValue {
...
+    TypedValue value = 2 [json_name="Value", (gogoproto.jsontag) = "Value,omitempty"];
 
     // 'removed' indicates whether this is a delete
     bool removed = 3 [(gogoproto.jsontag) = "Remove,omitempty"];
 }
+
+// PathValue is an individual Path/Value combination - it is like Value above
+// without the removed flag - it is not used in the DeviceChange store
+// Instead it is useful for handling OpState where `removed` is not relevant
+message PathValue {
```
Andrea-Campanella: Instead of PathValue with is confusing, I think `StateValue` is actually what you mean here. This signifies a Value that is State, meaning Read only no r/w

pkg/store/change/change_test.go
SeanCondon: this is the test of the String() function

pkg/store/change/change_test.go
SeanCondon: done

pkg/types/change/device/types.proto
SeanCondon: As agreed left `PathValue` as it as it is for more than just State

Instead moved `Value` in types to be `ChangeValue`

PR #760, opened 2019-10-16
Replaced types in NBI diags with new Types from stores

SeanCondon:
The main files to look at here are
- diags.proto and
- admin.proto

Instead of using the old `ChangeValue` from `admin.proto` I'm reusing the `ChangeValue` and `PathValue` from types

This helps simplify the other files as the types can be directly passed across


SeanCondon:
The main files to look at here are 
* diags.proto and 
* admin.proto

Instead of using the old `ChangeValue` from `admin.proto` I'm reusing the `ChangeValue` and `PathValue` from `types`

This helps simplify the other files as the types can be directly passed across

PR #762, opened 2019-10-16
Extending NBI diags with RO access to NetworkChanges and DeviceChanges stores

SeanCondon:
The main changes are in diags.proto

Added in 4 new RPCs
* Get of Network Changes
* Get of Device Changes
* List of Network Changes
* List of Device Changes

Not implemented yet - just wanted to give a heads up on the Protos


pkg/northbound/admin/admin.proto
```diff
@@ -16,8 +16,8 @@ limitations under the License.
 
 syntax = "proto3";
 
-// Package admin defines the administrative and diagnostic gRPC interfaces.
-package proto;
+// Package admin defines the administrative gRPC interfaces.
+package onos.config.admin;
```
Andrea-Campanella: is this what we want ? also please do update the comment

SeanCondon: I was copying the style from Jordan's types.proto

PR #767, opened 2019-10-16
Set SB plumbing, Get and Validate in parallel with Atomix

Andrea-Campanella:
This patch provides in parallel operations based on the new atomic store for set in SB, Get and Validate changes:
- in the change/device/controller.go we are now converting the change to a gNMI setRequest.
- Get in the SB works against the atomix based stores
- Validate validates the changes based on the atomic stored elements.
- store-api_test.go are only partially ported.

### DIFF
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -35,7 +35,7 @@ import (
 	"github.com/onosproject/onos-config/pkg/store/mastership"
 	devicesnap "github.com/onosproject/onos-config/pkg/store/snapshot/device"
 	networksnap "github.com/onosproject/onos-config/pkg/store/snapshot/network"
-	types "github.com/onosproject/onos-config/pkg/types/change/device"
+	devicechangetypes "github.com/onosproject/onos-config/pkg/types/change/device"
 	devicepb "github.com/onosproject/onos-topo/pkg/northbound/device"
 	"google.golang.org/grpc"
 	log "k8s.io/klog"
@@ -67,7 +67,7 @@ type Manager struct {
 	OperationalStateChannel   chan events.OperationalStateEvent
 	SouthboundErrorChan       chan events.DeviceResponse
 	Dispatcher                *dispatcher.Dispatcher
-	OperationalStateCache     map[devicepb.ID]types.TypedValueMap
+	OperationalStateCache     map[devicepb.ID]devicechangetypes.TypedValueMap
 }
 
 // NewManager initializes the network config manager subsystem.
@@ -105,7 +105,7 @@ func NewManager(configStore *store.ConfigurationStore, leadershipStore leadershi
 		OperationalStateChannel:   make(chan events.OperationalStateEvent, 10),
 		SouthboundErrorChan:       make(chan events.DeviceResponse, 10),
 		Dispatcher:                dispatcher.NewDispatcher(),
-		OperationalStateCache:     make(map[devicepb.ID]types.TypedValueMap),
+		OperationalStateCache:     make(map[devicepb.ID]devicechangetypes.TypedValueMap),
 	}
 
 	changeIds := make([]string, 0)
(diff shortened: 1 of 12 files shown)


pkg/store/change/device/utils.go
```diff
@@ -0,0 +1,98 @@
...
+)
+
+// ExtractFullConfig retrieves the full consolidated config for a Configuration
+// This gets the change up to and including the latest
+// Use "nBack" to specify a number of changes back to go
+// If there are not as many changes in the history as nBack nothing is returned
+func ExtractFullConfig(deviceID devicetopo.ID, newChange *device.Change, changeStore Store,
+	nBack int) ([]*device.PathValue, error) {
+
+	// Have to use a slice to have a consistent output order
```
Andrea-Campanella: change comment

pkg/manager/manager.go
```diff
@@ -348,6 +348,41 @@ func (m *Manager) computeChange(updates types.TypedValueMap,
...
+		changeValue, err := devicechangetypes.NewChangeValue(path, value, false)
+		if err != nil {
+			log.Warningf("Error creating value for %s %v", path, err)
+			continue
+		}
+		newChanges = append(newChanges, changeValue)
+	}
+	//deletes
+	for _, path := range deletes {
+		changeValue, _ := devicechangetypes.NewChangeValue(path, devicechangetypes.NewTypedValueEmpty(), true)
```
SeanCondon: I suggest calling it `deleteValue`

pkg/manager/manager.go
```diff
@@ -348,6 +348,41 @@ func (m *Manager) computeChange(updates types.TypedValueMap,
...
+// ComputeNewDeviceChange computes a given device change the given updates and deletes, according to the path
+// on the configuration for the specified target
+func (m *Manager) ComputeNewDeviceChange(deviceName string, version string,
+	deviceType string, updates devicechangetypes.TypedValueMap,
+	deletes []string, description string) (*devicechangetypes.Change, error) {
+
+	var newChanges = make([]*devicechangetypes.ChangeValue, 0)
+	//updates
+	for path, value := range updates {
+		changeValue, err := devicechangetypes.NewChangeValue(path, value, false)
```
SeanCondon: I suggest calling it `updateValue`

pkg/manager/manager.go
```diff
@@ -348,6 +348,41 @@ func (m *Manager) computeChange(updates types.TypedValueMap,
...
+			continue
+		}
+		newChanges = append(newChanges, changeValue)
+	}
+	//deletes
+	for _, path := range deletes {
+		changeValue, _ := devicechangetypes.NewChangeValue(path, devicechangetypes.NewTypedValueEmpty(), true)
+		newChanges = append(newChanges, changeValue)
+	}
+	//description := fmt.Sprintf("Originally created as part of %s", description)
```
SeanCondon: Probably time to dump the `description`. It will now always be traceable back to the network change, which will have the description

pkg/store/change/device/utils.go
```diff
@@ -0,0 +1,98 @@
...
+	devicetopo "github.com/onosproject/onos-topo/pkg/northbound/device"
+	"sort"
+	"strings"
+)
+
+// ExtractFullConfig retrieves the full consolidated config for a Configuration
+// This gets the change up to and including the latest
+// Use "nBack" to specify a number of changes back to go
+// If there are not as many changes in the history as nBack nothing is returned
+func ExtractFullConfig(deviceID devicetopo.ID, newChange *device.Change, changeStore Store,
```
SeanCondon: Would need some thorough unit tests. 

Also I expected this to be more of a static function that takes a set of deviceChanges - did not expect to see the call to List() in it. Also the order of them is *very* important

pkg/manager/manager.go
Andrea-Campanella: done

pkg/store/change/device/utils.go
Andrea-Campanella: @ray-milkey store-api_test.go were the tests I mentioned that need adaptation to this new method under `utils.go`

PR #768, opened 2019-10-16
Continuing of the extending on the NBI

SeanCondon:
> I think I merged the previous PR by accident - this might have broken the integration tests

The think to look at here is the diags.proto and the new `ChangeService` with just 2 methods:

- `ListNetworkChanges` and 
- `ListDeviceChanges`

I've also added 2 CLI commands

- `onos config watch device-changes <device-change-id wildcard>` and
- `onos config watch network-changes <network-change-id wildcard>`

I haven't finished wiring these up to the actual stores yet.

There are test cases for the CLI commands


pkg/cli/root.go
```diff
@@ -26,7 +26,7 @@ func GetCommand() *cobra.Command {
 
 	cmd.AddCommand(getGetCommand())
 	cmd.AddCommand(getAddCommand())
-	//cmd.AddCommand(getRemoveCommand())
```
Andrea-Campanella: why is this removed ?

pkg/northbound/diags/diags.proto
```diff
@@ -123,6 +103,10 @@ message ListDeviceChangeRequest {
     // subscribe indicates whether to subscribe to events (e.g. ADD, UPDATE, and REMOVE) that occur
     // after all devices have been streamed to the client
     bool subscribe = 1;
+
+    // option to specify a specific device change - if blank or '*' then select all
+    // Can support `*` (match many chars) or '?' (match one char) as wildcard
+    string changeid = 2;
```
kuujo: FYI you can use gogo to generate a Go struct field that's properly cased:
```go
string change_id = 2 [(gogoproto.customname) = "ChangeID"];
```
and also properly typed
```go
string change_id = 2 [(gogoproto.customname) = "ChangeID", casttype = "github.com/onosproject/onos-config/pkg/types/change/network.ID"];
```

pkg/cli/root.go
SeanCondon: it had been commented out because you could not remove (unload) a plugin. I was removing the commented text here

pkg/northbound/diags/diags.proto
SeanCondon: Thanks  - will do that

pkg/cli/devicechanges.go
```diff
@@ -0,0 +1,76 @@
+// Copyright 2019-present Open Networking Foundation.
```
Andrea-Campanella: maybe `watchdevicechanges.go` is a better name ?

pkg/cli/networkchanges.go
```diff
@@ -0,0 +1,97 @@
+// Copyright 2019-present Open Networking Foundation.
```
Andrea-Campanella: same maybe `watchnetworkchanges.go`

PR #769, opened 2019-10-16
Start controllers on manager Run

kuujo:
Start change/snapshot controllers when Manager is started


pkg/manager/manager.go
```diff
@@ -238,6 +264,16 @@ func validateConfiguration(configObj store.Configuration, changeStore map[string
 // Run starts a synchronizer based on the devices and the northbound services.
 func (m *Manager) Run() {
 	log.Info("Starting Manager")
+
+	// Start the NetworkChange controller
+	_ = m.networkChangeController.Start()
```
SeanCondon: not possible for it to fail?

PR #772, opened 2019-10-17
Adding deviceType to Change

Andrea-Campanella:
This PR add back a device_type in the Proto definition of a Change, such type is useful to make sure changes are used according to the model they are intended to and the device supports.


kuujo:
@Andrea-Campanella shouldn't the device type come from the device service? If we need the device type to validate changes for devices that don't exist yet, what do we do when the device is added with a different type?

Andrea-Campanella:
> @Andrea-Campanella shouldn't the device type come from the device service? If we need the device type to validate changes for devices that don't exist yet, what do we do when the device is added with a different type? If we fail the changes when the device is added with a different type, shouldn't we just be validating the changes when the device info is available anyways?

@kuujo If the device is added with a different type then we should not apply that config to that device. but we do not "fail" the config, at a later point in time that device can change type. The validation happens not against a device but against a set of Yang models and existing config (if any) for that target. What this does is enable us to validate config at any given time independently of the device. makes sense ?

kuujo:
What's the difference, then, between the device type and device version? Both seem to constrain the device configuration in the same way. What is the value of separating them?

kuujo:
So, then the device type and version both need to match the change type and version to be applied to a device?

I think this has implications for snapshots. Snapshots need to not only be taken for a device + version, but device + type + version?

What happens if a change is validated against the type + version it specifies, and a device is added with that type + version, but the actual device doesn’t match the model? In other words, the model doesn’t match the device? Can a model be changed after changes have been validated against it? Does an administrator have to roll back each of the changes first?

kuujo:
I think this PR is fine still, but there are some problems with the way we say we handle the device type/version extensions. I will open an issue separately to address them.

PR #777, opened 2019-10-17
Check previous change when determining change order in NetworkChange controller

kuujo:
This PR fixes an inefficiency in the `NetworkChange` controller. Currently, the controller checks whether a change can be applied (in order) by iterating through changes. This change uses the `GetPrev` method on the `NetworkChange` store to check only the previous change(s) until it determines whether the change can be applied.

### DIFF
--- a/pkg/controller/change/network/controller.go
+++ b/pkg/controller/change/network/controller.go
@@ -51,8 +51,6 @@ func NewController(leadership leadershipstore.Store, deviceStore devicestore.Sto
 type Reconciler struct {
 	networkChanges networkchangestore.Store
 	deviceChanges  devicechangestore.Store
-	// changeIndex is the index of the highest sequential network change applied
-	changeIndex networktypes.Index
 }
 
 // Reconcile reconciles the state of a network configuration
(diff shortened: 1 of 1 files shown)


pkg/controller/change/network/controller.go
```diff
@@ -155,23 +153,30 @@ func (r *Reconciler) createDeviceChanges(networkChange *networktypes.NetworkChan
...
-			return false, err
-		} else if priorChange != nil {
-			if priorChange.Status.State == changetypes.State_PENDING || priorChange.Status.State == changetypes.State_RUNNING {
-				if isIntersectingChange(change, priorChange) {
+	prevChange, err := r.networkChanges.GetPrev(change.Index)
+	if err != nil {
+		return false, err
+	}
+
+	for prevChange != nil {
```
SeanCondon: is there a risk this could whizz around at 100% CPU until it's `COMPLETE`? Could it do with a sleep or be event driven?

kuujo: There's no risk of that. At the end of the loop, the previous index is read:
```go
prevChange, err = r.networkChanges.GetPrev(prevChange.Index)
if err != nil {
	return false, err
}
```
Basically this is just iterating through the changes in reverse until the first intersecting change is found. So normally it will be a few reads.

PR #781, opened 2019-10-18
Change snapshot to have a PathValue rather than ChangeValue

SeanCondon:
`PathValue` is like the `ChangeValue` without the `Removed` boolean.

Snapshots never use the `Removed` attribute

This also tweaks how snapshot controller deals with `removed` values - it now not only deletes the deleted path, but all of its children too. The children should not exist if the parent has been removed

Also extended test cases to execute this concept 
4 paths created

- /foo/msg
- /foo/meaning
- /bar/msg
- /bar/meaning

Then **/bar** is removed
Only 

- /foo/msg
- /foo/meaning

remain


kuujo:
Looks fine to me

PR #789, opened 2019-10-21
Set Response based on new atomix stores

Andrea-Campanella:
This PR introduces a way to create the SetResponse for a setRequest on the GNMI NB based on the feedback from the Atomix stores.

### DIFF
--- a/pkg/northbound/gnmi/set.go
+++ b/pkg/northbound/gnmi/set.go
@@ -24,7 +24,9 @@ import (
 	"github.com/onosproject/onos-config/pkg/modelregistry/jsonvalues"
 	"github.com/onosproject/onos-config/pkg/store"
 	"github.com/onosproject/onos-config/pkg/store/change"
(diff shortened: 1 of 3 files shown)


pkg/northbound/gnmi/set.go
```diff
@@ -527,6 +550,52 @@ func doRollback(changes mapNetworkChanges, mgr *manager.Manager, target string,
 		errResp.Error(), rolledbackIDs)
 }
 
+func listenAndBuildResponse(mgr *manager.Manager, changeID network.ID) ([]*gnmi.UpdateResult, error) {
+	networkChan := make(chan *network.NetworkChange)
+	errWatch := mgr.NetworkChangesStore.Watch(networkChan)
```
SeanCondon: It's a pity this Watch() can't be filtered by "changeID" . You could get a very big set of results here and you only want 1. If you could filter by "changeID" you might get max 3.

PR #793, opened 2019-10-22
Add options to change store Watch methods to allow watching for changes to specific devices

kuujo:
This PR adds new options to the `NetworkChange` and `DeviceChange` store `Watch` methods. The options allow callers to control log replay and filter events by specific change IDs.

To replay events when adding a watch, use `WithReplay`:
```go
store.Watch(ch, networkchangestore.WithReplay())
```

To watch changes to a specific object, use `WithChangeID`:
```go
store.Watch(ch, networkchangestore.WithChangeID(change.ID))
```


kuujo:
One concern with this pattern is that the underlying gRPC streams may not be closed once the watcher is done watching changes for a specific object. We may need another feature of the `Watch` method to ensure the caller can close the stream.

PR #795, opened 2019-10-22
Change python script to use python3

SeanCondon:
Python 2 End Of Life is 1st January 2020 - it has already been delayed 5 years

https://pythonclock.org/

I want to find a balance between what works on Mac and on Linux - this seems to work for both

https://howchoo.com/g/nwuznmzjmzm/how-to-install-and-use-python-3-on-macos

I will update the prerequisites.md, but that's in onos-docs


SeanCondon:
I think so - not to mind legacy onos, stratum etc.

PR #799, opened 2019-10-23
Support closing streams in client code

kuujo:
#793 mentioned there was no way to ensure streams are closed when a `Watch` call is partially completed.

This PR adds a `stream.Context` which is returned by `List` and `Watch` calls in stores. The `stream.Context` allows client code to `Close()` the stream when they're done with it. This will cause the Atomix clients to close underlying gRPC streams.

### DIFF
--- a/pkg/northbound/diags/diags_new_test.go
+++ b/pkg/northbound/diags/diags_new_test.go
@@ -21,6 +21,7 @@ import (
 	"github.com/onosproject/onos-config/pkg/certs"
 	"github.com/onosproject/onos-config/pkg/manager"
 	"github.com/onosproject/onos-config/pkg/northbound"
+	"github.com/onosproject/onos-config/pkg/store/stream"
 	mockstore "github.com/onosproject/onos-config/pkg/test/mocks/store"
 	changetypes "github.com/onosproject/onos-config/pkg/types/change"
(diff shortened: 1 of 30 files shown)


pkg/northbound/diags/diags_new_test.go
```diff
@@ -100,17 +101,17 @@ func Test_ListNetworkChanges(t *testing.T) {
...
 	assert.Assert(t, ok, "casting mock store")
-	mockNwChStore.EXPECT().List(gomock.Any()).DoAndReturn(func(ch chan<- *networkchangetypes.NetworkChange) error {
+	mockNwChStore.EXPECT().List(gomock.Any()).DoAndReturn(func(ch chan<- *networkchangetypes.NetworkChange) (stream.Context, error) {
 		// Send our network changes as a streamed response to store List()
 		go func() {
 			for _, nwch := range networkChanges {
 				ch <- nwch
 			}
-			// Finally send an empty one like the real List() does
-			ch <- nil
```
kuujo: @SeanCondon FYI the store does not send a `nil` value on the channel when it's finished. It just closes the channel, and a channel read returns `nil` when the channel is closed.

SeanCondon: ok

PR #800, opened 2019-10-23
Add device type/version cache

kuujo:
This PR adds a cache for device types/versions needed to validate changes on the northbound. The cache listens for `NetworkChange` events and keeps track of the devices that have been specified through the northbound API. Using the `NetworkChange` store events ensures the cache is sequentially consistent as it's based on data from the strongly ordered `NetworkChange` store. Northbound APIs should use the cache to determine the device type/version for changes before writing them to the store. So the store informs the cache, and the cache informs later writes to the store.

### DIFF
--- a/pkg/store/device/cache.go
+++ b/pkg/store/device/cache.go
@@ -42,15 +42,24 @@ type Info struct {
 type Cache interface {
 	io.Closer
 
+	// GetDevicesByID returns the devices that match the given device ID
+	GetDevicesByID(id device.ID) []*Info
+
+	// GetDevicesByType gets all devices of the given type
+	GetDevicesByType(deviceType string) []*Info
+
+	// GetDevicesByVersion gets all devices of the given type/version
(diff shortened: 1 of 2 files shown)


kuujo:
This is just a work in progress to get feedback. Need to add tests.

pkg/store/device/cache.go
```diff
@@ -0,0 +1,106 @@
...
+	Version  string
+}
+
+// Cache is a device type/version cache
+type Cache interface {
+	io.Closer
+
+	// GetDevices returns the set of devices in the cache
+	GetDevices() []Info
+}
```
SeanCondon: Would be useful to also have:
GetDevices(deviceID string) - because we want to get all versions of that ID
GetDevices(deviceVersion string, deviceType string) - because we want to get all devices of that type and version
GetDevices(deviceType string) - because we want to get all devices (all versions) of that type

kuujo: okay thanks!

kuujo:
Need to merge this for #784

PR #805, opened 2019-10-23
Previous - Computing the delta of a change that needs to be rolledback

### LINKED ISSUE #802: Rollback does not compute the delta of the change.
Putting a change in ROLLBACK, PENDING does not compute it's delta and apply it to the device. The change gets' propagated down to the device with the value that it held, meaning the last applied value for a certain element

To reproduce this please do:
`onit create cluster`
`onit add simulator`
`onit onos-cli`
`onos config watch network-changes`
`onit run test single-path`
get the latest change in the watch list that should be COMPLETE e.g. `mystifying_kowalevski`
`onit onos-cli`
`onos config rollback-new <change_id>`
mystifying_kowalevski was removing the value for `timezone/name`
get the logs and you will see something like 
```doRollback id:"mystifying_kowalevski:device-1813397878" index:12 revision:85 network_change:<id:"mystifying_kowalevski" index:12 > change:<device_id:"device-1813397878" device_version:"1.0.0" device_type:"Devicesim" values:<path:"/system/clock/config/timezone-name" value:<> removed:true > > status:<phase:ROLLBACK state:RUNNING > created:<seconds:1571840302 nanos:252215000 > updated:<seconds:1571840384 nanos:341037900 >```
if you are using the swap patch #796 the same empty value will be returned if you try to do 
```gnmi_cli -get -address onos-config:5150 -proto "path: <target: 'device-1813397878', elem: <name: 'system'> elem: <name: 'clock' > elem: <name: 'config'> elem: <name: 'timezone-name'>>" -timeout 5s -alsologtostderr -client_crt /etc/ssl/certs/client1.crt -client_key /etc/ssl/certs/client1.key -ca_crt /etc/ssl/certs/onfca.crt```

This in the value should instead contain the previous value `Europe/Dublin` that was set for that key. 

The rollback should compute the delta of the Change and send down the previous value for the given paths.

### PR DESCRIPTION
**Pull request by Andrea-Campanella**
This PR computes the full delta of the change that is requested to administratively rollback and applies it to the store as a new change containing all the reverse values for each path of the rollback requested one.
fixes #802

### DIFF
--- a/pkg/manager/rollbackconfig.go
+++ b/pkg/manager/rollbackconfig.go
@@ -20,6 +20,7 @@ import (
 	"github.com/onosproject/onos-config/pkg/events"
 	"github.com/onosproject/onos-config/pkg/store"
 	"github.com/onosproject/onos-config/pkg/store/change"
+	devicechangeutils "github.com/onosproject/onos-config/pkg/store/change/device/utils"
 	networkchangestore "github.com/onosproject/onos-config/pkg/store/change/network"
 	"github.com/onosproject/onos-config/pkg/store/stream"
 	changetypes "github.com/onosproject/onos-config/pkg/types/change"
@@ -111,8 +112,9 @@ func computeRollback(m *Manager, target string, configname store.ConfigName) (ch
 	return id, updates, deletes, nil
 }
 
-//Deprecated computeRollback works on old non atomix methods
+// computeRollback returns a change containing the previous value for each path of the rollbackChange
 func computeNewRollback(m *Manager, rollbackChange *networkchangetypes.NetworkChange) (*networkchangetypes.NetworkChange, error) {
+	//TODO make sure the changeID is unique.
 	deltaChange := &networkchangetypes.NetworkChange{
 		ID: networkchangetypes.ID(namesgenerator.GetRandomName(0)),
 		Status: changetypes.Status{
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: pkg/manager/rollbackconfig.go [hunk 1]
```diff
@@ -105,6 +110,58 @@ func computeRollback(m *Manager, target string, configname store.ConfigName) (ch
...
+			State:   changetypes.State_PENDING,
+			Reason:  changetypes.Reason_NONE,
+			Message: "Change generated for administratively requested rollback of " + string(rollbackChange.ID),
+		},
+	}
+	prevChanges := make([]*devicechangetypes.Change, 0)
+	for _, deviceChange := range rollbackChange.Changes {
+		previousValues := make([]*devicechangetypes.ChangeValue, 0)
+		for _, value := range deviceChange.Values {
+			preVal, err := m.GetTargetNewConfig(string(deviceChange.DeviceID), value.Path, 1)
```
kuujo: We might want to consider doing reverse iteration to get the previous value for a path instead of reading up to the previous change for the path. We should expect the change lists to get pretty large. We can even add a reverse iteration method to stores to stream previous changes from Raft nodes. I don’t think it’s something that needs to be fixed here, but we probably want to make a todo to optimize the way this computation is done since it’s coming from distributed stores.

kuujo:
I think this is fine. It seems to be safe to compute deltas outside the controllers since they only look back in time and not forward. Going to open another issue to further discuss the implementation of rollback.

pkg/manager/rollbackconfig.go
Andrea-Campanella: introducing a TODO

pkg/manager/rollbackconfig.go
```diff
@@ -106,6 +111,60 @@ func computeRollback(m *Manager, target string, configname store.ConfigName) (ch
 	return id, updates, deletes, nil
 }
 
+//Deprecated computeRollback works on old non atomix methods
```
SeanCondon: change the comment

pkg/manager/rollbackconfig.go
```diff
@@ -106,6 +111,60 @@ func computeRollback(m *Manager, target string, configname store.ConfigName) (ch
...
+func computeNewRollback(m *Manager, rollbackChange *networkchangetypes.NetworkChange) (*networkchangetypes.NetworkChange, error) {
+	deltaChange := &networkchangetypes.NetworkChange{
+		ID: networkchangetypes.ID(namesgenerator.GetRandomName(0)),
```
SeanCondon: should probably check that the name has not been used before

pkg/manager/rollbackconfig.go

Andrea-Campanella:
closing because we decide to implement it at the device level in the device/change/controller.go

PR #807, opened 2019-10-24
Added a Get option for DeviceChanges and NetworkChanges

SeanCondon:
In `diags` NBI:
- Corrected a problem with ListDeviceChanges - was not doing replay
- Made Replay programmable on ListDeviceChanges and ListNetworkChanges

In CLI:
- Added `onos config get network-changes` and `onos config get device-changes`
- Marked old `get config`, `get-changes` and `rollback` as deprecated
- Split `opstate` in to 2 - now in `onos config watch opstate` and `onos config get opstate` for the subscribe and non-subscribe options respectively


pkg/cli/devicechanges.go
```diff
@@ -74,3 +95,30 @@ func runWatchDeviceChangesCommand(cmd *cobra.Command, args []string) error {
...
+		wrapped = make([]string, linesC+1)
+	} else {
+		wrapped = make([]string, linesC)
+	}
+	var i int
+	for i = 0; i < linesC; i++ {
+		wrapped[i] = path[i*lineLen : i*lineLen+lineLen]
+	}
+	if over > 0 {
+		overFmt := fmt.Sprintf("%s-%ds", "%", lineLen)
```
Andrea-Campanella: is this % correct ? I was expecting a value

SeanCondon: it's creating a formatting string, which will be used in a further `printf` - weird how it highlights it here

PR #811, opened 2019-10-25
Refactor device change store to support per-version logs

kuujo:
This PR replaces #784 

This PR creates separate logs for each version of a device. This includes a lot of changes to add device versions throughout the stores, controllers, and northbound API.

New device types are added in a `types/device` package for the device `ID`, `Type`, `Version`, and `ID`+`Version`. These can be migrated to the topo API separately, but it was easier to include them in this PR for now.

This is a work in progress. I'm sure some of the changes to the northbound codes are not correct. Need to go through it more carefully, but I wanted to provide an opportunity to question the broader API changes and note any particularly important issues.


pkg/manager/manager.go
```diff
@@ -401,21 +409,30 @@ func (m *Manager) storeChange(configChange *change.Change) (change.ID, error) {
 	return configChange.ID, nil
 }
 
-//ExtractTypeAndVersion gets a deviceType and a Version based on the available store, topo and extension info
-func ExtractTypeAndVersion(target devicetopo.ID, storedDevice *devicetopo.Device, versionExt string, deviceTypeExt string) (TypeVersionInfo, error) {
+// ExtractTypeAndVersion gets a deviceType and a Version based on the available store, topo and extension info
+func (m *Manager) ExtractTypeAndVersion(target devicetopo.ID, storedDevice *devicetopo.Device, versionExt string, deviceTypeExt string) (TypeVersionInfo, error) {
```
SeanCondon: nice

pkg/northbound/gnmi/get.go
```diff
@@ -216,3 +236,16 @@ func buildUpdate(prefix *gnmi.Path, path *gnmi.Path, configValues []*devicechang
...
 	}, nil
 }
+
+func extractGetVersion(req *gnmi.GetRequest) (devicetype.Version, error) {
+	var version devicetype.Version
+	for _, ext := range req.GetExtension() {
+		if ext.GetRegisteredExt().GetId() == GnmiExtensionVersion {
+			version = devicetype.Version(ext.GetRegisteredExt().GetMsg())
+		} else {
+			return "", status.Error(codes.InvalidArgument, fmt.Errorf("unexpected extension %d = '%s' in Set()",
```
SeanCondon: Comment should say in `Get()`

Do we really want to throw error though?
We could be more forgiving in Get() and just log the unexpected extension

pkg/northbound/gnmi/get.go
```diff
@@ -159,6 +165,20 @@ func getUpdate(prefix *gnmi.Path, path *gnmi.Path) (*gnmi.Update, error) {
...
+	typeVersionInfo, errTypeVersion := manager.GetManager().ExtractTypeAndVersion(devicetopo.ID(target),
+		storedDevice, string(version), "")
+	if errTypeVersion != nil {
+		//TODO return instead of log
+		log.Error(errTypeVersion)
+		typeVersionInfo = manager.TypeVersionInfo{
+			DeviceType: "",
+			Version:    "",
+		}
+		//return nil, errTypeVersion
```
SeanCondon: for Get() i think it's fine to be tolerant of no version

pkg/northbound/gnmi/subscribe.go
```diff
@@ -433,3 +439,16 @@ func getChangeFromEvent(update events.ConfigEvent, mgr *manager.Manager) (string
...
 	return target, changeInternal
 }
+
+func extractSubscribeVersion(req *gnmi.SubscribeRequest) (devicetype.Version, error) {
+	var version devicetype.Version
+	for _, ext := range req.GetExtension() {
+		if ext.GetRegisteredExt().GetId() == GnmiExtensionVersion {
+			version = devicetype.Version(ext.GetRegisteredExt().GetMsg())
+		} else {
+			return "", status.Error(codes.InvalidArgument, fmt.Errorf("unexpected extension %d = '%s' in Set()",
```
SeanCondon: Also for subscribe we could just log that the extension is unexpected

PR #813, opened 2019-10-25
Remove references to deprecated stores from manager tests

ray-milkey:
This PR removes most reference to the deprecated store types from the manager tests. There are still some references in the manager code to these stores, so minimal ones are still created. Test logic has been updated to use the new APIs with mocks of the Atomix based stores.

### DIFF
--- a/pkg/manager/manager_test.go
+++ b/pkg/manager/manager_test.go
@@ -161,9 +161,36 @@ func setUp(t *testing.T) (*Manager, map[string]*change.Change, map[store.ConfigN
 		Created: now,
 	}
 
-	mockDeviceChangesStore.EXPECT().Get(deviceChange1.ID).Return(deviceChange1, nil).AnyTimes()
-	mockDeviceStore.EXPECT().Get(device1.ID).Return(device1, nil).AnyTimes()
+	deviceChanges := make(map[devicetopo.ID]*devicechange.DeviceChange)
+	deviceChanges[device1.ID] = deviceChange1
+
+	mockDeviceChangesStore.EXPECT().Get(deviceChange1.ID).DoAndReturn(
+		func(deviceID devicetopo.ID) (*devicechange.DeviceChange, error) {
+			return deviceChanges[deviceID], nil
+		}).AnyTimes()
 	mockNetworkChangesStore.EXPECT().Get(networkChange1.ID).Return(networkChange1, nil).AnyTimes()
+	mockNetworkChangesStore.EXPECT().List(gomock.Any()).DoAndReturn(
+		func(c chan<- *network.NetworkChange) error {
+			go func() {
+				c <- networkChange1
+				close(c)
+			}()
+			return nil
+		}).AnyTimes()
+	mockDeviceChangesStore.EXPECT().List(gomock.Any(), gomock.Any()).DoAndReturn(
+		func(deviceID devicetopo.ID, c chan<- *devicechange.DeviceChange) error {
+			deviceChange := deviceChanges[deviceID]
+			go func() {
+				if deviceChange != nil {
+					c <- deviceChange1
+				}
+				close(c)
+			}()
+			if deviceChange != nil {
+				return nil
+			}
+			return errors.New("no Configuration found")
+		}).AnyTimes()
 
 	mgrTest, err = NewManager(
 		&store.ConfigurationStore{
@@ -263,12 +290,12 @@ func Test_LoadManagerBadNetworkStore(t *testing.T) {
 	assert.Assert(t, err != nil, "should have failed to load manager")
 }
 
-func Test_GetNetworkConfig(t *testing.T) {
+func Test_GetNewNetworkConfig(t *testing.T) {
 
 	mgrTest, _, _, _, _, _ := setUp(t)
 
-	result, err := mgrTest.GetTargetConfig("Device1", "Running", "/*", 0)
-	assert.NilError(t, err, "GetTargetConfig error")
+	result, err := mgrTest.GetTargetNewConfig("Device1", "/*", 0)
+	assert.NilError(t, err, "GetTargetNewConfig error")
 
 	assert.Equal(t, len(result), 3, "Unexpected result element count")
 
@@ -463,7 +490,7 @@ func TestManager_GetAllDeviceIds(t *testing.T) {
 func TestManager_GetNoConfig(t *testing.T) {
 	mgrTest, _, _, _, _, _ := setUp(t)
 
-	result, err := mgrTest.GetTargetConfig("No Such Device", "Running", "/*", 0)
+	result, err := mgrTest.GetTargetNewConfig("No Such Device", "/*", 0)
 	assert.Assert(t, len(result) == 0, "Get of bad device does not return empty array")
 	assert.ErrorContains(t, err, "no Configuration found")
 }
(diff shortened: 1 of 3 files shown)


pkg/manager/manager_test.go
```diff
@@ -50,93 +57,161 @@ const (
...
+	DeviceType                = "TestDevice"
+	NetworkChange1 network.ID = "NetworkChange1"
+)
 
-	var change1 *change.Change
-	var device1config *store.Configuration
-	var mgrTest *Manager
+func setUp(t *testing.T) (*Manager, *mockstore.MockDeviceStore, *mockstore.MockNetworkChangesStore, *mockstore.MockDeviceChangesStore) {
 
+	log.SetOutput(os.Stdout)
```
Andrea-Campanella: an alternative to this is t.log()

pkg/manager/manager_test.go
```diff
@@ -50,93 +57,161 @@ const (
...
+		func(networkChange *networkchangetypes.NetworkChange) error {
+			networkChangesList = append(networkChangesList, networkChange)
+			return nil
+		}).AnyTimes()
+	mockNetworkChangesStore.EXPECT().Get(gomock.Any()).DoAndReturn(
+		func(id networkchangetypes.ID) (*networkchangetypes.NetworkChange, error) {
+			for _, networkChange := range networkChangesList {
+				if networkChange.ID == id {
+					return networkChange, nil
+				}
```
Andrea-Campanella: you can avoid this 'if' by passing the ID in the .Get() at line 126, basically it will fail if it receives anything other that that ID.
same applies for all other calls.

pkg/manager/manager_test.go
```diff
@@ -214,197 +299,194 @@ func Test_LoadManagerBadNetworkStore(t *testing.T) {
...
+	deletes := []string{Test1Cont1ACont2ALeaf2C}
+	updatesForDevice2, deletesForDevice2, deviceInfo2 := makeDeviceChanges("Device2", updates, deletes)
+	err := mgrTest.SetNewNetworkConfig(updatesForDevice2, deletesForDevice2, deviceInfo2, "Device2")
+	assert.NilError(t, err, "SetTargetConfig error")
+	updatesForDevice3, deletesForDevice3, deviceInfo3 := makeDeviceChanges("Device2", updates, deletes)
+	err = mgrTest.SetNewNetworkConfig(updatesForDevice3, deletesForDevice3, deviceInfo3, "Device3")
+	assert.NilError(t, err, "SetTargetConfig error")
 
 	deviceIds := mgrTest.GetAllDeviceIds()
+	// TODO - this doesn't see the new devices
```
Andrea-Campanella: i know i'm being a PITA but can you explain this TODO better, so all of use can create issues for these and reference the line, uncomment the .skip() and verify the passing of new code ?

pkg/manager/manager_test.go
```diff
@@ -430,72 +512,74 @@ func TestManager_GetAllConfig(t *testing.T) {
...
-	assert.Check(t, !bytes.Equal(config.Changes[len(config.Changes)-1], changeID), "Did not remove last change")
-	assert.Equal(t, changes[Test1Cont1ACont2ALeaf2B].Type, devicechangetypes.ValueType_FLOAT, "Wrong value to set after rollback")
-	assert.Equal(t, (*devicechangetypes.TypedFloat)(changes[Test1Cont1ACont2ALeaf2B]).Float32(), float32(ValueLeaf2B159), "Wrong value to set after rollback")
-
-	assert.Equal(t, changes[Test1Cont1ACont2ALeaf2A].Type, devicechangetypes.ValueType_FLOAT, "Wrong value to set after rollback")
-	assert.Equal(t, (*devicechangetypes.TypedFloat)(changes[Test1Cont1ACont2ALeaf2A]).Float32(), float32(ValueLeaf2B159), "Wrong value to set after rollback")
-
-	assert.Equal(t, deletesRoll[0], Test1Cont1ACont2ALeaf2D, "Path should be deleted")
+	// TODO - use new APIs to execute the roll back
+	//_, changes, deletesRoll, errRoll := computeRollback(mgrTest, Device1, *configName)
```
Andrea-Campanella: I think you can use here computeNewRollback()

pkg/manager/manager_test.go
ray-milkey: This was in the original tests, I just moved it from a Main() func into this set up one. I think the idea is to redirect all of the logging output of the entire system to the console for these tests.

pkg/manager/manager_test.go
ray-milkey: The Create() method can dynamically update the mock. You can't use static mocks here as the tests modify the store as they run.

pkg/manager/manager_test.go
ray-milkey: Sure. The real issue is that the GetAllDeviceIds() func is not yet implemented for the Atomix stores. I'll modify the comment accordingly.

pkg/manager/manager_test.go
ray-milkey: I updated the test to use the new API

Andrea-Campanella:
i'll discuss tomorrow with @SeanCondon but we can merge IMHO

pkg/manager/manager_test.go
```diff
@@ -214,197 +324,194 @@ func Test_LoadManagerBadNetworkStore(t *testing.T) {
...
 }
 
 // When device type is given it is like extension 102 and allows a never heard of before config to be created
 func Test_SetNetworkConfig_NewConfig(t *testing.T) {
-	mgrTest, changeStoreTest, configurationStoreTest, _, _, _ := setUp(t)
+	mgrTest, _, _, _ := setUp(t)
+
+	// Making change
+
+	const Device5 = "Device5"
```
SeanCondon: could these be lowercase?

pkg/manager/manager_test.go
```diff
@@ -214,197 +324,194 @@ func Test_LoadManagerBadNetworkStore(t *testing.T) {
...
+	updates[Test1Cont1ACont2ALeaf2A] = devicechangetypes.NewTypedValueFloat(ValueLeaf2B314)
+	deletes := []string{Test1Cont1ACont2ALeaf2C}
 
-	// Making change
-	updates[Test1Cont1ACont2ALeaf2A] = (*devicechangetypes.TypedValue)(devicechangetypes.NewTypedValueFloat(ValueLeaf2B314))
-	deletes = append(deletes, Test1Cont1ACont2ALeaf2C)
+	updatesForDevice, deletesForDevice, deviceInfo := makeDeviceChanges(Device5, updates, deletes)
 
-	changeID, configName, err := mgrTest.SetNetworkConfig("Device5", "1.0.0", "Devicesim", updates, deletes, "Test_SetNetworkConfig_NewConfig")
+	err := mgrTest.SetNewNetworkConfig(updatesForDevice, deletesForDevice, deviceInfo, NetworkChangeAddDevice5)
```
SeanCondon: `deviceInfo` is nil here - it should be "1.0.0" and "TestDevice"
in fact there is an error in Manager.computeNewNetworkConfig() that is ignoring it, and also it's putting the delete in to the update change - i'll tackle this in a separate PR

pkg/manager/manager_test.go
```diff
@@ -214,197 +324,194 @@ func Test_LoadManagerBadNetworkStore(t *testing.T) {
...
+	updatesForDevice, deletesForDevice, deviceInfo := makeDeviceChanges("Device6", updates, deletes)
 
-	_, _, err := mgrTest.SetNetworkConfig("Device6", "1.0.0", "", updates, deletes, "Test_SetNetworkConfig_NewConfig")
+	err := mgrTest.SetNewNetworkConfig(updatesForDevice, deletesForDevice, deviceInfo, "Test_SetNetworkConfig_NewConfig")
+	// TODO - the error for the missing type is currently not detecting the error of the unknown device type
+	t.Skip()
 	assert.ErrorContains(t, err, "no configuration found matching 'Device6-1.0.0' and no device type () given Please specify version and device type in extensions 101 and 102")
-	_, okupdate := configurationStoreTest["Device6-1.0.0"]
-	assert.Assert(t, !okupdate, "Expecting not to find Device6-1.0.0")
+	//_, okupdate := configurationStoreTest["Device6-1.0.0"]
```
SeanCondon: OK - will handle as a separate PR

Issue #806, opened 2019-10-23
Rollbacks implementation issues


kuujo:
I think it makes sense to append a new change to roll back an old one. This gives us a complete history of events. But after thinking about how to implement this safely, I realized a couple of issues:

1. Rollbacks must contain a reference to the original change

When the `NetworkChange` controller sees a rollback, it is responsible for ensuring the rollback is applied at the right time/in the correct order. That means ensuring a change is not rolled back before any later change to the device has been rolled back.

https://github.com/onosproject/onos-config/blob/master/pkg/controller/change/network/controller.go#L368

 But in order to make this determination, the controller must know _which_ change is being rolled back and at what time. To do this, the `NetworkChange` for rollbacks must include a reference to the original change ID.

I propose that the `NetworkChange` type use a `oneof` field to specify either a `Change` or a `Rollback`. The `Rollback` would include a reference to the original change being rolled back, and may or may not contain the computed `Change`. That would allow the `NetworkChange` controller to validate rollbacks and ensure they’re run in order.

2. We must differentiate between administrative rollbacks and automatic rollbacks

But appending rollbacks still creates a problem for automatic rollbacks due to partial failures.

First, I want to verify my understanding of the desired behavior is correct:
The sense I got from talking to @SeanCondon is that the failure of an individual device change should essentially never cause a network change to fail. So AIUI, we want a `NetworkChange` to remain enqueued to eventually be applied regardless of the failures that might occur during any single attempt to apply the change. So, in the case where the change is successful on one device but fails on another, we want to roll back the change to ensure we maintain consistency in the network, but following that rollback the change should remain enqueued to be attempted again later. The change will remain enqueued until an administrative rollback. 

https://github.com/onosproject/onos-config/blob/master/pkg/controller/change/network/controller.go#L208

In the case of automatic rollback, the problem is if we append a new change for the rollback without also appending a new change for the retry, we’ll create an erratic ordering of changes and inconsistent state. A change could be followed by multiple automatic rollbacks which would appear to cancel the original change. But to maintain the order of changes, if the controller were to append a new change for each retry, it would also have to append new changes for all later changes to maintain the order of changes.

Because of this, and assuming my understanding of the intended response to failures is correct, I suspect we _don’t_ want to apply a new change in the case of automatic rollbacks. Instead, we can use a third phase - `CANCEL` - to separate an administrative rollback which is permanent and is thus appended to the log from an automatic rollback which is an ephemeral state of the original change.

[...]

PR #814, opened 2019-10-28
Moving the southbound connection to controller.go for Atomix based stores

Andrea-Campanella:
This PR transitions the interaction with the device to the `change/device/controller.go` so it's based on the new atomix based flow. 

WIP, needs further testing and rebasing.

### DIFF
--- a/pkg/controller/change/device/controller.go
+++ b/pkg/controller/change/device/controller.go
@@ -168,7 +168,8 @@ func (r *Reconciler) translateAndSendChange(change *devicechangetypes.Change) er
 		return err
 	}
 	log.Info("Reconciler set request ", setRequest)
-	deviceTarget, err := southbound.GetTarget(change.DeviceID)
+	log.Info("Device ", change.DeviceID)
+	deviceTarget, err := southbound.GetTarget(devicetopo.ID(change.DeviceID))
 	if err != nil {
 		//TODO revisit change state.
 		log.Infof("Device %s is not connected, accepting change", change.DeviceID)
--- a/pkg/controller/change/device/controller.go
+++ b/pkg/controller/change/device/controller.go
@@ -17,6 +17,7 @@ package device
 import (
 	"fmt"
 	"github.com/onosproject/onos-config/pkg/controller"
+	"github.com/onosproject/onos-config/pkg/southbound"
 	changestore "github.com/onosproject/onos-config/pkg/store/change/device"
 	devicechangeutils "github.com/onosproject/onos-config/pkg/store/change/device/utils"
 	devicestore "github.com/onosproject/onos-config/pkg/store/device"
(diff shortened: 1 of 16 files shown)


pkg/controller/change/device/controller.go
```diff
@@ -158,17 +153,34 @@ func (r *Reconciler) reconcileRollback(change *devicechangetypes.DeviceChange) (
...
+	setRequest, err := values.NativeNewChangeToGnmiChange(change)
+	if err != nil {
+		return err
+	}
+	log.Info("Reconciler set request ", setRequest)
+	log.Info("Device ", change.DeviceID)
+	deviceTarget, err := southbound.GetTarget(devicetopo.ID(change.DeviceID))
+	if err != nil {
+		//TODO revisit change state.
+		log.Infof("Device %s is not connected, accepting change", change.DeviceID)
```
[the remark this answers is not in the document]
kuujo: I would suggest we still return the error here. The `DeviceChange` will be `FAILED` with an `ERROR`, but since we  currently treat all errors as retryable regardless of their context, the change will be retried by the network controller.

Andrea-Campanella:
@kuujo I completly agree with your points. I returned an error in the latest version of the patch. 
Altough I'd still like to have a mechanism that prevents re-trying on `OFFLINE` devices. Either the Network Controller or the Change controller should have machinery to deal with this scenario. It's outside the scope of this patch but there are a few options to do a retry when a device is offline:
1) Retry a finite amount of time 1/2 and the hold the retry until a device connected event arrives
2) Exponential backoff retry with interception of a device connected event.

In any case the system should listen to a device-connected even, check for existing `PENDING` configs and apply them immediately. The old workflow was doing so here:
https://github.com/onosproject/onos-config/blob/master/pkg/southbound/synchronizer/synchronizer.go#L103-L139
We held a similar discussion here for rollback here:
https://github.com/onosproject/onos-config/issues/806

we might want to merge the two approaches.

pkg/manager/setconfig.go
```diff
@@ -262,6 +264,7 @@ func (m *Manager) computeNewNetworkConfig(targetUpdates map[string]devicechanget
 		}
 		log.Infof("Appending device change %v", newChange)
 		deviceChanges = append(deviceChanges, newChange)
+		delete(targetRemoves, target)
```
SeanCondon: this clashes with the patch I made - but I prefer your solution - cleaner

pkg/northbound/gnmi/set.go
```diff
@@ -99,6 +99,7 @@ func (s *Server) Set(ctx context.Context, req *gnmi.SetRequest) (*gnmi.SetRespon
 		return nil, extErr
 	}
 
+	//TODO update to new methods
```
SeanCondon: I removed this old call and make the new one down a bit in the function

PR #819, opened 2019-10-29
Removing unused CLI commands for old change, config and netchanges

SeanCondon:
No one uses these any more for unit or integration tests


SeanCondon:
Integration tests are working for me with this

PR #827, opened 2019-10-31
Changed how checkForReadOnly works

SeanCondon:


pkg/northbound/gnmi/set_test.go
```diff
@@ -954,16 +954,16 @@ func TestSet_checkForReadOnly(t *testing.T) {
 	updateT1[cont1aCont2aLeaf2b] = devicechangetypes.NewTypedValueString("2b on t1")
 
 	updateT2 := make(map[string]*devicechangetypes.TypedValue)
-	updateT1[cont1aCont2aLeaf2a] = devicechangetypes.NewTypedValueUint64(11)
-	updateT1[cont1aCont2aLeaf2b] = devicechangetypes.NewTypedValueString("2b on t2")
-	updateT1[cont1aCont2aLeaf2c] = devicechangetypes.NewTypedValueString("2c on t2 - ro attribute")
+	updateT2[cont1aCont2aLeaf2a] = devicechangetypes.NewTypedValueUint64(11)
```
SeanCondon: I think this is the reason my test was passing but the function was not working like it should

PR #852, opened 2019-11-05
Changed watch for topo device store to get reconnect Fixes #828

### LINKED ISSUE #828: When onos-config is restarted, it does not reconnect to devices
**Describe the bug**
When `onos-config` is restarted, it does not reconnect to devices. 

**To Reproduce**
With `onit` create a cluster with 2 simulators
```
go run github.com/onosproject/onos-test/cmd/onit create cluster sean-1 --image-tags="topo=latest,config=latest,ztp=latest" && onit add simulator devicesim-1 && onit add simulator devicesim-2
```

Using `onit onos-cli` to get in to the *cli* instance and verify that the devices can be seen.
```
onos config get opstate devicesim-1
```

Then redeploy `onos-config` with:
```
onit set image onos-config --image onosproject/onos-config:latest
```
Wait for the pod to be replaced - progress can be tracked with:
```
watch kubectl -n sean-1 get pods
```

Now try getting opstate again 
```
~ $ onos config get opstate devicesim-1
OPSTATE CACHE: devicesim-1
PATH                                                                              |VALUE               |
rpc error: code = Unknown desc = no Operational State cache available for devicesim-1
~ $ 
```

**Expected behavior**
opstate access here is just a symptom - the root cause is that onos-config does not query onos-topo after restarting and so is not aware of the presence of the device

**Logs, text or screenshots**

**Additional context**

SeanCondon:


pkg/southbound/synchronizer/factory.go
```diff
@@ -35,9 +35,10 @@ func Factory(topoChannel <-chan *devicetopo.ListResponse, opStateChan chan<- eve
 	modelRegistry *modelregistry.ModelRegistry, operationalStateCache map[devicetopo.ID]devicechangetypes.TypedValueMap) {
 
 	for topoEvent := range topoChannel {
+		log.Infof("Receieved TopoChannel event %v %v", topoEvent.Type, topoEvent)
 		notifiedDevice := topoEvent.Device
 		//TODO evaluate the need for fine-grained checking
-		if topoEvent.Type == devicetopo.ListResponse_ADDED {
+		if topoEvent.Type == devicetopo.ListResponse_NONE || topoEvent.Type == devicetopo.ListResponse_ADDED {
```
Andrea-Campanella: please comment what NONE is, otherwise it's a bit opaque.

PR #857, opened 2019-11-05
Replay changes in device cache

kuujo:
Fixes a bug in the device cache to force it to replay changes at startup.


kuujo:
Oops broke a test 🤷‍♂

PR #858, opened 2019-11-05
Implement a check for bad type info when computing a network config

ray-milkey:
This PR allows us to enable tests that are looking for errors when a Set() operation is requested on a non existent device, and there is no device type information specified.

### DIFF
--- a/pkg/manager/setconfig.go
+++ b/pkg/manager/setconfig.go
@@ -22,6 +22,8 @@ import (
 	networkchangetypes "github.com/onosproject/onos-config/pkg/types/change/network"
 	devicetype "github.com/onosproject/onos-config/pkg/types/device"
 	"github.com/onosproject/onos-config/pkg/utils"
+	"google.golang.org/grpc/codes"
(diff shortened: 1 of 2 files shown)


pkg/manager/setconfig.go
```diff
@@ -99,6 +101,10 @@ func (m *Manager) computeNewNetworkConfig(targetUpdates map[string]devicechanget
 		//FIXME this is a sequential job, not parallelized
 		version := deviceInfo[devicetype.ID(target)].Version
 		deviceType := deviceInfo[devicetype.ID(target)].Type
+		deviceType, version, err := mgr.CheckCacheForDevice(devicetype.ID(target), deviceType, version)
```
SeanCondon: this was already done at the `Set` level - line 112 of `set.go`. Also whatever you do to updates, you must do to **removes** also - i.e. the loop below

Andrea-Campanella: I'd avoid doing this check there, duplicated cycles spent. move the tests in set_test.go and we shoudl be ok.

Issue #824, opened 2019-10-30
Change to logging framework with more granular logging


kuujo:
**Is your feature request related to a problem? Please describe.**
The klog library does not support more granular logging levels. Distributed systems components will log massive amounts of data in e.g. benchmarks. 

**Describe the solution you'd like**
We need to be able to do trace/debug logging and control the logging output. We might want to investigate a new logging framework for this reason. The Uber framework supports granular logging and claims exceptional performance:
https://github.com/uber-go/zap

**Describe alternatives you've considered**
As in any language, Go has many logging frameworks. We should take some time to investigate them.

**Additional context**
I intend to experiment with the zap library in Atomix first. Hopefully that can inform this issue.

Andrea-Campanella:
done

PR #859, opened 2019-11-06
Refactored to separate public APIs into top-level /api folder; fixes …

tomikazi:
…issue #824


tomikazi:
This represents a bit of a departure from the standard Go project layout however. I’d like to see Sean’s, Andrea’s and Adib’s view on this. I’m wondering whether we ought to put it under /pkg/api instead to stick with the convention and still achieve the desired goal. It would be Go-specific, but it really already is the way it is now.

pkg/cli/networkchanges_test.go
```diff
@@ -18,10 +18,10 @@ package cli
 import (
 	"bytes"
 	"fmt"
-	"github.com/onosproject/onos-config/pkg/northbound/diags"
-	"github.com/onosproject/onos-config/pkg/types/change"
-	"github.com/onosproject/onos-config/pkg/types/change/device"
-	"github.com/onosproject/onos-config/pkg/types/change/network"
+	"github.com/onosproject/onos-config/api/diags"
+	"github.com/onosproject/onos-config/api/types/change"
+	"github.com/onosproject/onos-config/api/types/change/device"
```
SeanCondon: should use consistent prefix throughout for this package - called `devicechange` else where

pkg/controller/change/device/computedelta_test.go
```diff
@@ -16,10 +16,10 @@ package device
 
 import (
 	"fmt"
+	"github.com/onosproject/onos-config/api/types"
+	"github.com/onosproject/onos-config/api/types/change"
+	devicechangetypes "github.com/onosproject/onos-config/api/types/change/device"
```
SeanCondon: This is why `github.com/onosproject/onos-config/api/types/change/device` was given the alias `devicechangetypes` consistently through out the code. `devicechanges` had been used for the package `github.com/onosproject/onos-config/pkg/types/change/device`

I don't care what alias is used for either as long as the usage is consistent throughout the project. It includes code readability and consistency

Andrea-Campanella:
@tomikazi I like the approach of having the APIs in the too level `/api` folder, it makes it easier for somebody to go and check them out. Kubernetes takes it one step further with https://github.com/kubernetes/api repo, but I prefer the separation ATM. 
I also agree with both @SeanCondon comments, you need to be consistent with the aliases across the whole repo or it make a big mess of understanding exactly what we are pointing at. 
refactoring it in small chunks also makes it easier to spot if there is an issue. I'd split this one up in multiple PRs.

tomikazi:
Issuing a new PR with resolved conflicts.

PR #861, opened 2019-11-06
Removing new from refactored methods

Andrea-Campanella:


pkg/manager/setconfig.go
```diff
@@ -99,29 +99,29 @@ func (m *Manager) computeNewNetworkConfig(targetUpdates map[string]devicechanget
 		//FIXME this is a sequential job, not parallelized
 		version := deviceInfo[devicetype.ID(target)].Version
 		deviceType := deviceInfo[devicetype.ID(target)].Type
-		newChange, err := m.ComputeNewDeviceChange(
```
SeanCondon: I thought `newChange` here was actually a good name for the **new** change

pkg/cli/networkchanges_test.go
```diff
@@ -34,7 +34,7 @@ func generateNetworkChangeData(count int) {
 	now := time.Now()
 
 	for cfgIdx := range networkChanges {
-		networkID := fmt.Sprintf("a_new_network_change-%d", cfgIdx)
```
SeanCondon: I thought the existing name was fine for a **new** network change

pkg/manager/setconfig.go
Andrea-Campanella: brought back

pkg/cli/networkchanges_test.go
Andrea-Campanella: reverted

PR #868, opened 2019-11-06
Listen to the Device Cache in the Device Watcher Fixes #866

### LINKED ISSUE #866: gNMI Set on new "unconnected" device never completes
**Describe the bug**
When a Set() operation is done on an unconnected device that has not previously been created in the system, the Set() hangs

**To Reproduce**

- Use the `onit` tool to create a new cluster - does not matter if simulators are present
- Run the following gNMI Set() command on the system
> Use 'onit onos-cli' to get access to the `gnmi_cli` too on the command line
```
gnmi_cli -address onos-config:5150 -set \
    -proto "update: <path: <target: 'my-td', elem: <name: 'cont1a'> elem: <name: 'cont2a'> elem: <name: 'leaf2a'>> val: <uint_val: 12>> extension: <registered_ext: <id: 100, msg:'update_devicesim-1'>> extension: <registered_ext: <id: 101, msg:'1.0.0'>> extension: <registered_ext: <id: 102, msg:'TestDevice'>>" \
    -timeout 5s -alsologtostderr -client_crt /etc/ssl/certs/client1.crt -client_key /etc/ssl/certs/client1.key -ca_crt /etc/ssl/certs/onfca.crt
```

**Expected behavior**
The Set() should return and the Device Change should be marked as `COMPLETE`. The system is meant to be able to take new configurations even though there's no device `my-td` created in the Topology system.

**Logs, text or screenshots**
```
I1106 13:47:51.083427       1 set.go:62] gNMI Set Requestupdate:<path:<elem:<name:"cont1a" > elem:<name:"cont2a" > elem:<name:"leaf2a" > target:"my-td" > val:<uint_val:12 > > extension:<registered_ext:<id:100 msg:"update_devicesim-1" > > extension:<registered_ext:<id:101 msg:"1.0.0" > > extension:<registered_ext:<id:102 msg:"TestDevice" > > 
I1106 13:47:51.083559       1 set.go:236] Set called with extensions; 100: update_devicesim-1, 101: , 102: 
I1106 13:47:51.083569       1 set.go:236] Set called with extensions; 100: update_devicesim-1, 101: 1.0.0, 102: 
I1106 13:47:51.083573       1 set.go:236] Set called with extensions; 100: update_devicesim-1, 101: 1.0.0, 102: TestDevice
I1106 13:47:51.085648       1 manager.go:243] Device my-td not found in topo store
I1106 13:47:51.087490       1 set.go:421] Validating change my-td:TestDevice:1.0.0
I1106 13:47:51.089292       1 setconfig.go:66] New Configuration for my-td, with version 1.0.0 and type TestDevice, is Valid according to model TestDevice-1.0.0
I1106 13:47:51.089306       1 set.go:429] Validating change my-td:TestDevice:1.0.0 DONE
I1106 13:47:51.089314       1 set.go:330] Testing /cont1a/cont2a/leaf2a for read only
I1106 13:47:51.089366       1 setconfig.go:108] Appending device change device_id:"my-td" device_version:"1.0.0" device_type:"TestDevice" values:<path:"/cont1a/cont2a/leaf2a" value:<bytes:"\014\000\000\000\000\000\000\000" type:UINT > > 
I1106 13:47:51.096094       1 controller.go:137] Creating DeviceChange network_change:<id:"update_devicesim-1" index:1 > change:<device_id:"my-td" device_version:"1.0.0" device_type:"TestDevice" values:<path:"/cont1a/cont2a/leaf2a" value:<bytes:"\014\000\000\000\000\000\000\000" type:UINT > > > status:<> created:<seconds:-62135596800 > updated:<seconds:-62135596800 > 
I1106 13:47:51.098934       1 set.go:365] Received notification for change ID update_devicesim-1, phase CHANGE, state PENDING
I1106 13:47:51.101388       1 controller.go:105] Running NetworkChange id:"update_devicesim-1" index:1 revision:213 status:<state:RUNNING > created:<seconds:1573048071 nanos:90256603 > updated:<seconds:1573048071 nanos:98241848 > changes:<device_id:"my-td" device_version:"1.0.0" device_type:"TestDevice" values:<path:"/cont1a/cont2a/leaf2a" value:<bytes:"\014\000\000\000\000\000\000\000" type:UINT > > > refs:<device_change_id:"update_devicesim-1:my-td:1.0.0" > 
I1106 13:47:51.102964       1 set.go:365] Received notification for change ID update_devicesim-1, phase CHANGE, state RUNNING
I1106 13:47:51.105759       1 controller.go:236] Running DeviceChange id:"update_devicesim-1:my-td:1.0.0" index:1 revision:177 network_change:<id:"update_devicesim-1" index:1 > change:<device_id:"my-td" device_version:"1.0.0" device_type:"TestDevice" values:<path:"/cont1a/cont2a/leaf2a" value:<bytes:"\014\000\000\000\000\000\000\000" type:UINT > > > status:<state:RUNNING > created:<seconds:1573048071 nanos:96935412 > updated:<seconds:1573048071 nanos:96935412 > 
```

Running the `onos` CLI in another windows I can see that the Network Change and the Device Change are present, but the Device Change is not complete.
```
~ $ onos config get network-changes
CHANGE                          INDEX  REVISION  PHASE    STATE     REASON   MESSAGE
update_devicesim-1              1       214      CHANGE   RUNNING   NONE     
	Device: my-td (1.0.0)

~ $ onos config get device-changes my-td
CHANGE                          INDEX  REVISION  PHASE    STATE     REASON   MESSAGE
update_devicesim-1:my-td:1.0.0  1       178      CHANGE   RUNNING   NONE     
	update_devicesim-1	1	my-td	1.0.0
	|/cont1a/cont2a/leaf2a                             |(UINT) 12                               |false  |

~ $
```

**Additional context**
I would expect `translateAndSendChange()` and `doChange()` to be called in the `device/controller.go` but nothing from its logs show up.

SeanCondon:
Added a Watch function on the Device Cache

Changed the network/contoller and the device/controller to *watch* on the `DeviceCache` rather than the `DeviceStore` because the latter will have all configurable devices in it (even new ones as creating something in the `NetworkChangesStore` updates the cache

> One small issue that needs attention is the double **send** to the channel in cache.go listen()


pkg/store/device/cache.go
```diff
@@ -142,6 +158,12 @@ func (c *networkChangeStoreCache) GetDevices() []*Info {
 	return devices
 }
 
+func (c *networkChangeStoreCache) Watch(ch chan<- *Info) {
```
kuujo: In this implementation, a `Watch` only sees devices that are added after the watch call. Controllers need to see all past and future devices to avoid race conditions resulting in missed events. The `Watch` method could instead look a lot like the `listen()` method does now. It would call `c.networkChangeStore.Watch(ch, networkchangestore.WithReplay())` and pass a new channel down to that call. Replaying the entire list of devices is really important for controllers to ensure events are not lost.

Also, we probably want to use the `stream.Context` that's used in other stores to avoid memory leaks:
https://github.com/onosproject/onos-config/blob/ed8b26d85b55cae8cfc485be311fe87e94908855/pkg/store/change/network/store.go#L135

This allows the `Watch` caller to close the stream when it stops listening. In controller watchers, when the watcher is stopped the stream is closed. Otherwise state changes in the cluster can result in the controllers continuously adding channels that are never removed.

SeanCondon: what I was concerned about is that watcher of this cache only wants the **updates** to the **cache** (which is only the device type and version) and not for instance to get an event every time a network change is modified. To do this it has to keep the current list of cache items, and only update it in one place - this is currently done in **listen**.

I get it that a replay is needed and the stream.Context - I will work on these, but I was reluctant to send up every network change

pkg/store/device/cache.go
```diff
@@ -142,6 +163,39 @@ func (c *networkChangeStoreCache) GetDevices() []*Info {
...
+// A replay option allows former entries to be replayed to the caller
+func (c *networkChangeStoreCache) Watch(ch chan<- stream.Event, replay bool) (stream.Context, error) {
+	c.mu.RLock()
+	_, ok := c.listeners[ch]
+	c.mu.RUnlock()
+	if ok {
+		return nil, fmt.Errorf("already listening to channel %v", ch)
+	}
+	if replay {
+		c.mu.RLock()
```
kuujo: There's a race condition with how this is ordered. Between replaying the existing devices and adding the new channel to `listeners`, events that are received will be missed by the new listener. We might want to just lock the entire replay when it's requested to ensure past events are serialized with future events.

SeanCondon: ok - will have a look

SeanCondon: OK - hopefully I have addressed it - I put a single lock around the replay and the adding of the listener - it should not be possible for the listen() loop to inject any new events while we are doing this because it will not be able to acquire the lock. I hope we cannot miss any - I presume the channel on the network change store will wait for the lock.

PR #872, opened 2019-11-07
New Manager Test setup to use real controllers

SeanCondon:
I think the manager test can use a lot less mocking and instead use the real controllers and `Local` atomix stores. I think this will help testing and understanding of how the manager interacts with the controllers in a local unit test environment

This has been changed to have a new test file `manager_deep_test.go` to hold a new setup and copied some of the tests over. Just those essential to cover some essential cases.

Not all of the tests are suitable for using this `deep` mocking, and are better being tested with higher level controller mocks. Also I did not want to remove all of the mocks setup


pkg/controller/change/network/controller_test.go
```diff
@@ -323,26 +317,12 @@ func TestReconcilerError(t *testing.T) {
...
-
-	devices, err := devicestore.NewStore(client)
-	assert.NoError(t, err)
+func newStores(t *testing.T) (networkchanges.Store, devicechanges.Store) {
 	networkChanges, err := networkchanges.NewLocalStore()
 	assert.NoError(t, err)
 	deviceChanges, err := devicechanges.NewLocalStore()
 	assert.NoError(t, err)
-	return devices, networkChanges, deviceChanges
+	return networkChanges, deviceChanges
```
SeanCondon: devices had not been used in either of the 2 cases above

pkg/manager/manager.go
```diff
@@ -104,21 +104,15 @@ func NewManager(leadershipStore leadership.Store, mastershipStore mastership.Sto
...
-		return nil, err
-	}
-
 	return NewManager(leadershipStore, mastershipStore, deviceChangesStore, deviceStore, deviceCache,
 		networkChangesStore, networkSnapshotStore, deviceSnapshotStore, topoChannel)
 }
 
 // Run starts a synchronizer based on the devices and the northbound services.
-func (m *Manager) Run() {
+func (m *Manager) Run(targetFn func() southbound.TargetIf) {
```
SeanCondon: this is necessary to allow a target to be mocked

pkg/manager/manager_deep_test.go
```diff
@@ -0,0 +1,279 @@
...
+	// Set the new change
+	const testNetworkChange networkchange.ID = "Test_SetNetworkConfig"
+	err := mgrTest.SetNetworkConfig(updatesForDevice1, deletesForDevice1, deviceInfo, string(testNetworkChange))
+	assert.NilError(t, err, "SetTargetConfig error")
+	time.Sleep(time.Millisecond * 10) // Have to wait for change to be reconciled
+	testUpdate, _ := mgrTest.NetworkChangesStore.Get(testNetworkChange)
+	assert.Assert(t, testUpdate != nil)
+	assert.Equal(t, testUpdate.ID, testNetworkChange, "Change Ids should correspond")
+	assert.Equal(t, changetypes.Phase_CHANGE, testUpdate.Status.Phase)
+	assert.Equal(t, changetypes.State_COMPLETE, testUpdate.Status.State)
```
SeanCondon: this is the most essential line in the whole patch - checks that the Network change is complete - i.e. has gone down to device controller and on to synchronizer, made the change (to a mocked target), got its result back and the reconciler marked the `DeviceChange` and the `NetworkChange` complete

cmd/onos-config/onos-config.go
```diff
@@ -159,7 +165,7 @@ func main() {
 			}
 		}
 
-		mgr.Run()
+		mgr.Run(southbound.NewTarget)
```
Andrea-Campanella: I know this is necessary for a target to be mocked but it completely breaks boundaries and abstractions, this is a no go for the whole patch. the target is an interface in the SB. it should never be placed in the cmd for the manager to run. we need another way to pass this in it can't come from here.

Andrea-Campanella: what I would do instead is to change only the SB code with a public variable (accessibile from tests with southbound.BuildFunction or something similar) in factory that contains the TargetFn build function. that variable is set to southbound.NewTarget for production code. Before doing a .run() on the manager you swap it for `mockTargetCreationfn`
this is a solution that touches just the factory and the SB, with NO appearances in upper levels

pkg/manager/manager.go
```diff
@@ -148,7 +142,7 @@ func (m *Manager) Run() {
 	go listenOnResponseChannel(m.SouthboundErrorChan, m)
 	//TODO we need to find a way to avoid passing down parameter but at the same time not hve circular dependecy sb-mgr
 	go synchronizer.Factory(m.TopoChannel, m.OperationalStateChannel, m.SouthboundErrorChan,
-		m.Dispatcher, m.ModelRegistry, m.OperationalStateCache)
+		m.Dispatcher, m.ModelRegistry, m.OperationalStateCache, targetFn)
```
Andrea-Campanella: this will be removed with my other approach

cmd/onos-config/onos-config.go
SeanCondon: What about if this was passed nil and the southbound.NewTarget was taken to be the default implementation (down in Factory or synchronizer). In that way we could spare referencing southbound from here, which is what I think you objection is

Andrea-Campanella: no, my objection is making the initialisation of the manager on something that should pertain only in the southbound.
what do you think of the variable solution ?

PR #873, opened 2019-11-07
Updating helm-chart

Andrea-Campanella:


deployments/helm/onos-config/templates/deployment.yaml
```diff
@@ -43,7 +43,9 @@ spec:
             - name: ATOMIX_APP
               value: {{ template "fullname" . }}
             - name: ATOMIX_RAFT_GROUP
-              value: {{ template "fullname" . }}-raft
+              value: raft
```
kuujo: This shouldn't work. The name of the Raft `PartitionSet` is `{{ template "fullname" . }}-raft`

Andrea-Campanella: I took it from
https://github.com/onosproject/onos-test/blob/master/pkg/onit/k8s/onos-config.go#L145
does this make sense ?

Andrea-Campanella:
Closing #910

PR #881, opened 2019-11-08
Fix data races in unit tests

ray-milkey:
This PR enables race detection in the unit tests and fixes several data race problems.


pkg/southbound/synchronizer/synchronizer_test.go
```diff
@@ -217,8 +218,14 @@ func TestNew(t *testing.T) {
 	).Return(nil).MinTimes(1)
 
 	// Called asynchronously as after building up the opStateCache it subscribes and waits
-	go s.syncOperationalStateByPaths(context2.Background(), mockTarget, params.responseChan)
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go func() {
+		s.syncOperationalStateByPaths(context2.Background(), mockTarget, params.responseChan)
+		wg.Done()
```
SeanCondon: You could have the wg.Done() in the syncOperationalStateByPaths() but i guess it's the same thing

pkg/southbound/synchronizer/synchronizer_test.go
```diff
@@ -217,8 +218,14 @@ func TestNew(t *testing.T) {
...
+	}()
 
+	wg.Wait()
 	time.Sleep(200 * time.Millisecond) // Wait for response message
```
SeanCondon: can remove the sleep now i suppose

PR #883, opened 2019-11-08
Add locks to references to the opstate cache to prevent data races

ray-milkey:
This PR adds mutex locks to references to the opstate cache to prevent data race conditions.

### DIFF
--- a/pkg/southbound/synchronizer/factory.go
+++ b/pkg/southbound/synchronizer/factory.go
@@ -25,7 +25,7 @@ import (
 	"github.com/onosproject/onos-config/pkg/utils"
 	topodevice "github.com/onosproject/onos-topo/api/device"
 	log "k8s.io/klog"
-	"sync"
+	syncPrimitives "sync"
 )
 
 // Factory is a go routine thread that listens out for Device creation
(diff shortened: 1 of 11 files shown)


ray-milkey:
Added DNM label because integration tests fail.

pkg/southbound/synchronizer/factory.go
```diff
@@ -54,10 +56,12 @@ func Factory(topoChannel <-chan *topodevice.ListResponse, opStateChan chan<- eve
 			} else {
 				mStateGetMode = modelregistry.GetStateMode(mPlugin.GetStateMode())
 			}
+			operationalStateCacheLock.Lock()
 			operationalStateCache[notifiedDevice.ID] = make(devicechange.TypedValueMap)
 			target := southbound.NewTarget()
 			sync, err := New(ctx, notifiedDevice, opStateChan, southboundErrorChan,
 				operationalStateCache[notifiedDevice.ID], mReadOnlyPaths, target, mStateGetMode)
+			operationalStateCacheLock.Unlock()
```
Andrea-Campanella: I think here you are keeping the lock for too long. By keeping it until the return of the New method you are not enabling the synchroniser to acquire the lock in order to populate the cache during startup and this might cause the failure you were seeing in the integration tests run. to me the Unlock shoudl happen after line 60, right after make.

pkg/southbound/synchronizer/factory_test.go
```diff
@@ -16,9 +16,9 @@ package synchronizer
 
 import (
 	devicechange "github.com/onosproject/onos-config/api/types/change/device"
-	"github.com/onosproject/onos-config/pkg/dispatcher"
+	dispatcherpkg "github.com/onosproject/onos-config/pkg/dispatcher"
```
Andrea-Campanella: why did you go with the pkg ?

pkg/southbound/synchronizer/factory.go
SeanCondon: I think you should be passing the operationalStateCacheLock in tot he synchronizer with New() and replace the mutex created in there in synchronizer.go:65

pkg/southbound/synchronizer/factory_test.go
tomikazi: I suspect this is to disambiguate the package 'dispatcher' from the object reference 'dispatcher'.

ray-milkey: @SeanCondon is correct. There were clashes between variable names and the import name.

PR #897, opened 2019-11-14
Enforce manadatory check against plugins Fixes #890

### LINKED ISSUE #890: An start time flag should allow mandatory checks against plugins
**Is your feature request related to a problem?**
Currently when a configuration change is made through gnmi SetRequest a device type and version can be given. If there is not a model plugin for this type and version the configuration can be any free form structure, and the validation check will be skipped.

Some users may not want this, and instead may want validation to be mandatory. This feature is to enable this as a start time option

**Additional context**

SeanCondon:
by default only configurations that have a model plugin will be allowed

The flag on the command line will allow this to be disabled for testing purposes or for users who want to use the system without having to create model plugins

For unit tests the flag option is set to true, but can be easliy changed

Also in this patch I have increased the timeout for tests failing, as i sometimes became a problem

### DIFF
--- a/cmd/onos-config/onos-config.go
+++ b/cmd/onos-config/onos-config.go
@@ -20,6 +20,7 @@ gives a gNMI interface northbound for other systems to connect to, and an
 Admin service through gRPC
 
 Arguments
+-allowUnvalidatedConfig <allow configuration for devices without a corresponding model plugin>
 
 -modelPlugin (repeated) <the location of a shared object library that implements the Model Plugin interface>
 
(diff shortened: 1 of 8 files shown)


cmd/onos-config/onos-config.go
```diff
@@ -70,7 +71,7 @@ func (i *arrayFlags) Set(value string) error {
 // The main entry point
 func main() {
 	var modelPlugins arrayFlags
-
+	allowUnvalidatedConfig := flag.Bool("-allowUnvalidatedConfig", false, "allow configuration for devices without a corresponding model plugin")
```
tomikazi: Does making the flag start with "-" make it so that usage has to be "--allowUnvalidatedConfig" rather than "-allowUnvalidateConfig"? Just curious.

SeanCondon: Oops. Copy and paste error. Good catch. The idea is that if nothing is specified, then it will enforce validation against plugins. I hope the choice of name makes this clear

tomikazi: Yes, the name is perfectly clear; I just wasn't entirely sure if the extra "-" had some other meaning to the flags package. :)

PR #902, opened 2019-11-15
Updated admin.proto with docs. Also removed unused RegisterModel Fixes #882

### LINKED ISSUE #882: Improve comments in protobuf files
To generate better docs out of protobuf files, the comments in protobuf files should be improved. For example, some of the fields don't have comments yet and they don't show up in the generated docs.

SeanCondon:
Updated the comments in admin.proto
which regenerated all of the ProtoBufs

I also removed the old `RegisterModel` RPC as this was to add a model plugin from the file system of the onos-config. If it was a new plugin, it would have to be put there by SCP or some kind of mount - anyway not nice. Instead there is the UploadRegisterModel which send the plugin over RPC

I created a CLI implementation for this that streams chunks of file to the server but have not tested it yet.

I skipped the old test for RegisterPlugin for the moment until we can convert it to use the UploadRegisterModel.

A question for the K8S experts - how would this work in a multiple instance installation of onos-config? How could we be sure that the model plugin is registered on to all pods?


adibrastegarnia:
@SeanCondon 
Please take a look at travis. there is a counter which has a linter issuer:
pkg/cli/models.go:151:3: ineffectual assignment to `count` (ineffassign)

		count++

		^

PR #923, opened 2019-11-26
Fix bullets in onos-config deployment doc

adibrastegarnia:


Andrea-Campanella:
this file is going to be scrapped by tomorrow/thrusday so I believe not to be worth the commit.

PR #935, opened 2019-12-03
Initial get on the device and store of config in the onos store

Andrea-Campanella:
This PR allows to get the existing device configuration and store it into ONOS. 
The request is done with and empty path that fetches the whole device configuration in `json` form. 
The` json` is then parsed and stored as initial configuration for that device in the `onos-config` store. 

Currently passes the `SetNetworkConfig` as a function to factory and synchroniser, this avoids circular dependencies in the functional code but provides the same issues on the `test`.


pkg/southbound/synchronizer/synchronizer.go
```diff
@@ -44,6 +46,9 @@ var log = logging.GetLogger("southbound", "synchronizer")
 
 const matchOnIndex = `(\=.*?]).*?`
 
+type SetNetworkConfig func(targetUpdates map[string]devicechange.TypedValueMap,
```
SeanCondon: might need to declare this in some new low level package in `utils`, to make sure it has no dependencies other than `api`

pkg/manager/manager.go
```diff
@@ -145,7 +145,8 @@ func (m *Manager) Run() {
 	go listenOnResponseChannel(m.SouthboundErrorChan, m)
 	//TODO we need to find a way to avoid passing down parameter but at the same time not hve circular dependecy sb-mgr
 	go synchronizer.Factory(m.TopoChannel, m.OperationalStateChannel, m.SouthboundErrorChan,
-		m.Dispatcher, m.ModelRegistry, m.OperationalStateCache, southbound.TargetGenerator, m.OperationalStateCacheLock, m.DeviceChangesStore)
+		m.Dispatcher, m.ModelRegistry, m.OperationalStateCache, southbound.TargetGenerator, m.OperationalStateCacheLock,
+		m.DeviceChangesStore, m.GetTargetConfig)
```
SeanCondon: should this be `m.SetTargetConfig`?

pkg/southbound/synchronizer/synchronizer.go
```diff
@@ -121,6 +130,52 @@ func New(context context.Context,
...
+		Type:     gnmi.GetRequest_CONFIG,
+		Encoding: gnmi.Encoding_JSON,
+	}
+	configResponse, errGet := target.Get(context, getAllRequest)
+	if errGet != nil {
+		log.Error("Can't load configuration on device %s : %s", device.ID, errGet)
+		//TODO propagate
+	}
+	log.Infof("complete response of configurable parameters %s", configResponse)
+	//THis shoudl really be just one, iterating for better safety
```
ray-milkey: Minor typo - "This should"

pkg/southbound/synchronizer/factory.go
```diff
@@ -57,13 +58,19 @@ func Factory(topoChannel <-chan *topodevice.ListResponse, opStateChan chan<- eve
 			} else {
 				mStateGetMode = modelregistry.GetStateMode(mPlugin.GetStateMode())
 			}
+			mReadWritePaths, ok := modelRegistry.ModelReadWritePaths[modelName]
+			if !ok {
+				log.Warnf("Cannot check for read write paths for target %s with %s because "+
```
ray-milkey: Following this code as best as I can to the end, it looks to me like jsonpaths.CorrectJSONPaths will get a nil dereference if mReadWritePaths is nil. I think you have to set it to an empty map here.

Andrea-Campanella:
closing since this will be completly redone after we have a sable environment after merging and testing @kuujo changes.

PR #939, opened 2019-12-04
Removing device simulator helm chart

Andrea-Campanella:
`helm install -n micro-onos new-device deployments/helm/device-simulator/`
then in `onos-cli`
`onos topo add device new-device --address new-device-device-simulator:10161 --type Devicesim --version 1.0.0`


adibrastegarnia:
It would not be better to add them to the simulator  repo? We already have some files there

PR #948, opened 2019-12-06
Fix bugs in controller retries

kuujo:
This PR fixes various bugs in the `NetworkChange` controller to ensure liveness of the protocol for retrying failed `DeviceChange`s.

In order to ensure `DeviceChange`s can safely be retried in an environment where both `NetworkChange`s and the `DeviceChange`s for which they're responsible are being updated by multiple controllers concurrently, an `attempt` field is added to both the `NetworkChange` and `DeviceChange` schemas. The attempt field is used by the network change controller to correlate a network change's state with its device change states. If a device change has failed the current network change attempt, the network change needs to be retried. But if a device change failed a past network change attempt, the device change can be retried. This ensures failures and retries are correctly propagated from the device controller to the network controller and back.

Secondly, the network change controller checks device connectivity states before attempting to apply a change. This will drastically reduce the likelihood of a network change having to be retried due to lack of connectivity to a device. However, because network changes will be blocked until devices become available, we do _need to update the network controller watchers to ensure changes are reconciled once devices are connected_, so this PR is a work in progress.


pkg/controller/change/network/controller.go
```diff
@@ -158,6 +163,19 @@ func (r *Reconciler) createDeviceChanges(networkChange *networkchange.NetworkCha
 
 // canApplyChange returns a bool indicating whether the change can be applied
 func (r *Reconciler) canApplyChange(change *networkchange.NetworkChange) (bool, error) {
+	// First, check if the devices affected by the change are available
+	for _, deviceChange := range change.Changes {
```
SeanCondon: I think this is going to break the ability to have "config only" changes. 

It looks like in that case "device" will not be found and getProtocolState() will return Unknown and the overall result will be false?

PR #951, opened 2019-12-06
Adding multiple targets examples

Andrea-Campanella:


docs/gnmi.md
```diff
@@ -165,6 +165,27 @@ SetRequest() with the 100 extension at the end of the -proto section like:
...
+An example of an attribute on two targets is:
+```bash
+gnmi_cli -address onos-config:5150 -set \
+    -proto "update: <path: <target: 'devicesim-1', elem: <name: 'interfaces'> elem: <name: 'interface' key:<key:'name' value:'eth1' >> elem: <name: 'config'> elem: <name: 'name'>> val: <string_val: 'eth1'>> extension: <registered_ext: <id: 100, msg:'added_devicesim-1-IF'>> update: <path: <target: 'devicesim-2', elem: <name: 'interfaces'> elem: <name: 'interface' key:<key:'name' value:'eth1' >> elem: <name: 'config'> elem: <name: 'name'>> val: <string_val: 'eth1'>> extension: <registered_ext: <id: 100, msg:'2nd_devicesim'>> extension: <registered_ext: <id: 101, msg:'1.0.0'>> extension: <registered_ext: <id: 102, msg:'Devicesim'>>" \
+    -timeout 5s -alsologtostderr \
+    -client_crt /etc/ssl/certs/client1.crt \
+    -client_key /etc/ssl/certs/client1.key \
+    -ca_crt /etc/ssl/certs/onfca.crt
+```
+And example of two attributes on one target and at the same time one on another is:
```
SeanCondon: this is 2 attributes on each device `hold/up` and `hold/down` on both

Also watch out for `And` at start of sentence

PR #953, opened 2019-12-06
Watch device connectivity states to re-queue events when devices connect

kuujo:
This PR improves the network controller's device watcher to ensure changes are re-queued when devices reconnect.

### DIFF
--- a/pkg/controller/change/network/watcher.go
+++ b/pkg/controller/change/network/watcher.go
@@ -22,8 +22,10 @@ import (
 	"github.com/onosproject/onos-config/pkg/controller"
 	devicechangestore "github.com/onosproject/onos-config/pkg/store/change/device"
 	networkchangestore "github.com/onosproject/onos-config/pkg/store/change/network"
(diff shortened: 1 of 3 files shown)


pkg/controller/change/network/watcher.go
```diff
@@ -119,25 +139,40 @@ func (w *DeviceWatcher) Start(ch chan<- types.ID) error {
...
 
+	deviceID := device.NewVersionedID(device.ID(topodevice.ID), device.Version(topodevice.Version))
+	log.Infof("Updating watch for device %v", deviceID)
+
+	// If the protocol state is connected, ensure a stream is open
+	// If the protocol state is not connected, close any stream that's open
 	ctx := w.streams[deviceID]
-	if ctx != nil {
-		log.Errorf("Network DeviceWatcher for device changes %s - ctx already found", deviceID)
+	state := getProtocolState(topodevice)
```
Andrea-Campanella: this does not allow for the watcher to start for a device which is not present in TOPO

PR #955, opened 2019-12-06
Refactor reconciler API to allow controllers to requeue arbitrary changes

kuujo:
This PR refactors the `Reconciler` interface to allow reconcilers to requeue arbitrary changes. This allows changes to dequeued until requeued once another (prior) change has been processed.

Rather than returning a `bool` indicating whether an event was successfully processed, reconcilers return a `Result` which may contain an `ID` to requeue. This allows the `NetworkChange` reconciler to dequeue a change when it is waiting for a prior change. Once the prior change is `COMPLETE`, it requeues the next change to be processed.


kuujo:
This PR is not critical but addresses an efficiency issue. We should leave it open until the integration tests are stable and then revisit.

PR #963, opened 2019-12-10
Add integration test for offline devices

ray-milkey:
The PR adds an integration test for the following scenario:

client requests setting a value on a device that does not exist
simulator is started
test checks that the previously set value gets assigned to the device

The client side currently requires a goroutine because the set() operation on the unknown device hangs. This may not be correct behavior.

### DIFF
--- a/test/gnmi/offlinedevicetest.go
+++ b/test/gnmi/offlinedevicetest.go
@@ -29,7 +29,7 @@ import (
 const (
 	modPath           = "/system/clock/config/timezone-name"
 	modValue          = "Europe/Rome"
-	offlineDeviceName = "offline-dev-1"
+	offlineDeviceName = "TestOfflineDevice-1"
 )
 
 // TestOfflineDevice tests set/query of a single GNMI path to a single device that is initially not in the config
--- a/test/gnmi/offlinedevicetest.go
+++ b/test/gnmi/offlinedevicetest.go
@@ -16,6 +16,7 @@
 package gnmi
 
 import (
+	"github.com/onosproject/onos-config/api/types/change/network"
 	"github.com/onosproject/onos-config/pkg/northbound/gnmi"
 	"github.com/onosproject/onos-test/pkg/onit/env"
 	"github.com/openconfig/gnmi/proto/gnmi_ext"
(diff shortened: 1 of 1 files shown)


test/gnmi/offlinedevicetest.go
```diff
@@ -0,0 +1,94 @@
...
+	go func() {
+		_, extensions, errorSet := gNMISet(MakeContext(), c, setPath, noPaths, extensions)
+		assert.NoError(t, errorSet)
+		assert.Equal(t, 2, len(extensions))
+		extensionBefore := extensions[0].GetRegisteredExt()
+		assert.Equal(t, extensionBefore.Id.String(), strconv.Itoa(gnmi.GnmiExtensionNetwkChangeID))
+	}()
+
+	// Check that the value was set correctly
+	simulator.AddOrDie()
```
SeanCondon: should not be necessary to have a simulator - this could be a completely "config only" device that only ever exists inside onos-config (i.e. not in topo)

Maybe 2 tests are needed here one for "offline device that eventually comes online" and another for "config only device that never has a real equivalent in topo"

test/gnmi/offlinedevicetest.go
```diff
@@ -0,0 +1,94 @@
...
+	time.Sleep(5 * time.Second)
+
+	valueAfter, extensions, errorAfter := gNMIGet(MakeContext(), c, makeDevicePath(offlineDeviceName, modPath))
```
SeanCondon: The Get here would be the same regardless of whether the device is present or not - the values are taken from onos-config - not from the device

ray-milkey: Good point. Andrea suggests interrogating the device as well as the cache here. I can try that. I was hoping to check that the pending value was actually written to the device.

test/gnmi/offlinedevicetest.go
ray-milkey: I think if I add the check of the value on the device mentioned below, it will test both scenarios you mentioned.

ray-milkey:
This works now with @kuujo change to unlink NB and SB

PR #964, opened 2019-12-10
Do not block gNMI sets

kuujo:
Removes event listeners from the gNMI _set_ path.

### DIFF
--- a/pkg/northbound/gnmi/set.go
+++ b/pkg/northbound/gnmi/set.go
@@ -18,17 +18,13 @@ import (
 	"context"
 	"fmt"
 	"github.com/docker/docker/pkg/namesgenerator"
-	changetypes "github.com/onosproject/onos-config/api/types/change"
 	devicechange "github.com/onosproject/onos-config/api/types/change/device"
-	networkchange "github.com/onosproject/onos-config/api/types/change/network"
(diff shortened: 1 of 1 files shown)


pkg/northbound/gnmi/set.go
```diff
@@ -351,52 +359,6 @@ func (s *Server) checkForReadOnly(target string, deviceType devicetype.Type, ver
...
-	}
-	defer ctx.Close()
-	updateResults := make([]*gnmi.UpdateResult, 0)
-	for changeEvent := range networkChan {
-		change := changeEvent.Object.(*networkchange.NetworkChange)
-		log.Infof("Received notification for change ID %s, phase %s, state %s", change.ID,
-			change.Status.Phase, change.Status.State)
-		if change.Status.Phase == changetypes.Phase_CHANGE {
-			switch changeStatus := change.Status.State; changeStatus {
-			case changetypes.State_COMPLETE:
```
Andrea-Campanella: as per our discussion today we need to move this to changetypes.State_PENDING so we know atomix received the change. We can remove line 388 and passthrough with no action on any other events

kuujo:
Opening a new PR for this work as there are issues with this simplistic approach

PR #965, opened 2019-12-10
Avoid hung for not present devices.

Andrea-Campanella:
Removing the device topo presence check in the network change reconciler.

Allows for changes for not yet existing device existing devices to be brought into the system.


Andrea-Campanella:
closing after discussion, this is proper behaviour, just other elements need to be tuned to accomodate the `pending` state for offline/non preset devices

PR #971, opened 2019-12-11
103 extension for gNMI Unavailable devices

### LINKED ISSUE #969: extension 103 should report also unreachable devices
103 currently reports devices that are not known to TOPO.

need to report also unreachable devices as 103.

Andrea-Campanella:
Adding 103 extension if the device is known to topo but gNMI is not available.

Fixes #969

### DIFF
--- a/pkg/northbound/gnmi/set.go
+++ b/pkg/northbound/gnmi/set.go
@@ -150,8 +150,7 @@ func (s *Server) Set(ctx context.Context, req *gnmi.SetRequest) (*gnmi.SetRespon
 		}
 
 		deviceTopo, errDevice := mgr.DeviceStore.Get(topodevice.ID(target))
-		if (errDevice != nil && status.Convert(errDevice).Code() == codes.NotFound) ||
-			(getProtocolState(deviceTopo) == topodevice.ServiceState_UNAVAILABLE) {
+		if errDevice != nil && status.Convert(errDevice).Code() == codes.NotFound {
 			log.Infof("Device is not known to topo %s, %s", target, errDevice)
 			disconnectedDevices = append(disconnectedDevices, target)
 		} else if errDevice != nil {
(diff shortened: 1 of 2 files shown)


pkg/northbound/gnmi/set.go
```diff
@@ -428,3 +434,18 @@ func validateChange(target string, deviceType devicetype.Type, version devicetyp
 	log.Infof("Validating change %s:%s:%s DONE", target, deviceType, version)
 	return nil
 }
+
+func getProtocolState(device *topodevice.Device) topodevice.ServiceState {
```
SeanCondon: I saw something like this in Jordan's code the other day - hope it will be reused

pkg/northbound/gnmi/set.go
```diff
@@ -149,12 +149,18 @@ func (s *Server) Set(ctx context.Context, req *gnmi.SetRequest) (*gnmi.SetRespon
 			Version:  version,
 		}
 
-		_, errDevice := mgr.DeviceStore.Get(topodevice.ID(target))
-		if errDevice != nil && status.Convert(errDevice).Code() == codes.NotFound {
+		deviceTopo, errDevice := mgr.DeviceStore.Get(topodevice.ID(target))
+		if (errDevice != nil && status.Convert(errDevice).Code() == codes.NotFound) ||
+			(getProtocolState(deviceTopo) == topodevice.ServiceState_UNAVAILABLE) {
```
SeanCondon: this seems very closely related to line 160 below - is there an overlap?

pkg/northbound/gnmi/set.go
Andrea-Campanella: that was the channel, this is the service. at the moment I'm figuring let's just put it in then we can refactor.

pkg/northbound/gnmi/set.go
Andrea-Campanella: thanks for spotting

PR #973, opened 2019-12-11
Add integration test for set() is called on a device in the topo but offline

ray-milkey:


test/gnmi/offlineintopotest.go
```diff
@@ -0,0 +1,110 @@
...
+	setPath[0].pathDataType = StringVal
+
+	_, extensionsSet, errorSet := gNMISet(MakeContext(), c, setPath, noPaths, extensions)
+	assert.NoError(t, errorSet)
+	assert.Equal(t, 1, len(extensionsSet))
+	extensionBefore := extensionsSet[0].GetRegisteredExt()
+	assert.Equal(t, extensionBefore.Id.String(), strconv.Itoa(gnmi.GnmiExtensionNetwkChangeID))
+
+	//  Start the simulator
+	simulatorEnv := simulator.AddOrDie()
```
SeanCondon: I think adding the simulator could be done after the Get below

test/gnmi/offlineintopotest.go
```diff
@@ -0,0 +1,110 @@
...
+	time.Sleep(2 * time.Second)
+
+	// Check that the value was set correctly
+	valueAfter, extensions, errorAfter := gNMIGet(MakeContext(), c, makeDevicePath(offlineInTopoModDeviceName, offlineInTopoModPath))
```
SeanCondon: I think this `Get` could be done before the Simulator is started above

The get will read from onos-config only and not from the device.
It would also be good to check the state of the `NetworkChange` at this stage - it should be in `PENDING`

test/gnmi/offlineintopotest.go
```diff
@@ -0,0 +1,110 @@
...
+	assert.NoError(t, errorAfter)
+	assert.Equal(t, 0, len(extensions))
+	assert.NotEqual(t, "", valueAfter, "Query after set returned an error: %s\n", errorAfter)
+	assert.Equal(t, offlineInTopoModValue, valueAfter[0].pathDataValue, "Query after set returned the wrong value: %s\n", valueAfter)
+
+	// interrogate the device to check that the value was set properly
+	deviceGnmiClient := getDeviceGNMIClient(t, simulatorEnv)
+	checkDeviceValue(t, deviceGnmiClient, makeDevicePath(offlineInTopoModDeviceName, offlineInTopoModPath), offlineInTopoModValue)
```
SeanCondon: After this I think you could check the `NetworkChange` again and now its state should be `COMPLETE`

test/gnmi/offlineintopotest.go
```diff
@@ -0,0 +1,110 @@
...
+
+	simulator := env.NewSimulator().SetName(offlineInTopoModDeviceName)
+
+	// Make a GNMI client to use for requests
+	c, err := env.Config().NewGNMIClient()
+	assert.NoError(t, err)
+	assert.True(t, c != nil, "Fetching client returned nil")
+
+	// Set a value using gNMI client to the offline device
+	extNameDeviceType := gnmi_ext.Extension_RegisteredExt{
```
SeanCondon: I don't think you need to give these `DeviceType` and `DeviceVersion` extensions here  - because the Device is in Topo already and the type and version are already known. 

You could add the Network Change Id (100) though here and then you would have a predictable name for the Network change below.

test/gnmi/offlineintopotest.go
```diff
@@ -99,11 +86,35 @@ func (s *TestSuite) TestOfflineDeviceInTopo(t *testing.T) {
...
+	assert.True(t, changeServiceClient != nil)
+	listNetworkChangeRequest := &diags.ListNetworkChangeRequest{
+		Subscribe:     false,
+		ChangeID:      networkChangeID,
+		WithoutReplay: false,
+	}
+	listNetworkChangesClient, listNetworkChangesClientErr := changeServiceClient.ListNetworkChanges(context.Background(), listNetworkChangeRequest)
+	assert.Nil(t, listNetworkChangesClientErr)
+	assert.True(t, listNetworkChangesClient != nil)
+	networkChangeResponse, networkChangeResponseErr := listNetworkChangesClient.Recv()
```
SeanCondon: nice

test/gnmi/offlineintopotest.go
```diff
@@ -0,0 +1,114 @@
...
+	assert.Nil(t, networkChangeResponseErr)
+	assert.True(t, networkChangeResponse != nil)
+	assert.Equal(t, change.State_PENDING, networkChangeResponse.Change.Status.State)
+
+	// Start the device simulator
+	simulator := env.NewSimulator().SetName(offlineInTopoModDeviceName).SetAddDevice(false)
+	simulatorEnv := simulator.AddOrDie()
+	time.Sleep(100 * time.Second)
```
kuujo: This shouldn't be needed

kuujo: Nvm it is needed otherwise waiting for the network change times out. We'll just need a separate PR to address that and remove this sleep.

PR #978, opened 2019-12-12
Test for a set operation on an unreachable device

ray-milkey:
This PR adds an integration test to check that a write to a device that cannot be reached completes properly and sets the cache properly.

### DIFF
--- a/test/gnmi/unreachabledevicetest.go
+++ b/test/gnmi/unreachabledevicetest.go
@@ -1,0 +1,98 @@
+// Copyright 2019-present Open Networking Foundation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package gnmi
+
+import (
+	"context"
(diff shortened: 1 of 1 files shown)


test/gnmi/unreachabledevicetest.go
```diff
@@ -0,0 +1,98 @@
...
+	// Make a GNMI client to use for requests
+	c, err := env.Config().NewGNMIClient()
+	assert.NoError(t, err)
+	assert.True(t, c != nil, "Fetching client returned nil")
+
+	// Set a value using gNMI client to the device
+	extNameDeviceType := gnmi_ext.Extension_RegisteredExt{
+		RegisteredExt: &gnmi_ext.RegisteredExtension{
+			Id:  gnmi.GnmiExtensionDeviceType,
+			Msg: []byte("Devicesim"),
```
SeanCondon: you could reuse `unreachableDeviceModDeviceType` defined at the top

test/gnmi/unreachabledevicetest.go
```diff
@@ -0,0 +1,98 @@
...
+			Msg: []byte(unreachableDeviceModDeviceVersion),
+		},
+	}
+	extensions := []*gnmi_ext.Extension{{Ext: &extNameDeviceType}, {Ext: &extNameDeviceVersion}}
+
+	setPath := makeDevicePath(unreachableDeviceModDeviceName, unreachableDeviceModPath)
+	setPath[0].pathDataValue = unreachableDeviceModValue
+	setPath[0].pathDataType = StringVal
+
+	_, extensionsSet, errorSet := gNMISet(MakeContext(), c, setPath, noPaths, extensions)
```
SeanCondon: As i understand it - this will not return - it should do, but we're waiting on Jordan's patch

ray-milkey: Correct. This cannot go in until the other changes are merged.

ray-milkey:
This one works with @kuujo change to decouple NB and SB

PR #979, opened 2019-12-13
Handling Topo device UPDATE and DELETE

SeanCondon:
As it was onos-config ignored the deletion and update of devices in Topo

* Handle a device DELETE event from Topo
* Handle a device UPDATE event from Topo for **Address** and **Timeout** (others can be added)
* Ignore other UPDATEs
* Treat NONE (got at startup) in the same way as ADDED

### DIFF
--- a/pkg/southbound/synchronizer/factory.go
+++ b/pkg/southbound/synchronizer/factory.go
@@ -16,6 +16,7 @@ package synchronizer
 
 import (
 	"context"
+	"fmt"
 	devicechange "github.com/onosproject/onos-config/api/types/change/device"
 	devicetype "github.com/onosproject/onos-config/api/types/device"
 	"github.com/onosproject/onos-config/pkg/dispatcher"
@@ -32,7 +33,7 @@ import (
 
 var connectionMonitors = make(map[topodevice.ID]*connectionMonitor)
 var connections = make(map[topodevice.ID]bool)
-var commMonitorMu = syncPrimitives.RWMutex{}
+var connMonitorMu = syncPrimitives.RWMutex{}
 
 const backoffInterval = 10 * time.Millisecond
 const maxBackoffTime = 5 * time.Second
@@ -55,8 +56,9 @@ func Factory(topoChannel <-chan *topodevice.ListResponse, opStateChan chan<- eve
 			}
 
 			device := topoEvent.Device
-			commMonitorMu.Lock()
+			connMonitorMu.Lock()
 			connMon, ok := connectionMonitors[device.ID]
+			connMonitorMu.Unlock()
 			if !ok && topoEvent.Type != topodevice.ListResponse_REMOVED {
 				log.Infof("Topo device %s %s", device.ID, topoEvent.Type)
 				connMon = &connectionMonitor{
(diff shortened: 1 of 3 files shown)


pkg/southbound/synchronizer/factory.go
```diff
@@ -70,7 +72,40 @@ func Factory(topoChannel <-chan *topodevice.ListResponse, opStateChan chan<- eve
...
+				changed := false
+				if synchronizer.device.Address != topoEvent.Device.Address {
+					oldAddress := synchronizer.device.Address
+					synchronizer.mu.Lock()
+					synchronizer.device.Address = topoEvent.Device.Address
+					changed = true
+					synchronizer.mu.Unlock()
+					log.Infof("Topo device %s UPDATED address %s -> %s ", device.ID, oldAddress, topoEvent.Device.Address)
+					synchronizer.close()
+					time.Sleep(maxBackoffTime + time.Millisecond*10) // close might not take effect until timeout
```
Andrea-Campanella: is there a way to avoid the sleep here, maybe making the close() method return something and thus making it synchronous ?

pkg/southbound/synchronizer/factory.go
```diff
@@ -70,7 +72,40 @@ func Factory(topoChannel <-chan *topodevice.ListResponse, opStateChan chan<- eve
...
+					synchronizer.reopen()
+					go synchronizer.connect()
+					log.Infof("Topo device %s UPDATED address %s -> %s ", device.ID, oldAddress, topoEvent.Device.Address)
```
Andrea-Campanella: log is repeated

SeanCondon: fixed

pkg/southbound/synchronizer/factory.go
SeanCondon: The reality is that it should block for the longer of either
* the gRPC timeout OR
* the retry value (up to maxBackoffTime)

The reason why that there is currently no way of interrupting Blocking a DialContext() call - as is made through the gNMI client. This means we have to wait until the DialContext() times out in case we are dealing with a non-reachable device.

See the issue #981

pkg/southbound/synchronizer/factory.go
```diff
@@ -68,9 +70,54 @@ func Factory(topoChannel <-chan *topodevice.ListResponse, opStateChan chan<- eve
...
+				waitTime := *connMon.device.GetTimeout() //Use the old timeout in case it has changed
+				if maxBackoffTime > waitTime {
+					waitTime = maxBackoffTime
+				}
+				time.Sleep(waitTime + time.Millisecond*20) // close might not take effect until timeout
+				log.Infof("Topo device %s REMOVED", device.ID)
+			} else {
+				log.Warnf("Unhandled event from topo service %v", topoEvent)
 			}
+			commMonitorMu.Unlock()
```
Andrea-Campanella: maybe we need to discuss this. but we are locking the whole map for the entire duration of the connection process. Is there a way we can only use the the read/write during the intiial reat at line 58  unlocking it immediately then do the operations and then lock, unlock before the update ? because in this way the whole operations are not locked but the r/w is.

PR #987, opened 2019-12-18
Decouple gNMI northbound from southbound

kuujo:
This PR decouples gNMI sets and gets on the northbound from southbound synchronization.

This change modifies the consistency model provided by the northbound API with respect to the propagation of onos-config state to devices. onos-config is the source of truth for the state of the devices below it, and as such once it accepts a change it considers that change to be part of the state of the device whether or not it has yet been propagated. This creates a linearizable consistency model for the state of a device according to onos-config, but a sequential consistency model for the state of a physical device. When a set is received, onos-config waits only long enough for the set to be propagated to its stores. When a get is received, the get is applied to the state of all changes in the store whether applied to the device or not, and this is consistent with the behavior of sets. Once a set returns, a get on the same path will return the set value.


pkg/northbound/gnmi/get.go
```diff
@@ -38,8 +35,6 @@ func (s *Server) Get(ctx context.Context, req *gnmi.GetRequest) (*gnmi.GetRespon
 
 	prefix := req.GetPrefix()
 
-	disconnectedDevicesMap := make(map[topodevice.ID]bool)
```
SeanCondon: lol

PR #990, opened 2019-12-19
Ensure config reconnects to devices on disconnection

kuujo:
Fixes a bug that prevents config from reconnecting to devices once disconnected.

### DIFF
--- a/pkg/southbound/synchronizer/factory.go
+++ b/pkg/southbound/synchronizer/factory.go
@@ -155,11 +155,18 @@ type connectionMonitor struct {
 	deviceChangeStore         device.Store
 	device                    *topodevice.Device
 	target                    southbound.TargetIf
+	cancel                    context.CancelFunc
 	closed                    bool
 	mu                        syncPrimitives.RWMutex
 }
 
 func (cm *connectionMonitor) reconnect() {
+	cm.mu.Lock()
+	if cm.cancel != nil {
+		cm.cancel()
+		cm.cancel = nil
+	}
+	cm.mu.Unlock()
 	cm.operationalStateCacheLock.Lock()
 	delete(cm.operationalStateCache, cm.device.ID)
 	cm.operationalStateCacheLock.Unlock()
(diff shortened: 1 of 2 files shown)


pkg/southbound/synchronizer/factory.go
```diff
@@ -56,9 +55,7 @@ func Factory(topoChannel <-chan *topodevice.ListResponse, opStateChan chan<- eve
 			}
 
 			device := topoEvent.Device
-			connMonitorMu.Lock()
 			connMon, ok := connectionMonitors[device.ID]
```
SeanCondon: I know this should be a read lock, but why doesn't this need a mutex? - i'd expect this to give a data race error

pkg/southbound/synchronizer/factory.go
```diff
@@ -219,8 +232,6 @@ func (cm *connectionMonitor) synchronize() error {
 		cm.operationalStateCacheLock.Lock()
 		delete(cm.operationalStateCache, cm.device.ID)
 		cm.operationalStateCacheLock.Unlock()
-		cm.southboundErrorChan <- events.NewErrorEventNoChangeID(events.EventTypeErrorDeviceConnect,
```
SeanCondon: good riddance

pkg/southbound/synchronizer/factory.go
```diff
@@ -137,6 +130,26 @@ func Factory(topoChannel <-chan *topodevice.ListResponse, opStateChan chan<- eve
 			log.Infof("Received event %v", event)
 			deviceID := topodevice.ID(event.Subject())
 			switch event.EventType() {
+			case events.EventTypeErrorDeviceConnect:
+				deviceID := topodevice.ID(event.Subject())
+				synchronizer, ok := connectionMonitors[deviceID]
+				if ok && connections[deviceID] {
+					synchronizer.close()
+					synchronizer = &connectionMonitor{
```
SeanCondon: i can't figure out why this needs to be recreated

pkg/southbound/synchronizer/factory.go
kuujo: The `connectionMonitors` map is only written and read in a loop from that one goroutine, so no lock is necessary

pkg/southbound/synchronizer/factory.go
kuujo: Hmm... actually I realized this code was correct. The reason is because the factory loop maintains the `connections` map, and a synchronizer is only recreated if it was already connected before receiving a disconnect event. But I think we probably have to keep the `connections` map to avoid reconnecting multiple times when concurrent disconnection events occur.

PR #1021, opened 2020-01-09
Avoid opening new streams for each NetworkChange watch

kuujo:
This PR refactors the `NetworkChange` store to avoid opening new streams for each watch of a specific change ID. This significantly reduces the amount of interaction the northbound API has to do with stores during a write on the northbound. To ensure events are not lost since they're not replayed, this PR reorders the gNMI set logic to register the watch before adding the change.


adibrastegarnia:
@kuujo  closing this PR since it is not relevant at the moment

PR #1032, opened 2020-01-13
Made snapshots CLI command consistent with other commands

SeanCondon:
now is more like `.. get network-changes`

`onos config get snapshots` to get all snapshots - printing all fields
`onos config get snapshots -v` to get all snapshots - printing all fields and showing Path-Values
`onos config get snapshots <wildcard>` to get all snapshots - for device names, using wildcards for matching  like * and ?
`onos config watch snapshots` to watch all snapshots - replay all present and wait for updated and new ones


SeanCondon:
I need to update this after rebasing, so don't merge yet

PR #1034, opened 2020-01-14
Changed List Response type in to an enum, for use in NBI, like topo

SeanCondon:
For the diags NBI `ListNetworkChangeResponse` and `ListDeviceChangeResponse` I'm adding an enumerated **ListResponseType** enum in `types/change/types.proto` with values:
* NONE
* ADDED
* UPDATED
* REMOVED

as these need to be exposed to clients e.g. GUI

This is essentially what `stream.EventType` was doing. No point in repeating it, so I've made the Proto definition the canonical definition, and replaced `stream.EventType`


pkg/store/stream/stream.go
```diff
@@ -48,24 +51,10 @@ func (c *closerContext) Close() {
...
-	// Created indicates an object was created
-	Created EventType = "Created"
-	// Updated indicates an object was updated
-	Updated EventType = "Updated"
-	// Deleted indicates an object was deleted
-	Deleted EventType = "Deleted"
-)
-
 // Event is a stream event
 type Event struct {
```
kuujo: This type was meant to be a generic type for store events. I think it's fine to couple this to changes, but maybe we should make the value type safe as well and create separate events for network and device change stores.

kuujo: I guess the way to make this consistent with the other interfaces would be to have a protobuf message for `NetworkChange` list responses and `DeviceChange` list responses and use those instead of this `Event` in the store APIs.

SeanCondon: Thanks for looking at it, and the reply - if I went down the road of using separate types for NetworkChange and for DeviceChange, then they would not be reusable here in `stream.go` because it is used for both types.

I can do it straight like Topo and embed the enum in to the proto message. Will push another patch

PR #1048, opened 2020-01-20
Add Junos 19.3R1 YANG and generated .go files

k1eran:
Initial commit with generated code 
Later merge requests will include any necessary hand-modifications or UT updates


Andrea-Campanella:
@SeanCondon I'm wondering if we ought to have these model plugins in a separate repo I'm in favour of not weighing down the onos-config repo with code that is meant to be build and loaded externally. it could be the case with the openconfig and stratum plugins too.

SeanCondon:
I'd agree with @Andrea-Campanella it would be better not to weigh down onos-config with YANG files like this. I think it would really have the potential to slow down the build and I found that linting took over 10 minutes when I had the resulting Go code in my path.

I think we need another repo for models like this, maybe a repo per device might be too heavy.

In any case when it comes to deployment, I think the important thing is that we would not expect to bake the plugins in to the Docker image - there are a couple of things that come to mind:
1) the plugin could become part of a Config Map that's loaded in to K8s or
2) the plugins could become their own pod, just holding the files which could be selectively mounted as needed

Maybe a discussion is needed on TST

Andrea-Campanella:
@SeanCondon +1 to discussing this to TST. If you can please do prepare some slides. In Helm we can provide a fully qualified location of the plugins in the values.yaml file.

k1eran:
@SeanCondon @Andrea-Campanella Regarding the refactor of some (all?) plugins out of onos-config,  is that something that can happen after we get the new plugin merged in or does that need done to before hand with e.g. Stratum moved over as an example ?

Andrea-Campanella:
@k1eran let's discuss on TST tomorrow. I'd like for this to happen sooner rather than later, but it depends on how much @SeanCondon has on his plate. He si the one that needs to take care of this because he knows the plugins inside out.

k1eran:
> @k1eran let's discuss on TST tomorrow.

Sounds good @Andrea-Campanella @SeanCondon 
Regarding the `golangci-lint` execution in onos-config project: 
I see currently the same checks are applied to the “hand-written” code as to the ygot-generated code. With the auto-generated code  “growing“ as in the pull request we see` golangci-lint` giving out-of-mem errs on travis, though it’s  fine on my  laptop, albeit slow the first time it executes. I looked at a couple of things in that area:
1.   Updated to more recent `golangci-lint`  to test on Travis but unfortunately it still fails -recent  improvements  mentioned at https://github.com/golangci/golangci-lint/issues/337  were not enough to get the build to pass.
2. Experimented with disabling linting for the autogenerated code  
    - Via the skip-dirs attribute but as explained at https://raw.githubusercontent.com/golangci/golangci-lint/master/.golangci.example.yml   _“they will be analyzed, but issues from them won't be reported.”_ In other words  it doesn't help re. high memory usage ;  it just reduces final reporting.
    - Looked at extending use of  `// +build modelplugin` constraint to the `generated.go` but some of of them (not Stratum?) are used in UTs  so introducing build constraints had  ripple effects.

Splitting off the generated models would I assume make it easier to have a more lightweight (e.g. zero?)  lint execution for autogenerated code which seems like a good thing ?

Andrea-Campanella:
@k1eran you are perfectly right in that assumption. Splitting off the code would enable not to have these issue of listing enormous portions of code that is autogenerated while keeping a tight leash on the hand-written parts in `config`. This is really another argument in favour of the split. Let's discuss later today. I think @SeanCondon has put some slides together to prime the discussion.

PR #1057, opened 2020-01-23
Feature/build junos plugin

k1eran:
Building a  new model plugin for Junos. It is not added to the Helm Chart and not loaded **by default** into onos-config.

Note: As discussed at 2020/01/22 TST meeting, the plan is to move the YANG, generated code & build for plugins  (Including this one) out of onos-config repo to new location in the future.


modelplugin/Junos-19.3R1.8/modelmain.go
```diff
@@ -0,0 +1,74 @@
...
+)
+
+type modelplugin string
+
+const modeltype = "Junos"
+const modelversion = "19.3R1.8"
+const modulename = "junos.so.19.3R1.8"
+
+var modelData = []*gnmi.ModelData{
+    {Name:"junos-conf-interfaces",Organization:"Juniper",Version:"2019-01-01"},
```
SeanCondon: shouldn't `junos-conf-system` be in here too - the same 2 as are listed in the ".env" file

modelplugin/Junos-19.3R1.8/modelmain.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/openconfig/gnmi/proto/gnmi"
+	"github.com/openconfig/goyang/pkg/yang"
+	"github.com/openconfig/ygot/ygot"
+)
+
+type modelplugin string
+
+const modeltype = "Junos"
+const modelversion = "19.3R1.8"
+const modulename = "junos.so.19.3R1.8"
```
SeanCondon: I'm not sure if a name like this is valid for a .so file - with a letter in it - I hope this won't break something down the line

Makefile
```diff
@@ -48,6 +48,12 @@ build/_output/stratum.so.1.0.0: modelplugin/Stratum-1.0.0/modelmain.go modelplug
 build/_output/stratum-debug.so.1.0.0: modelplugin/Stratum-1.0.0/modelmain.go modelplugin/Stratum-1.0.0/stratum_1_0_0/generated.go
 	CGO_ENABLED=1 go build -o build/_output/debug/stratum-debug.so.1.0.0 -gcflags "all=-N -l" -buildmode=plugin -tags=modelplugin ./modelplugin/Stratum-1.0.0
 
+build/_output/junos.so.19.3R1.8: modelplugin/Junos-19.3R1.8/modelmain.go modelplugin/Junos-19.3R1.8/junos_19_3R1_8/generated.go
+	CGO_ENABLED=1 go build -o build/_output/junos.so.19.3R1.8 -buildmode=plugin -tags=modelplugin ./modelplugin/Junos-19.3R1.8
+
+build/_output/junos-debug.so.19.3R1.8: modelplugin/junos-19.3R1.8/modelmain.go modelplugin/junos-19.3R1.8/junos_19_3R1_8/generated.go
```
SeanCondon: I think you can leave out the "debug" target - we used to use a debug version of onos-config at one stage, but now I don't think we can use it at all.

modelplugin/Junos-19.3R1.8/modelmain.go
k1eran: >I’m not sure if a name like this is valid for a .so file - with a letter in it - I hope this won’t break something down the line

@SeanCondon  I have come round to thinking we shouldn’t be using the (alphanumeric) YANG / device version as the .so version, and standard lib versioning approach mentioned at https://stackoverflow.com/questions/663209/can-someone-explain-about-linux-library-naming is needed.

Potentially we could need to track multiple plugins that match same version of the YANG  
e.g. if we find a bug in the generated go-code, and then need to fix code-generator tool then recreate the lib, we'll need a new version number for that .so file.

One simple way would be just incorporating the YANG version in the TYPE. e.g.
```
modelplugin/Junos-19.3R1.8.env:TYPE=Junos19_3R1_8
modelplugin/Junos-19.3R1.8.env:VERSION=1.0.0 
```
//Note: I’m deliberately avoiding dots and dashes in the TYPE to avoid compilation issues re. Package name.

 
This results in the  `modelplugin/Junos19_3R1_8-1.0.0/modelmain.go`
```
30:const modeltype = “Junos19_3R1_8"
31:const modelversion = “1.0.0”
32:const modulename = “junos19_3r1_8.so.1.0.0"
```

This results in these shared libs …
```
build/_output
├── devicesim.so.1.0.0
├── junos19_3r1_8.so.1.0.0
├── stratum.so.1.0.0
├── testdevice.so.1.0.0
└── testdevice.so.2.0.0
```

Later we can decide if the other four existing .so files need same update. WDYT ?

Makefile
k1eran: @SeanCondon ok, I can either remove it in this request or do a follow-up separately  to remove all 5 `*debug.so*` files in a single go (which would avoid having inconsistencies across the plugins for a while). WDYT ?

SeanCondon: I'd say just remove this JunOS one - best not to add it in. I will remove the other 4 when I move to own repo for plugins

modelplugin/Junos-19.3R1.8/modelmain.go
SeanCondon: I was thinking that the version number was a version of the **collection** of YANGs for a particular device. So I think junos.so.<yangs version> is fine, as long as the <yangs version> only has digits in it. So maybe junos.so.19.3.18 or junos.so.19.3.1.8 would work for me.

The way I had intended the versions to work was that the type would describe a family of versions. 

Re your comment about the bug in the generated go code, I think the versioning system of docker is there to take care of that scenario, as in future each plugin will be its own Docker image - so will be like 
`onosproject/junos_19.3R1_8:latest`

I know I was a bit undecided on the TST call last week, but since then I have figured out that I will create each plugin as a Docker image, and as many plugins as are wanted will be loaded alongside onos-config container in a "sidecar" fashion. This would mean that when I move to this pattern, the Helm chart would define and pull in the plugins needed, and they would be available in the same Pod because they are sidecars.

modelplugin/Junos-19.3R1.8/modelmain.go
k1eran: thanks - yes  there was a missing `;` in the  ".env" file

k1eran:
Parking this pull request until a new mechanism for handling plugins is in place  e.g. potentially  a `onosproject/config-models` repo,  so that there is no need to store temporary build artefacts like generated.go  for any plugin (including this one)

PR #1071, opened 2020-02-05
Test subscribe to an opstate cache path

ray-milkey:
This PR introduces two new subscribe tests to check on opstate cache paths that the device is updating. One used the diags API, one uses a straight GNMI call. This also refactors the existing subscribe tests to use utility routines to cut down on duplicate code.

### DIFF
--- a/test/gnmi/subscribetest.go
+++ b/test/gnmi/subscribetest.go
@@ -1,4 +1,4 @@
-// Copyright 2019-present Open Networking Foundation.
+// Copyright 2020-present Open Networking Foundation.
 //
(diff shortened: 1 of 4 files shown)


test/gnmi/subscribetest.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2019-present Open Networking Foundation.
+// Copyright 2020-present Open Networking Foundation.
```
SeanCondon: probably shouldn't update if the file existed in 2019

ray-milkey: Good catch. Not sure why I did that...

PR #1082, opened 2020-02-14
Bump go compiler version to 1.13

ray-milkey:


go.mod
```diff
@@ -9,14 +9,10 @@ require (
 	github.com/atomix/go-local v0.0.0-20200211010611-c99e53e4c653
 	github.com/cenkalti/backoff v2.2.1+incompatible
 	github.com/docker/docker v1.13.1
-	github.com/go-logfmt/logfmt v0.4.0 // indirect
```
SeanCondon: I don't know why these would disappear now - I'd say ignore this go.mod in this commit

SeanCondon: actually if Travis is happy I'm happy

PR #1103, opened 2020-03-13
Migrate portions of tests and benchmarks to Helm-based API

kuujo:
This PR is the first step to migrating the tests to use the charts using the Helm-based test API. All deployments are migrated, but the connection helpers still need to be replaced. Just putting it up here to share the new API for feedback.


kuujo:
Still working on adding all the APIs we use

test/cli/suite.go
```diff
@@ -15,53 +15,60 @@
...
 
-// SetupTestSuite sets up the onos-config CLI test suite
-func (s *HATestSuite) SetupTestSuite() {
-	setup.Atomix()
-	setup.Database().Raft()
-	setup.Topo().SetReplicas(2)
-	setup.Config().SetReplicas(2)
-	setup.SetupOrDie()
+	// Install the onos-topo chart
+	err = helm.Helm().
```
SeanCondon: how come there's no namespace here?

build/onos-config-tests/Dockerfile
```diff
@@ -1,9 +1,12 @@
 FROM alpine:3.8
```
SeanCondon: should go for 3.9

PR #1127, opened 2020-04-22
Update logging configuration based on new changes

adibrastegarnia:


pkg/config/config.go
```diff
@@ -26,8 +25,6 @@ var config *Config
 type Config struct {
 	// Atomix is the Atomix configuration
 	Atomix atomix.Config `yaml:"atomix,omitempty"`
-	// Logging is the logging configuration
-	Logging logging.Config `yaml:"logging,omitempty"`
```
SeanCondon, line 30: you were getting a config from yaml in here and now the lib is getting it?

adibrastegarnia: Yeah. Originally, it was part of subsystem configuration but we found out loggers will be initialized (when go imports library packages) before we have access to the config file in the main or even in init. Logging package has an init function and if a config file is available for logging will read it and create loggers otherwise the loggers defined in the system will be defined as it is.

PR #1137, opened 2020-05-14
CLI. Added a facility to load configuration from a YAML file in bulk

SeanCondon:


adibrastegarnia:
@SeanCondon 
nice feature. I think the same sort of representation will be needed for configuration (i.e. gnmi) in ztp side (something to think about)

PR #1158, opened 2020-07-30
Use new onos-topo entity-relationship api

shadansari:


pkg/controller/change/network/watcher_test.go
```diff
@@ -154,51 +156,74 @@ func TestDeviceWatcher(t *testing.T) {
...
+	deviceClient.EXPECT().Subscribe(gomock.Any(), gomock.Any()).DoAndReturn(func(ctx context.Context, request *topo.SubscribeRequest) (topo.Topo_SubscribeClient, error) {
+		listClient := NewMockTopo_SubscribeClient(ctrl)
+		listClient.EXPECT().Recv().Return(&topo.SubscribeResponse{
+			Update: &topo.Update{
+				Type: topo.Update_UNSPECIFIED,
+				Object: &topo.Object{
+					ID:   topo.ID("device-1"),
+					Type: topo.Object_ENTITY,
+					Attributes: map[string]string{
+						topo.Type:    "DeviceSim",
```
SeanCondon, line 169: I think the type should come from the "kind"

pkg/manager/manager.go
```diff
@@ -240,10 +241,11 @@ func (m *Manager) CheckCacheForDevice(target devicetype.ID, deviceType devicetyp
 	version devicetype.Version) (devicetype.Type, devicetype.Version, error) {
 
 	deviceInfos := mgr.DeviceCache.GetDevicesByID(target)
-	topoDevice, errTopoDevice := mgr.DeviceStore.Get(topodevice.ID(target))
+	object, errTopoDevice := mgr.DeviceStore.Get(topo.ID(target))
 	if errTopoDevice != nil {
 		log.Infof("Device %s not found in topo store", target)
```
SeanCondon, line 246: probably should throw error here - no point continuing without `object`

pkg/manager/devicehandler.go
```diff
@@ -40,12 +41,12 @@ func (m *Manager) DeviceDisconnected(id topodevice.ID, err error) error {
...
 
-func updateDevice(deviceStore device.Store, id topodevice.ID, connectivity topodevice.ConnectivityState, channel topodevice.ChannelState,
-	service topodevice.ServiceState) error {
-	topoDevice, err := deviceStore.Get(id)
+func updateDevice(deviceStore device.Store, id topodevice.ID, connectivity topodevice.ConnectivityState, channel topodevice.ChannelState, service topodevice.ServiceState) error {
+	object, err := deviceStore.Get(topo.ID(id))
 	if err != nil {
 		return err
 	}
+	topoDevice := topo.ObjectToDevice(object)
```
SeanCondon, line 49: I think it is a bit of a concern that `ObjectToDevice` does not give any error - there could be may things wrong with the object - e.g. no address, unknown kind, no version - because we're running off an attribute map, there are many possibilities

pkg/manager/manager.go
shadansari: done

pkg/controller/change/network/watcher_test.go
shadansari: done

pkg/manager/devicehandler.go
shadansari: @SeanCondon - I plan to move ObjectToDevice and DeviceToObject functions from onos-topo to onos-config (since Device is only used by onos-config now). Will add validation as part of that change. Currently, onos-topo allows creation of "anonymous" entities (i.e. entities with no Kind). Need to add that check in onos-topo as well.

shadansari: https://github.com/onosproject/onos-config/issues/1163

SeanCondon: It would be worth investigating if `Device` can be removed from `onos-config` altogether - it could be replaced globally with `Entity`

PR #1168, opened 2020-08-07
Update integration tests badge

adibrastegarnia:


adibrastegarnia:
@ray-milkey 
I was wondering why it is always green :D

PR #1172, opened 2020-08-13
Fix update device function 

adibrastegarnia:
This PR fixes the issue that exists in updateDevice function. The problem is that updateDevice function was retrying to update the state of a device that has been removed already and it is not in the device store anymore.

### DIFF
--- a/pkg/manager/devicehandler.go
+++ b/pkg/manager/devicehandler.go
@@ -18,6 +18,8 @@ import (
 	"github.com/cenkalti/backoff"
 	"github.com/onosproject/onos-config/pkg/store/device"
 	topodevice "github.com/onosproject/onos-topo/api/device"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 )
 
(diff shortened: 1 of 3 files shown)


adibrastegarnia:
@ray-milkey 
That is great. Try it a few times and see how it works. I returned the TestSubscribe back and it is reliable now.

pkg/manager/devicehandler.go
```diff
@@ -43,9 +45,18 @@ func (m *Manager) DeviceDisconnected(id topodevice.ID, err error) error {
 func updateDevice(deviceStore device.Store, id topodevice.ID, connectivity topodevice.ConnectivityState, channel topodevice.ChannelState,
 	service topodevice.ServiceState) error {
 	topoDevice, err := deviceStore.Get(id)
+
+	st, ok := status.FromError(err)
+
+	// If the device doesn't exist then we should not update its state
+	if ok && err != nil && st.Code() == codes.NotFound {
```
adibrastegarnia, line 52: We can remove err != nil but it is OK

PR #1174, opened 2020-08-14
Fix offline device integration test

adibrastegarnia:
This PR fixes offline device integration test. The problem was the client for the device was not be able to connect target because of TLS handshake error. I don't think we support that in the simulator.

### DIFF
--- a/test/utils/gnmi/utils.go
+++ b/test/utils/gnmi/utils.go
@@ -44,9 +44,7 @@ import (
 	"github.com/openconfig/gnmi/proto/gnmi_ext"
 	"github.com/stretchr/testify/assert"
 	"google.golang.org/grpc"
-	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/credentials"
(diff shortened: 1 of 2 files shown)


test/utils/gnmi/utils.go
```diff
@@ -464,19 +457,16 @@ func GetDeviceGNMIClientOrFail(t *testing.T, simulator *helm.HelmRelease) client
 	services, err := simulatorClient.CoreV1().Services().List()
 	assert.NoError(t, err)
 	service := services[0]
-	conn, err := connectService(simulator)
 	assert.NoError(t, err)
 	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
 	defer cancel()
-	creds, err := getClientCredentials()
 	assert.NoError(t, err)
```
ray-milkey, line 463: Can remove this line now too

adibrastegarnia: Done. Good catch

Issue #962, opened 2019-12-09
Update device connectivity state on master only


kuujo:
Any node can currently update connectivity info for a device. This means if a device disconnects from one node, connectivity is recorded as down for all nodes regardless of whether the master can still reach the device. Nodes should only attempt to become the master for a device once they’ve established a connection to it, and only the master should record the connectivity state for a device. This ensures mastership can only fail over to a node that can reach the device, and nodes that can’t reach the device cannot overwrite the connectivity state of nodes that can.

adibrastegarnia:
This issue is also addressed.

PR #1176, opened 2020-08-19
Add GetMastership function to mastership store

adibrastegarnia:
This PR adds a function to mastership store that we can get mastership term based on a given device id. 
Part of this issue https://github.com/onosproject/onos-config/issues/962

### DIFF
--- a/pkg/controller/filter.go
+++ b/pkg/controller/filter.go
@@ -38,7 +38,24 @@ type Filter interface {
 type MastershipFilter struct {
 	Store    mastershipstore.Store
 	Resolver DeviceResolver
-	NodeID   cluster.NodeID
+	nodeID   cluster.NodeID
+}
+
+// NewMastershipFilter creates an instance of mastershipFilter
+func NewMastershipFilter(store mastershipstore.Store, resolver DeviceResolver, nodeID cluster.NodeID) *MastershipFilter {
+	if string(nodeID) == "" {
+		nodeID = cluster.GetNodeID()
+	}
+	return &MastershipFilter{
+		Store:    store,
+		Resolver: resolver,
+		nodeID:   nodeID,
+	}
+}
+
+// GetNodeID returns node id
+func (f *MastershipFilter) GetNodeID() cluster.NodeID {
+	return f.nodeID
 }
 
 // Accept accepts the given ID if the local node is the master
@@ -53,7 +70,7 @@ func (f *MastershipFilter) Accept(id types.ID) bool {
 	}
 
 	// checks whether the local node is the master
-	if master == nil || string(master.Master) != string(f.NodeID) {
+	if master == nil || string(master.Master) != string(f.GetNodeID()) {
 		return false
 	}
 
--- a/pkg/controller/filter.go
+++ b/pkg/controller/filter.go
@@ -17,6 +17,8 @@ package controller
 import (
 	"regexp"
 
+	"github.com/onosproject/onos-config/pkg/store/cluster"
+
 	"github.com/onosproject/onos-config/api/types"
 	mastershipstore "github.com/onosproject/onos-config/pkg/store/mastership"
 	topodevice "github.com/onosproject/onos-topo/api/device"
@@ -45,11 +47,15 @@ func (f *MastershipFilter) Accept(id types.ID) bool {
 		return false
 	}
 	master, err := f.Store.GetMastership(device)
-	if err == nil && master != nil {
-		return true
+	if err != nil {
+		return false
+	}
+	// checks whether the local node is the master
+	if master == nil || string(master.Master) != string(cluster.GetNodeID()) {
+		return false
 	}
 
-	return false
+	return true
 }
 
 var _ Filter = &MastershipFilter{}
--- a/pkg/controller/filter.go
+++ b/pkg/controller/filter.go
@@ -41,21 +41,13 @@ type MastershipFilter struct {
 	nodeID   cluster.NodeID
 }
 
-// NewMastershipFilter creates an instance of mastershipFilter
-func NewMastershipFilter(store mastershipstore.Store, resolver DeviceResolver, nodeID cluster.NodeID) *MastershipFilter {
-	if string(nodeID) == "" {
-		nodeID = cluster.GetNodeID()
-	}
-	return &MastershipFilter{
-		Store:    store,
-		Resolver: resolver,
-		nodeID:   nodeID,
-	}
-}
-
 // GetNodeID returns node id
-func (f *MastershipFilter) GetNodeID() cluster.NodeID {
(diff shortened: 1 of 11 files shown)


pkg/store/mastership/store.go
```diff
@@ -126,6 +131,15 @@ func (s *atomixStore) NodeID() cluster.NodeID {
 	return s.nodeID
 }
 
+func (s *atomixStore) GetTerm(deviceID topodevice.ID) (Term, error) {
+	if election, found := s.elections[deviceID]; found {
+		term := election.getTerm()
+		return term, nil
+	}
+
+	return 0, fmt.Errorf("device mastership does not exist for device %s", deviceID)
```
adibrastegarnia, line 140: @kuujo 
Can we assume the term starts from one and zero represents nothing or its initial value ?

kuujo: The term for a real master is always `> 0`. The term will only be `0` if no master has ever been elected.

adibrastegarnia: So it makes sense then.

pkg/store/mastership/store.go
```diff
@@ -37,6 +39,9 @@ type Store interface {
 	// IsMaster returns a boolean indicating whether the local node is the master for the given device
 	IsMaster(id topodevice.ID) (bool, error)
 
+	// GetTerm returns the mastership term for a given device
+	GetTerm(id topodevice.ID) (Term, error)
```
kuujo, line 43: We might as well make this `GetMastership` instead of `GetTerm` so all the relevant information is returned in one step. Otherwise, mastership can change between `IsMaster` and `GetTerm`

adibrastegarnia: Yeah. I was thinking about it... IsMaster function performs election as well which is a bit confusing. Suppose we have device1 and device2. If we call IsMaster for device1 then a master will be selected if it doesn't exist. So GetMastership performs election and returns all related information including Term assuming if a master doesn't exist for a device it picks one.

pkg/store/mastership/election.go
```diff
@@ -168,13 +169,13 @@ func (e *atomixDeviceMastershipElection) watchElection(ch <-chan *election.Event
 	}
 }
 
-func (e *atomixDeviceMastershipElection) isMaster() (bool, error) {
+func (e *atomixDeviceMastershipElection) getMastership() *Mastership {
 	e.mu.RLock()
 	defer e.mu.RUnlock()
 	if e.mastership == nil || string(e.mastership.Master) != e.election.ID() {
```
adibrastegarnia, line 175: I assume we still need to keep this condition

pkg/controller/filter.go
```diff
@@ -43,11 +44,12 @@ func (f *MastershipFilter) Accept(id types.ID) bool {
 	if err != nil {
 		return false
 	}
-	master, err := f.Store.IsMaster(device)
-	if err != nil {
-		return false
+	master, err := f.Store.GetMastership(device)
+	if err == nil && master != nil {
```
adibrastegarnia, line 48: @kuujo 
Please make sure it makes sense to you.

adibrastegarnia:
@kuujo 
I think I addressed your comments and see if you like the new changes. Please let me know if you have any more comments. I completely removed isMaster function and updated the places that is called instead of labeling it as Deprecated for now.

pkg/controller/filter.go
kuujo: This should be checking whether the local node is the master. This check merely checks whether some master exists.

adibrastegarnia:
I should test it a bit more to make sure tests are passing without any problems.

pkg/store/mastership/election.go
kuujo: This condition should not be here. We're returning the mastership regardless of whether the local node is the master, but this check only returns the mastership if the local node is the master. Other nodes should have access to mastership info even if they're not the master.

adibrastegarnia: Makes sense. So it should be moved to where it needs to be checked based on another comment.

kuujo: Yeah... the local node ID can be read from the mastership store interface to check it.

pkg/controller/snapshot/device/controller.go
```diff
@@ -40,6 +42,7 @@ func NewController(mastership mastershipstore.Store, changes changestore.Store,
 		Resolver: &Resolver{
 			snapshots: snapshots,
 		},
+		NodeID: cluster.GetNodeID(),
```
adibrastegarnia, line 45: NodeID injected that can be set by tests and here we use cluster.GetNodeID to get its value.

adibrastegarnia:
@kuujo 
I think it should be OK now.

pkg/controller/filter.go
```diff
@@ -35,6 +38,7 @@ type Filter interface {
 type MastershipFilter struct {
 	Store    mastershipstore.Store
 	Resolver DeviceResolver
+	NodeID   cluster.NodeID
```
kuujo, line 41: We should probably make this private and add a `New` function to force it to be set correctly. Feels wrong to force other packages to set the node ID here.

adibrastegarnia: @kuujo 
Agreed. Let me fix it.

adibrastegarnia: I added a NewMastershipFilter function which accepts three arguments. The last argument is nodeID which can be empty and when it is empty it will be set to cluster NodeID otherwise it will be set to whatever it is passed to it (like what we have in filter_test.go). I also added a GetNodeID to filter that I can retrieve the private field in Accept function for checking local cluster node.

adibrastegarnia:
@kuujo 
I used POD_NAME for retrieving nodeID for now until we use cluster package in onos-lib-go for stores.  Perhaps I should do that one first before using mastership info for southbound. Looks like multiple store need to be updated to use that information accordingly.

pkg/controller/filter.go
```diff
@@ -35,6 +38,24 @@ type Filter interface {
 type MastershipFilter struct {
 	Store    mastershipstore.Store
 	Resolver DeviceResolver
+	nodeID   cluster.NodeID
+}
+
+// NewMastershipFilter creates an instance of mastershipFilter
+func NewMastershipFilter(store mastershipstore.Store, resolver DeviceResolver, nodeID cluster.NodeID) *MastershipFilter {
```
kuujo, line 45: Actually, I think we should not have a `New*` function for the mastership filter. We should keep the API the same, but use `cluster.GetNodeID()` if the private `nodeID` field is empty.

adibrastegarnia: @kuujo 
Done.

Issue #960, opened 2019-12-09
Add concurrency control for device states in multi-node clusters


kuujo:
In multi-node clusters, device mastership changes can cause multiple nodes to believe themselves to be the master for a device at the same time, potentially resulting in erratic updates to device protocol states. And although onos-config may primarily be run in a single node configuration, multi-node clusters cannot be entirely avoided. Support for rolling updates will make this issue particularly relevant.

Config should use device attributes to implement concurrency control for device mastership. To do so, we simply need to include device mastership terms in device protocol state updates and relinquish mastership when a greater term is encountered.

adibrastegarnia:
This issue is addressed for now until we migrate to use new topo API and perhaps use entity relationship to implement mastership .

PR #1182, opened 2020-08-24
New cluster package for onos-config stores

adibrastegarnia:
This PR a work in progress to migrate onos-config to use the new cluster package in onos-lib-go
issue https://github.com/onosproject/onos-config/issues/960


pkg/store/snapshot/network/store.go
```diff
@@ -17,28 +17,30 @@ package network
...
-	"io"
-	"time"
+	"github.com/onosproject/onos-lib-go/pkg/cluster"
 )
 
 const snapshotsName = "network-snapshots"
 
-func init() {
+/*func init() {
 	uuid.SetNodeID([]byte(cluster.GetNodeID()))
```
adibrastegarnia, line 38: Is it safe to initialize this when we create the store?

pkg/store/change/network/store.go
```diff
@@ -17,28 +17,26 @@ package network
...
 const changesName = "network-changes"
 
-func init() {
-	uuid.SetNodeID([]byte(cluster.GetNodeID()))
-}
-
 // NewAtomixStore returns a new persistent Store
-func NewAtomixStore(config config.Config) (Store, error) {
+func NewAtomixStore(cluster cluster.Cluster, config config.Config) (Store, error) {
+	uuid.SetNodeID([]byte(cluster.Node().ID))
```
adibrastegarnia, line 39: I think it is safe to initialize it here instead initializing it at package level.

Issue #843, opened 2019-11-04
Finer grained Topo events filtering in Factory.go


Andrea-Campanella:
**Is your feature request related to a problem? Please describe.**
Currently the factory listens for all Topo events and uses just the ADDED ones to trigger device creation. See Line 39 of factory.go //TODO evaluate the need for fine-grained checking of conditions to connect a device. 

**Describe the solution you'd like**
A more tuned event handling should be considered, allowing for update and administrative disconnection.

adibrastegarnia:
A new version of session management is implemented which employs mastership information. Feel free to open it if you think it is not addressed yet.

PR #1189, opened 2020-08-31
Employ mastership info for session management 

adibrastegarnia:
This PR is making the required changes in the SB of the onos-config to employee mastership information for session management. It is a WIP and the code still needs more work and I will update this PR with more details about the changes. 

#960 https://github.com/onosproject/onos-config/issues/843 https://github.com/onosproject/onos-config/issues/962

P.S. there are some log messages for debugging but I will remove them at the end.


pkg/southbound/synchronizer/sessionManager.go
```diff
@@ -0,0 +1,246 @@
...
+			return err
+		}
+
+	case topodevice.ListResponse_NONE:
+		err := sm.createSession(event.Device)
+		if err != nil {
+			return err
+		}
+	case topodevice.ListResponse_UPDATED:
+		log.Info("Process device event updated")
```
adibrastegarnia, line 185: @kuujo 
In the current code, we check the address and if there is a change, we reconnect but if we change any other fields, we also need to update the state of a device. Should we create a new session anyway?

pkg/southbound/synchronizer/sessionManager_test.go
```diff
@@ -82,14 +69,25 @@ func TestFactory_Revert(t *testing.T) {
...
-		factory.TopoEventHandler()
-		wg.Done()
-	}()
+}
+
+/**
+ * Check device is added as a synchronizer correctly, times out on no gRPC device
+ * and then un-does everything
+ */
+func TestSessionManager(t *testing.T) {
```
adibrastegarnia, line 86: We need a new unit test here

pkg/southbound/synchronizer/session.go
```diff
@@ -0,0 +1,247 @@
...
+			s.mastershipState = state
+			s.mu.Unlock()
+			var currentTerm int
+			if s.device.Attributes == nil {
+				currentTerm = 0
+			} else {
+				currentTerm, _ = strconv.Atoi(s.device.Attributes[mastershipTermKey])
+			}
+
+			if state.Master == s.mastershipStore.NodeID() && uint64(state.Term) >= uint64(currentTerm) {
```
adibrastegarnia, line 97: @kuujo 
The logic is the same as what we have in onos-ric but with one addition to include mastership term but perhaps you can tell me if there is a better way to address it.

adibrastegarnia: It has a bug in it. Fixing it

test/gnmi/suite.go
```diff
@@ -65,6 +66,7 @@ func (s *TestSuite) SetupTestSuite() error {
 	err = helm.Chart("onos-config", onostest.OnosChartRepo).
 		Release("onos-config").
 		Set("image.tag", "latest").
+		Set("replicaCount", 2).
```
adibrastegarnia, line 69: I will change it later.

pkg/southbound/synchronizer/session.go
kuujo: Looks good

pkg/southbound/synchronizer/deviceUpdate.go
```diff
@@ -0,0 +1,117 @@
...
+	if err != nil {
+		return err
+	}
+
+	if topoDevice.Attributes == nil {
+		topoDevice.Attributes = make(map[string]string)
+	}
+
+	topoDevice.Attributes[mastershipTermKey] = strconv.FormatUint(uint64(mastershipState.Term), 10)
+	_, err = s.deviceStore.Update(topoDevice)
```
kuujo, line 69: IIRC the device service will reject this update if the device revision has changed. This is the concurrency control that ensures two nodes cannot overwrite each other’s changes. But since those updates could occur in any order (an older master might win the race, in which case the newer master’s write will be rejected), this entire read-modify-write sequence has to be retried until one of two conditions is met:
* The update is successful or
* Upon retry, the node encounters a mastership term greater than its own

pkg/southbound/synchronizer/sessionManager.go
kuujo: I think either way is fine.

kuujo (commented):
Looks really good. We just have to make sure the device state will be retried until it’s either successful or it’s superseded by a later master. Optimistic locking requires retries to function correctly. Also, if we’re going to retry device updates, we have to ensure multiple concurrent state changes to the same device occur in the correct order. If we just start a new goroutine for each update then they could be performed in any order. Perhaps in addition to the term we should store a local logical timestamp in the device attributes.

kuujo:
Either that or we can serialize updates to each device state on a separate channel.

pkg/southbound/synchronizer/sessionManager.go
```diff
@@ -0,0 +1,249 @@
...
+	operationalStateCache     map[topodevice.ID]devicechange.TypedValueMap
+	newTargetFn               func() southbound.TargetIf
+	operationalStateCacheLock *sync.RWMutex
+	deviceChangeStore         device.Store
+	mastershipStore           mastership.Store
+	mu                        sync.RWMutex
+}
+
+// NewSessionManager create a new session manager
+func NewSessionManager(options ...func(*SessionManager)) (*SessionManager, error) {
```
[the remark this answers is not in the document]
adibrastegarnia: @SeanCondon 
Thanks for the review. This is the PR that I opened 2 days ago and was waiting to get some feedback from you and @kuujo  before finalizing it. So I got some feedback now and I will fix them and after a bit testing we merge. I will add new unit tests later based on new changes.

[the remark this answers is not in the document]
adibrastegarnia: @SeanCondon 
The PR that you reviewed was a prerequisite to this one and we merged that one.

pkg/southbound/synchronizer/sessionManager.go
```diff
@@ -0,0 +1,249 @@
...
+	session := &Session{
+		opStateChan:               sm.opStateChan,
+		southboundErrorChan:       errChan,
+		dispatcher:                sm.dispatcher,
+		modelRegistry:             sm.modelRegistry,
+		operationalStateCache:     sm.operationalStateCache,
+		operationalStateCacheLock: sm.operationalStateCacheLock,
+		deviceChangeStore:         sm.deviceChangeStore,
+		device:                    device,
+		target:                    sm.newTargetFn(),
```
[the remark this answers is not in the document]
adibrastegarnia: I tried to avoid any code changes which are not related to session management thing but these are minor things and I will improve them in the same PR.

pkg/southbound/synchronizer/deviceUpdate.go
```diff
@@ -0,0 +1,158 @@
...
+			// TODO: Retry only on write conflicts
+			return backoff.Retry(func() error {
+
+				/* This entire read-modify-write sequence has to be retried until one of two conditions is met:
+				   - The update is successful or
+				   - Upon retry, the node encounters a mastership term greater than its own */
+
+				session := sm.sessions[id]
+				if session != nil {
+					currentTerm, err := session.getCurrentTerm()
```
kuujo, line 108: Even though we’re getting the device’s current term from the topology and checking it against the local mastership term, this particular implementation doesn’t guarantee the update cannot be overwritten by an older master. In order to prevent that, the read-modify-write sequence comparing the stored term with the local term and then updating the device _must be atomic_, and in order for it to be atomic it _must operate on the same revision_ of the device. When a device is read from the topology service, the device is returned with metadata that includes a revision number. When a device is updated, the topology service compares the device’s revision number with the stored revision number and rejects the update if the revisions don’t match. If an update is rejected, the client attempts the atomic read-modify-write sequence again.

But the code here is actually reading the device twice: once to retrieve the term, and once to update the device, so the sequence is not atomic. The term can change between the first and second read, and that would allow an older master to overwrite an update from a newer master. For example:
* Node A reads the device and verifies the stored term matches its local term
* Node B is elected master for a later term and updates the device state
* Node A proceeds to read the device again and updates it _without checking the term_, overwriting the update from new master node B

Both the condition checking the device’s stored term and the write updating the device state must occur on the same device object returned from the topology service.

pkg/southbound/synchronizer/sessionManager.go
```diff
@@ -0,0 +1,279 @@
...
+func (sm *SessionManager) Start() error {
+	log.Info("Session manager started")
+	err := sm.deviceStore.Watch(sm.topoChannel)
+	if err != nil {
+		return err
+	}
+
+	go sm.processDeviceEvents(sm.topoChannel)
+	go func() {
+		err := sm.updateDeviceState()
```
kuujo, line 154: I think we should create a separate event channel and goroutine for each session to allow session state changes to be notified and the topology to be updated concurrently. A new goroutine to process session state changes could be created in the session manager or in the session itself. We only need to maintain the order of updates within each session. Sharing a channel across sessions can only block the processing of session updates unnecessarily. A per-session loop could also simplify the algorithm for updating the device state.

pkg/southbound/synchronizer/deviceUpdate.go
```diff
@@ -0,0 +1,148 @@
...
+	} else {
+		protocolState = new(topodevice.ProtocolState)
+	}
+
+	protocolState.Protocol = topodevice.Protocol_GNMI
+	protocolState.ConnectivityState = connectivity
+	protocolState.ChannelState = channel
+	protocolState.ServiceState = service
+	topoDevice.Protocols = append(topoDevice.Protocols, protocolState)
+	mastershipState, err := s.mastershipStore.GetMastership(topoDevice.ID)
```
kuujo, line 58: To ensure the update is atomic we just need to compare the mastership state to the stored device term after this.

kuujo:
The file naming convention in Go is to use `lower_case_and_underscores.go`. We're mixing this with `lowerCamelCase.go` here and in other areas of onos-config. We should avoid doing that here and fix the places where it's already done in another PR.

pkg/southbound/synchronizer/device_update.go
```diff
@@ -0,0 +1,145 @@
...
+
+	// Read the current term for the given device
+	currentTerm, err := s.getTermPerDevice(topoDevice)
+	if err != nil {
+		return err
+	}
+
+	// Do not update the state of a device if the node encounters a mastership term greater than its own
+	if uint64(mastershipState.Term) < uint64(currentTerm) {
+		return errors.New("device mastership term is greater than node mastership term")
```
kuujo, line 82: When a higher mastership term is discovered, this node needs to stop attempting this update. But since this function is being called by the `backoff` library, this error will just cause it to retry and create a deadlock since the update will always fail after this condition is met. The error returned has to be `backoff.NewPermanent` or whatever the error was that stops the retries.

pkg/southbound/synchronizer/device_update.go
```diff
@@ -0,0 +1,145 @@
...
+	} else {
+		protocolState = new(topodevice.ProtocolState)
+	}
+
+	protocolState.Protocol = topodevice.Protocol_GNMI
+	protocolState.ConnectivityState = connectivity
+	protocolState.ChannelState = channel
+	protocolState.ServiceState = service
+	topoDevice.Protocols = append(topoDevice.Protocols, protocolState)
+	mastershipState, err := s.mastershipStore.GetMastership(topoDevice.ID)
```
kuujo, line 69: This mastership state doesn’t necessarily represent the state on which this update should be based. Since we’re reading from the mastership store on every update attempt, it’s possible for this node to read a _different_ (newer) mastership state (e.g. for a new master/term) than this update is based on. Who knows if this node is even the master any more. Rather than reading from the store here, this function should take as an argument the mastership state on which the update is based so that every attempt to perform the update is based on that state. Reading stores multiple times can alway introduce this sort of consistency issue. The only reason we should read the store here is to determine whether the state on which the update is based should be cancelled due to a new master/higher term.

pkg/southbound/synchronizer/device_update.go
adibrastegarnia: Ooops. You are right. We should not return the error here.

pkg/southbound/synchronizer/session.go
```diff
@@ -0,0 +1,223 @@
...
+
+	go func() {
+		_ = s.updateDeviceState()
+
+	}()
+	go func() {
+		s.mu.Lock()
+		s.connected = false
+		s.mu.Unlock()
+		state, _ := s.mastershipStore.GetMastership(s.device.ID)
```
kuujo, line 92: The session should be created with a mastership state. It shouldn't need to read the mastership state at all. That creates another race where the mastership could change between the time the session was created and the time it was opened. Just remove the `mastershipStore` field.

adibrastegarnia: @kuujo 
Done. I hope I addressed it properly.

PR #1199, opened 2020-09-06
JSON handling: new method DecomposeJSONWithPaths for decomposing in to Path Values Fixes #1197

### LINKED ISSUE #1197: gNMI Set does not handle YANG double keyed lists properly
Not a problem for any of the main plugins at the moment, but better to fix it

SeanCondon:
The old method could not handle double (or greater) keyed lists

This is used with Set() passing in JSON values and synchronize decomposing JSON results (op state) in to paths

`DecomposeJSONWithPaths()` replaces the combined approach of `DecomposeTree()` and `CorrectJSONPaths()`

Also changes the path format for the double key to `[key=1][key2=1]` to match what `BuildTree()` gives. 
e.g. give
`/system/logging/console/selectors/selector[facility=*][severity=*]/state`
instead of 
`/system/logging/console/selectors/selector[facility severity=*]/state` (wrong)


adibrastegarnia:
@SeanCondon 
A big improvement. Please also make sure if we made any assumption in integration tests about paths they consistent with new changes and they can pass.

adibrastegarnia (approved):

SeanCondon:
thanks @adibrastegarnia for review - integration tests are GTG - they don't use double indices

PR #1203, opened 2020-09-09
NBI Set: take in to account path when handling JSON payloads

SeanCondon:
Also fix problem where extra config paths coming up from Device (should only be state)
were causing the integration tests to fail


adibrastegarnia:
lgtm

adibrastegarnia (approved):

ray-milkey (approved):

Issue #1045, opened 2020-01-17
Performing gNMI Sets in rapid succession creates NetworkChanges, and DeviceChanges but stay PENDING


SeanCondon:
**Describe the bug**
When gNMI Sets are made in rapid succession, the corresponding `NetworkChange` objects are created, properly, but no corresponding `DeviceChange` objects are created.

One time I ran this there were no DeviceChanges, but I cannot reproduce it at the moment. Still this is worth investigating

**To Reproduce**
1) Create a cluster using `onit`:
```bash
onit create cluster --set onos-cli.enabled=true
```
2) add a device simulator
```bash
onit -c onos add simulator -n devicesim-1
```

3) Make a set of gnmi Set changes together by pasting all the following in to the terminal
```bash
gnmi_cli -address onos-config:5150 -set \
    -proto "update: <path: <target: 'devicesim-1', elem: <name: 'system'> elem: <name: 'clock' > elem: <name: 'config'> elem: <name: 'timezone-name'>> val: <string_val: 'Europe/Dublin'>>" \
    -timeout 5s -en PROTO -alsologtostderr \
    -client_crt /etc/ssl/certs/client1.crt -client_key /etc/ssl/certs/client1.key -ca_crt /etc/ssl/certs/onfca.crt

gnmi_cli -address onos-config:5150 -set \
    -proto "update: <path: <target: 'devicesim-1', elem: <name: 'system'> elem: <name: 'clock' > elem: <name: 'config'> elem: <name: 'timezone-name'>> val: <string_val: 'Europe/London'>>" \
    -timeout 5s -en PROTO -alsologtostderr \
    -client_crt /etc/ssl/certs/client1.crt -client_key /etc/ssl/certs/client1.key -ca_crt /etc/ssl/certs/onfca.crt

gnmi_cli -address onos-config:5150 -set \
    -proto "update: <path: <target: 'devicesim-1', elem: <name: 'system'> elem: <name: 'clock' > elem: <name: 'config'> elem: <name: 'timezone-name'>> val: <string_val: 'Europe/Edinburgh'>>" \
    -timeout 5s -en PROTO -alsologtostderr \
    -client_crt /etc/ssl/certs/client1.crt -client_key /etc/ssl/certs/client1.key -ca_crt /etc/ssl/certs/onfca.crt

gnmi_cli -address onos-config:5150 -set \
[...]

PR #1206, opened 2020-09-11
Add multiple  set (back to back) integration test

adibrastegarnia:
This PR adds a simple integration tests to run multiple back to back set operations and make sure all of the changes are in the Completed state. We have an issue https://github.com/onosproject/onos-config/issues/1045 which is not reproducible at the moment but this test will be detect if that happens again.


adibrastegarnia:
will open a new one

PR #1211, opened 2020-09-15
Refactoring to use umbrella chart with Atomix in the default namespace

ray-milkey:


test/ha/suite.go
```diff
@@ -15,60 +15,21 @@
...
-		Set("image.tag", "latest").
-		Set("replicaCount", "2").
-		Set("storage.controller", onostest.AtomixController(testName, onosComponentName)).
-		Install(true)
-	if err != nil {
-		return err
-	}
-
-	return nil
+	umbrella := charts.CreateUmbrellaRelease()
```
adibrastegarnia, line 33: For HA, we run two instance of onos-config but we are not ready yet to run two instance of onos-config for gnmi tests. Looks like we are using one function for both

kuujo: We should set the values on the returned release then rather than inside `CreateUmbrellaRelease()`.

kuujo: nvm I guess you already did that 🤷

benchmark/gnmi/suite.go
```diff
@@ -37,52 +34,12 @@ type BenchmarkSuite struct {
...
-		Set("storage.controller", onostest.AtomixController(testName, onosComponentName)).
-		Set("image.tag", "latest").
-		Install(true)
-	if err != nil {
-		return err
-	}
-	return nil
+	umbrella := charts.CreateUmbrellaRelease().
+		Set("onos-topo.replicaCount", 2).
+		Set("onos-config.replicaCount", 2)
```
kuujo, line 41: @adibrastegarnia this should be 1 replica, right?

test/ha/suite.go
kuujo: nvm I guess we're already doing that, but this should set the replica count to `2`, and the gNMI test should leave it at `1` AIUI

benchmark/gnmi/suite.go
ray-milkey: The previous code set it to two. Not sure why.

kuujo: I think that I was just being optimistic 😛  If the tests pass we should leave it there though.

adibrastegarnia (approved):

Issue #1210, opened 2020-09-14
gNMI Support for Multiple Client Roles and Master Arbitration


adibrastegarnia:
**Is your feature request related to a problem? Please describe.**
To have a solid HA in onos-config and support multiple replicas of a gNMI client and , we need to support **Master arbitration**  as it is explained here: https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-master-arbitration.md.  Master arbitration is  critical for making configuration of a device with multiple potential masters safe (i.e. Configuration management has to be strongly consistent)

Since we use mastership information to establish connections to the target and update the device state  (look at this PR and the issues that are linked to the PR: https://github.com/onosproject/onos-config/pull/1189), we can extend the usage of it for the SB operations (i.e. gnmi set requests) 

**Describe the solution you'd like**
To achieve the goal, first we need to implement the required logic for storing mastership term in the gNMI simulator to accept the requests from the master node that has a mastership term that is equal to or greater than the currently stored one. The logic which is needed in client and the target is explained here: https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-master-arbitration.md#32-setrequest-handling

PR #1215, opened 2020-09-17
Sets master arbitration extension 

adibrastegarnia:
This PR adds a function that will be called after we receive an event that onos-config instance is connected to the device but before updating the device stat to  send an empty Set RPC with the election ID (i.e. mastership term) only. 

#1210

### DIFF
--- a/pkg/southbound/synchronizer/device_update.go
+++ b/pkg/southbound/synchronizer/device_update.go
@@ -18,11 +18,16 @@ import (
 	"errors"
 	"strconv"
 
+	"github.com/openconfig/gnmi/proto/gnmi_ext"
+
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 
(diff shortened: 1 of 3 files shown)


pkg/southbound/synchronizer/device_update.go
```diff
@@ -120,11 +124,39 @@ func (s *Session) updateDisconnectedDevice() error {
...
+			ElectionId: electionID,
+		},
+	}
+
+	extensions := []*gnmi_ext.Extension{{Ext: &masterArbitrationExt}}
+
+	masterArbitrationReq := &gpb.SetRequest{
+		Extension: extensions,
+	}
+	_, err := s.target.Set(s.ctx, masterArbitrationReq)
```
adibrastegarnia, line 146: @kuujo 
Is this OK to store context in the session and use it here?

SeanCondon (approved):
LGTM

PR #1218, opened 2020-09-25
makefile improvements

sbconsulting:
Add ability to override kind cluster name and docker repository.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,8 @@ export GO111MODULE=on
 
 .PHONY: build
 
-ONOS_CONFIG_VERSION := latest
+KIND_CLUSTER_NAME ?= kind
+ONOS_CONFIG_VERSION ?= latest
 ONOS_BUILD_VERSION := v0.6.3
 ONOS_PROTOC_VERSION := v0.6.3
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -53,15 +54,18 @@ onos-config-base-docker: # @HELP build onos-config base Docker image
 onos-config-docker: onos-config-base-docker # @HELP build onos-config Docker image
 	docker build . -f build/onos-config/Dockerfile \
 		--build-arg ONOS_CONFIG_BASE_VERSION=${ONOS_CONFIG_VERSION} \
-		-t onosproject/onos-config:${ONOS_CONFIG_VERSION}
+		-t ${DOCKER_REPOSITORY}onos-config:${ONOS_CONFIG_VERSION}
```
SeanCondon, line 57: `${DOCKER_REPOSITORY}` does not seem to be set to any default here?

SeanCondon (commented):
I'd like @ray-milkey to have a look too

Makefile
sbconsulting: good catch, fixed.

ray-milkey: If I understand this change correctly, DOCKER_REPOSITORY should have a default of onosproject and this line should read ${ONOS_REPOSITORY}/onos-config

ray-milkey (changes requested):

Makefile
sbconsulting: I was starting to adopt the convention we used in Voltha, which was that each Makefile would expose a `DOCKER_REPOSITORY` environment variable, then for example a user does `DOCKER_REPOSITORY=myprivaterepo/ make images`. We did this rather than creating uniquely named environment variables for each repo. If it would be convenient to have discussion about how the variables are named we can do that next week -- I don't have a strong opinion, and just used the convention I was already used to. The default was missing from the initial patch; I've pushed an update today. There's a similar patch up for review for the config-models repo.

SeanCondon (approved):

ray-milkey (approved):

PR #1221, opened 2020-10-02
Enable transaction test

adibrastegarnia:
I think after the last bug that I fixed this test should be OK but I would like to see if it has any other problems than the previous one. 

P.S. after a little bit more testing, we can merge it.


ray-milkey (approved):
LGTM - ran four times with no troubles

PR #1226, opened 2020-10-09
Add a test for the case where rollback deletes a path

ray-milkey:


test/gnmi/deletetest.go
```diff
@@ -0,0 +1,82 @@
...
+	assert.NoError(t, err)
+	rollbackResponse, rollbackError := adminClient.RollbackNetworkChange(
+		context.Background(), &admin.RollbackRequest{Name: string(changeID)})
+
+	assert.NoError(t, rollbackError, "Rollback returned an error")
+	assert.NotNil(t, rollbackResponse, "Response for rollback is nil")
+	assert.Contains(t, rollbackResponse.Message, changeID, "rollbackResponse message does not contain change ID")
+
+	// Check that the value was really rolled back- should be an error here since the node was deleted
+	_, _, err = gnmi.GetGNMIValue(gnmi.MakeContext(), device1GnmiClient, devicePathsForGet)
```
adibrastegarnia, line 80: Perhaps you can add a TODO or something that we fix this again when we fix  the rollback.

ray-milkey: The next line asserts that an error was detected

adibrastegarnia (approved):

PR #1228, opened 2020-10-09
Check the devices to be sure the rollback happened

ray-milkey:


adibrastegarnia (approved):
@ray-milkey 
Yeah that is good to verify the device has the previous change after rollback.

Issue #1023, opened 2020-01-09
Validation of changes on gNMI set not consistent


kuujo:
Model validation in the northbound API is done prior to writing changes to stores. However, this actually creates a consistency issue for model validation. Even though a change has been validated, once it’s written to the store it may be reordered among concurrent changes, meaning a change is not necessarily validated against the state to which it will be applied.

To preserve validation in the northbound prior to writing to stores, stores would have to be locked to prevent concurrency in the northbound API. The only efficient solution is for changes to be written to stores before they’re validated, and for controllers to be allowed to validate and fail network changes.

PR #1229, opened 2020-10-13
Move validation from NB to the controller

adibrastegarnia:
This PR is moving the validation of changes from NB to the controllers. It is still a WIP and perhaps it will be changed. 

This issue for more information 
https://github.com/onosproject/onos-config/issues/1023


api/types/change/types.proto
```diff
@@ -53,8 +55,13 @@ enum State {
     // COMPLETE indicates the phase is complete
     COMPLETE = 2;
 
+
+
+    // VALIDATION_FAILED indicates validation phase is failed
+    VALIDATION_FAILED = 3;
```
kuujo, line 61: I think we should add validation as a failure `Reason` rather than a different state

adibrastegarnia: @kuujo 
I understand your point and I would like to do that as well but the reason that I used a different state, its because we don't need to rollback a change when it fails because of validation. That means in the code where we are rolling back a failed change, we need to make sure it is not because of validation which means we should include reason in checking of the conditions. Do you think Is this OK?

pkg/northbound/gnmi/set.go
```diff
@@ -418,20 +436,3 @@ func buildUpdateResult(pathStr string, target string, op gnmi.UpdateResult_Opera
 	return updateResult, nil
 
 }
-
-func validateChange(target string, deviceType devicetype.Type, version devicetype.Version,
-	targetUpdates devicechange.TypedValueMap, targetRemoves []string, lastWrite networkchange.Revision) error {
-	if len(targetUpdates) == 0 && len(targetRemoves) == 0 {
```
adibrastegarnia, line 424: I should add this to validation function

pkg/controller/change/device/validation.go
```diff
@@ -0,0 +1,87 @@
...
+
+	changeValues := change.Change.Values
+
+	modelName := utils.ToModelName(deviceType, version)
+	deviceModelYgotPlugin, ok := r.modelReg.ModelPlugins[modelName]
+	if !ok {
+		log.Warn("No model ", modelName, " available as a plugin")
+		return fmt.Errorf("no model %s available as a plugin", modelName)
+
+		//return nil
```
SeanCondon, line 40: can remove this

pkg/controller/change/device/validation.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	pathValues := make(map[string]*devicechange.TypedValue)
+	for _, configValue := range changeValues {
```
SeanCondon, line 44: can these 2 for loops be combined?

pkg/controller/change/device/validation.go
```diff
@@ -0,0 +1,87 @@
...
+import (
+	"fmt"
+	"sort"
+
+	devicechange "github.com/onosproject/onos-config/api/types/change/device"
+	"github.com/onosproject/onos-config/pkg/store"
+	"github.com/onosproject/onos-config/pkg/utils"
+)
+
+func (r *Reconciler) validateDeviceChange(change *devicechange.DeviceChange) error {
```
SeanCondon, line 26: validation should be on the whole configuration plus the DeviceChange - not just on a single DeviceChange on its own

A single DeviceChange might not be valid on its own, and even if it was it does not tell you whether the whole configuration is valid after the DeviceChange would be a applied

SeanCondon (commented):
validation should be on the whole configuration plus the DeviceChange - not just of a single DeviceChange on its own

pkg/controller/change/device/validation.go
adibrastegarnia: @SeanCondon Thanks for starting the review. Just to make it clear for myself. Do you mean a network change (i.e. multiple device changes) is not valid if any of device changes is not valid? If that is the case which is similar to what we do in NB by going through all the changes in a loop to validate, I took care of it in Network change controller to fail a network change if validation of device changes is failed. If you are talking about a new logic that should be added to what we have now, please explain it with more details and I will add it.

SeanCondon: @adibrastegarnia What I meant is that the validation should be applied to:
all previous DeviceChanges for that device + Snapshot for the Device + the new DeviceChange
(as if they were all squashed down to one config).

I think from the code you are applying the validation to only the DeviceChange in hand.

adibrastegarnia: @SeanCondon 
Yeah I didn't handle that because I just moved what we have in NB which seems validate just the current request coming from NB to the controller to validate the requested change but what would be the logic (validation function) look like for that kind of validation which should consider all of the previous changes and snapshot?  It sounds like we should retrieve everything build a new json tree using ygot and validate that tree which has the whole configuration including the new change.

SeanCondon: @adibrastegarnia See line 51 of setconfig.go in existing code
https://github.com/onosproject/onos-config/blob/d0ff6cf04aa47e02244fad1fe9a0398a69de522a/pkg/manager/setconfig.go#L51
It is pulling all the existing changes from store and merges them with the new proposed changes and then does build tree on the whole lot and passes this for validation

adibrastegarnia: @SeanCondon 
Thanks for the pointer. I thought that one is returning the last change so I missed it.

PR #1234, opened 2020-11-27
Update ygot and protobuf deps

adibrastegarnia:


SeanCondon (commented):
Looks good - it will be great to upgrade to these versions

I am seeing an error on linting though of 
`test/utils/gnmi/utils.go:30:2: SA1019: Package github.com/golang/protobuf/proto is deprecated: Use the "google.golang.org/protobuf/proto" package instead.  (staticcheck)`

We might have to turn off the deprecation warning in `.golangci.yml` until we can replace these classes

adibrastegarnia:
@SeanCondon 
Thanks. Yeah those updates are needed for new config models which use ygot 0.8.12. I was planning to disable linter for now because those changes are a little bit time time consuming to make sure everything is fine.

adibrastegarnia:
The dep for unit tests also need to be updated after we merge the PR in config models.

SeanCondon (approved):

PR #1239, opened 2020-12-03
Reemove integration tests and CLI build from merge commits

ray-milkey:


adibrastegarnia (changes requested):
why there is a change in go.mod?

adibrastegarnia:
these updates in go.mod affect plugins.

adibrastegarnia (approved):

PR #1240, opened 2020-12-03
Migrating away from onos-topo/api/device API to onos-api/go/onos/topo API

tomikazi:


pkg/store/snapshot/device/store.go
```diff
@@ -16,7 +16,7 @@ package device
 
 import (
 	"context"
-	"errors"
+	"github.com/atomix/go-client/pkg/client/errors"
```
SeanCondon, line 19: should this be using `onos-lib-go/pkg/errors` instead? It think it probably amounts to the same thing

SeanCondon (approved):
Looks good from my POV - 1 minor comment

pkg/device/device.go
```diff
@@ -0,0 +1,153 @@
...
+				Protocols: device.Protocols,
+			},
+		},
+	}
+
+	if o.Attributes == nil {
+		o.Attributes = make(map[string]string)
+	}
+
+	o.Attributes[topo.Type] = string(device.Type)
```
SeanCondon, line 131: @tomikazi I think part of the problem in the integration tests is the mapping between `Type`, `Version` and `Id` of `Device` which are fundamental to `onos-config` and the corresponding placeholders in the new Topo Entity-Kind model.

In the example topology loaded by `onos-umbrella` at 
https://github.com/onosproject/onos-helm-charts/blob/a764d20b4d7f8725ffb6b5f8609dece1d3ce2d5c/onos-umbrella/files/configs/berlin-grid-4-topo.yaml#L27
you can see that I did not add Type as an attribute, as I presumed it was already defined in the Entity/KindID. 

I see however that in this patch, you expect an explicity defined attribute called "Type", and the 'KindID' is set to none.

I think in general when Topo Entities are being synced to `onos-config`, if they do not have an explicit "Type" and "Version" attribute, then they should be ignored as being "**non-config**". It is always possible that there are non-config Entities in `onos-topo` as Topo will have more uses than just driving `onos-config`

pkg/store/snapshot/device/store.go
tomikazi: Good catch. Fixed.

kuujo: Actually, we need both. The errors being returned by the store should be using `onos-lib-go/pkg/errors`, but the errors being received from the Atomix primitive are of the `atomix/go-client/pkg/errors` type. So, a check like `if errors.IsNotFound(err)` on an error returned from the Atomix primitive _has_ to use the Atomix `errors` package, and Atomix errors should be converted to ONOS errors using the ONOS `errors.FromAtomix` function. We certainly should be doing that everywhere there's an error so stores are not returning Atomix errors.

What you can

kuujo: Check out the way it's implemented in onos-e2sub:
https://github.com/onosproject/onos-e2sub/blob/master/pkg/store/subscription/store.go
Every error returned by Atomix is converted to an ONOS error with `errors.FromAtomix`.

pkg/store/snapshot/device/store.go
```diff
@@ -316,7 +316,7 @@ func (s *atomixStore) Load(deviceID device.VersionedID) (*devicesnapshot.Snapsho
 	defer cancel()
 
 	entry, err := s.snapshots.Get(ctx, string(deviceID))
-	if err != nil {
+	if err != nil && !errors.IsNotFound(err) {
```
kuujo, line 319: This `errors.IsNotFound(err)` condition will _only_ work using the Atomix `errors` package since it's an Atomix error that's returned by `snaphsots.Get`. If you convert it into an ONOS error first, then you can use the ONOS `errors` package to check `IsNotFound`.

pkg/device/device.go
tomikazi: I've remapped KindId to Type and vice-versa.

PR #1242, opened 2020-12-15
Replace model registry with model API

kuujo:
This PR is the initial migration to the new model plugin API. This project is a work in progress. We'll need to do sufficient testing before we can merge this.


pkg/northbound/gnmi/set.go
```diff
@@ -313,14 +333,70 @@ func (s *Server) doDelete(prefix *gnmi.Path, u *gnmi.Path,
...
-	if err != nil {
-		return nil, err
-	}
 	deletes = append(deletes, path)
 	return deletes, nil
+}
+
+// iterate through the updates and check that none of them include a `set` of a
+// readonly attribute - this is done by checking with the relevant model
+func (s *Server) checkForReadOnly(target string, deviceType devicetype.Type, version devicetype.Version,
```
SeanCondon, line 342: `checkForReadOnly` that was there previously has been replaced - I use the `findPathFromModel` instead, because it has the benefit of finding the `type` that the path refers to, and at the same time searches throught the set of Read-Write paths in the model - and so Read-Only paths will be rejected

SeanCondon (commented):
My only issue is with `findPathFromModel`, otherwise this looks great

PR #1250, opened 2021-01-05
Do replay when watching topo

SeanCondon:
`onos-config` calls `Watch` on `onos-topo` on connection. This must replay existing entities - with `noreplay=true` only Entities that are created after startup will arrive in to `onos-config`.

An alternate approach is to do `List` first for existing devices and then watch, but this is never satisfactory as there will always be either a gap or an overlap with the 2 methods, with the risk of duplicates or missing entities.

Also I added some filters on Entity events from `onos-topo` - a **version** and a **kindid** are always required for onos-config.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -24,7 +24,7 @@ require (
 	github.com/onosproject/onos-api/go v0.7.0
 	github.com/onosproject/onos-lib-go v0.7.0
 	github.com/onosproject/onos-test v0.6.4
-	github.com/onosproject/onos-topo v0.7.0
+	github.com/onosproject/onos-topo v0.7.0 // indirect
 	github.com/openconfig/gnmi v0.0.0-20200617225440-d2b4e6a45802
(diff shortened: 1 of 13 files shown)


adibrastegarnia:
We create device simulators here: I think we need to add kind here as well:
https://github.com/onosproject/onos-config/blob/0699cfe765250c9ca0ed58526fb1fd3fcd19ac97/test/utils/gnmi/utils.go#L133

SeanCondon:
> We create device simulators here: I think we need to add kind here as well:
> https://github.com/onosproject/onos-config/blob/0699cfe765250c9ca0ed58526fb1fd3fcd19ac97/test/utils/gnmi/utils.go#L133

Kind is taken from the `Type` given here - I've ran the `gnmi` suite and they all pass

adibrastegarnia (approved):

tomikazi (approved):

PR #1254, opened 2021-01-19
Handling updated Value Type from onos-api

SeanCondon:
In onos-config the decoding of JSON follows RFC-7951 which needs int64 and uint64 to be represented as string rather than a number. For smaller widths they can be treated as number.

When YGOT is validating it goes through the model and expects the JSON encoded data to suit the data type given.

Basically the type cannot be deduced from the value alone.

This relies on the updated onos-api v0.7.2 that uses the "TypeOpts" field of the TypedValue to hold the number width, so that it can be used at encoding time.

This requires the ModelPlugin to be consulted for every gNMI Set() operation, with the impact:
* The old method of check for RO paths is no longer needed - we check for RW paths every time
* Some of the error messages have changes
* The model plugins now have to be loaded in to all of the gnmi_tests
* Some of the `set_tests` and `subscribe_tests` had to be changed to have valid data types and values.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,6 @@ require (
 	github.com/atomix/go-client v0.4.1
 	github.com/cenkalti/backoff v2.2.1+incompatible
 	github.com/docker/docker v1.13.1 // indirect
-	github.com/go-playground/overalls v0.0.0-20191218162659-7df9f728c018 // indirect
 	github.com/gogo/protobuf v1.3.1
 	github.com/golang/mock v1.4.4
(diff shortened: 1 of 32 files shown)


adibrastegarnia (approved):
LGTM.  would be nice to run integration tests locally and see if the changes are OK with assumptions in the tests.

SeanCondon:
Thanks @adibrastegarnia - i did run the integration tests (gnmi) already, and they all pass

PR #1257, opened 2021-01-29
Sc added authentication

SeanCondon:


SeanCondon:
Abandoning because it's based off an old branch

PR #1263, opened 2021-02-11
Introducing Role Based Access Control (RBAC) - build RBAC cache

SeanCondon:
Part 1 of enforcing RBAC - build the RBAC cache by listening to events from "rbac" DeviceChanges store

Only do this if authentication is turned on

### DIFF
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -17,6 +17,7 @@ package manager
 
 import (
 	"fmt"
+	"github.com/onosproject/onos-config/pkg/store/change/device/rbac"
 	"sync"
 
 	devicechange "github.com/onosproject/onos-api/go/onos/config/change/device"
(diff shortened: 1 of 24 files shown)


pkg/manager/manager.go
```diff
@@ -186,42 +190,6 @@ func GetManager() *Manager {
 	return &mgr
 }
 
-// ComputeDeviceChange computes a given device change the given updates and deletes, according to the path
-// on the configuration for the specified target
-func (m *Manager) ComputeDeviceChange(deviceName devicetype.ID, version devicetype.Version,
```
SeanCondon, line 191: I just moved this to `setconfig.go` in the same package - it's only ever used from there so did not need to be in this file

adibrastegarnia (approved):
LGTM. please make sure the tests are OK.

PR #1302, opened 2021-03-30
AETHER-1290: prevent setting invalid index value in list

SeanCondon:
https://jira.opennetworking.org/browse/AETHER-1290

When setting an item in a list the index value can be set like any other value - but it must agree with the corresponding Key value in the list - otherwise it's an invalid index

This is passing the gnmi suite in helmit tests


adibrastegarnia (approved):
LGTM but perhaps @sbconsulting  should also take a look.

PR #1310, opened 2021-04-20
Prevent SEGV in Rollback when called with no arg. And releasing v0.7.23

SeanCondon:
In Rollback - if no network change ID is given (blank string) then onos-config craches with SEGV

Also I changed errors to give the full response to caller, and not to log the error. This is because the action is synchronous and related only to the users action.

Also I am returning gRPC error codes as I do now in gNMI - by default it is `unknown`

I also fixed a helmit test whose expected error message had got out of sync


pkg/manager/rollbackconfig.go
```diff
@@ -20,31 +20,38 @@ import (
...
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 )
 
 // RollbackTargetConfig rollbacks the last change for a given configuration on the target, by setting phase to
 // rollback and state to pending.
 func (m *Manager) RollbackTargetConfig(networkChangeID networkchange.ID) error {
 
+	if networkChangeID == "" {
+		return status.Error(codes.InvalidArgument, "error on rollback. networkChangeID is empty")
```
adibrastegarnia, line 32: I think we should use onos-lib-go error package which has the ability to convert typed errors to gRPC errors and vice versa.

SeanCondon: @adibrastegarnia hopefully I got the ones you meant

adibrastegarnia (approved):

pkg/manager/rollbackconfig.go
adibrastegarnia: @SeanCondon thanks

PR #1323, opened 2021-05-11
Refactoring to work with modified onos-topo API

tomikazi:


pkg/device/device.go
```diff
@@ -124,85 +128,110 @@ func (x ListResponseType) String() string {
...
-		ID:         topo.ID(device.ID),
-		Revision:   device.Revision,
-		Attributes: device.Attributes,
-		Type:       topo.Object_ENTITY,
-		Obj: &topo.Object_Entity{
-			Entity: &topo.Entity{
-				KindID:    topo.ID(device.Type),
-				Protocols: device.Protocols,
+	o := device.Object
+	if o == nil {
```
adibrastegarnia, line 134: Why this condition is needed?

adibrastegarnia (approved):

adibrastegarnia:
If you can run integration tests then would be great @tomikazi

pkg/device/device.go
tomikazi: This is in case the original device object were synthesized from scratch.

tomikazi:
> If you can run integration tests then would be great @tomikazi

This will take some doing and will have to wait until the CLI is done as well. Can't do it before then.

tomikazi:
retest this please

PR #1325, opened 2021-05-16
Migrate to Atomix v2beta1 API

kuujo:
This PR migrates the onos-config stores to the new Atomix v2beta1 API. This change completes the decoupling of the code from the storage implementation, allowing storage semantics to be managed entirely by configuration via Kubernetes custom resources. The onos-topo container simply gets an instance of the primitive by name, and how and where the data is stored is managed by the Kubernetes control plane.

This change depends on changes to the Helm charts which will be submitted in additional PRs in those repos. To avoid inconsistencies between the Docker image and the Helm chart, do not merge this until both PRs have been tested and are ready to be released.


SeanCondon (approved):

pkg/controller/change/network/controller.go
```diff
@@ -201,7 +204,7 @@ func (r *Reconciler) createDeviceChanges(networkChange *networkchange.NetworkCha
 	prevChange, err := r.networkChanges.GetByIndex(networkChange.Index - 1)
 	if err != nil && !errors.IsNotFound(err) {
 		return controller.Result{}, err
-	} else if err == nil && !hasDeviceChanges(prevChange) {
+	} else if prevChange != nil && !hasDeviceChanges(prevChange) {
 		return controller.Result{Requeue: controller.NewID(string(networkChange.ID))}, nil
```
SeanCondon, line 208: It does no harm to `Requeue` - but probably not needed, since we have a firehose of events coming in to the Watcher anyway

SeanCondon, line 208: It does no harm to `Requeue` - but probably not needed, since we have a firehose of events coming in to the Watcher anyway

PR #1328, opened 2021-05-17
Use tag named modules

ray-milkey:


SeanCondon (approved):
this is great @ray-milkey - good to tidy it up

PR #1333, opened 2021-05-19
Changes to move Protocols into aspects

ray-milkey:


go.mod
```diff
@@ -5,22 +5,34 @@ go 1.14
...
+	github.com/kardianos/osext v0.0.0-20190222173326-2bc1f35cddc0 // indirect
+	github.com/kr/pty v1.1.8 // indirect
 	github.com/mattn/go-colorable v0.1.4 // indirect
 	github.com/mattn/go-isatty v0.0.12 // indirect
 	github.com/onosproject/config-models/modelplugin/devicesim-1.0.0 v0.6.28
 	github.com/onosproject/config-models/modelplugin/testdevice-1.0.0 v0.6.28
 	github.com/onosproject/config-models/modelplugin/testdevice-2.0.0 v0.6.28
 	github.com/onosproject/helmit v0.6.12
-	github.com/onosproject/onos-api/go v0.7.38
+	github.com/onosproject/onos-api/go v0.7.43
```
SeanCondon, line 35: you'll have to update `onos-config-model`'s `go.mod` to use this same version of the API - otherwise you'll get errors when `onos-operator` calls on `model-registry` to compile the YANG files - it's not all plain sailing with `onos-operator`.
When you do you'll have to do a new release of it (4.7) and then refer to it in the `onos-config` helm chart at 
https://github.com/onosproject/onos-helm-charts/blob/327693407fc0f8c93c44e7030fbfb90cdb6c184d/onos-config/values.yaml#L41

go.mod
```diff
@@ -5,22 +5,34 @@ go 1.14
...
 	github.com/onosproject/onos-config-model v0.4.3
 	github.com/onosproject/onos-lib-go v0.7.10
```
SeanCondon, line 37: can you update this to `v0.7.12` too if you don't mind - same in `onos-config-model` too

SeanCondon (approved):

go.mod
kuujo: @SeanCondon we should not need to update `onos-config-model` even when we change the dependencies it shared with `onos-config`. This is part of the problem it's intended to solve, so if it's necessary to change dependencies elsewhere then it's a bug. 

If you've run into this problem, it could just be a configuration issue. The `onos-config-model` compiler does not use its own dependencies to compile plugins. It downloads the dependencies from the `onos-config` module and uses those to compile the plugin. So, regardless of the dependencies in `onos-config-model`, the dependencies used to compile plugins should always match the dependencies in the `onos-config` module. 

The problem is ensuring the compiler is fetching the right _version_ of the `onos-config` module to extract its dependencies. To test this change, the Helm chart must be configured to fetch the module info from this pull request. That's done by overriding the `plugin.compiler.target` value in the Helm chart to point to this branch:

```
helmit test -n test ./cmd/onos-config-tests --set onos-umbrella.onos-config.plugin.compiler.target=github.com/onosproject/onos-config@protocol-agent --suite gnmi
```

By default, the Helm chart is configured to compile plugins using a release tag like `github.com/onosproject/onos-config@v1.2.3`. This tells the compiler to download `go.mod` from the `protocol-agent` branch instead. When I ran the tests on this branch with the above command, they all passed.

I actually have a solution to eliminate the need to set the `plugin.compiler.target` altogether -- by building the compiler image with the current `onos-config` `go.mod` file already in it -- but hadn't had the time to implement it.

PR #1347, opened 2021-08-18
Bumping API dependency

tomikazi:
Fixing uses of GetAspect to reflect the new usage which is more consistent with the Go marshallers pattern.

### DIFF
--- a/pkg/device/device.go
+++ b/pkg/device/device.go
@@ -191,46 +191,34 @@ func ToDevice(object *topo.Object) (*Device, error) {
 		return nil, fmt.Errorf("topo entity %s must have a 'kindid' to work with onos-config", object.ID)
 	}
 
-	assetMsg := object.GetAspect(&topo.Asset{})
-	if assetMsg == nil {
-		return nil, errors.NewInvalid("topo entity %s must have 'asset' attribute to work with onos-config", object.ID)
-	}
-	asset, ok := assetMsg.(*topo.Asset)
-	if !ok || asset == nil {
-		return nil, errors.NewInternal("invalid asset object aspect %s", object.ID)
+	asset := &topo.Asset{}
+	err := object.GetAspect(asset)
+	if err != nil {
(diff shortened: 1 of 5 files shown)


tomikazi:
retest this please

pkg/device/device.go
```diff
@@ -191,46 +191,34 @@ func ToDevice(object *topo.Object) (*Device, error) {
...
-	mastershipMsg := object.GetAspect(&topo.MastershipState{})
-	if mastershipMsg == nil {
-		return nil, errors.NewInvalid("topo entity %s must have 'mastershipState' attribute to work with onos-config", object.ID)
-	}
-	mastership, ok := mastershipMsg.(*topo.MastershipState)
-	if !ok || mastership == nil {
-		return nil, errors.NewInternal("invalid mastershipState object aspect %s", object.ID)
+	mastership := &topo.MastershipState{}
+	err = object.GetAspect(mastership)
+	if err != nil {
```
adibrastegarnia, line 208: Here checks if mastershi.Term == 0 then we will know for sure it is not set.

pkg/device/device.go
```diff
@@ -191,46 +191,34 @@ func ToDevice(object *topo.Object) (*Device, error) {
...
+		return nil, errors.NewInvalid("topo entity %s must have 'onos.topo.TLSOptions' aspect to work with onos-config", object.ID)
 	}
 
-	protocolsMsg := object.GetAspect(&topo.Protocols{})
-	protocols, ok := protocolsMsg.(*topo.Protocols)
+	protocols := &topo.Protocols{}
+	err = object.GetAspect(protocols)
 	var protocolStates []*topo.ProtocolState
-	if !ok || protocols == nil {
+	if err != nil || protocols == nil {
```
adibrastegarnia, line 221: protocols == nil will be always false here because the pointer is nil and passing to the golang function doesn't necessarily tell you if it is not nil anymore.

adibrastegarnia (approved):

PR #1350, opened 2021-08-31
AETHER-2110 In Topo Entity to Device mapping - type should be taken from Configurable (not Kind)

SeanCondon:
In the [topo.yaml](https://github.com/onosproject/sdran-helm-charts/blob/master/aether-roc-umbrella/templates/topo.yaml) the `Entity` has a `Kind` which is given in all lowercase as `aether` - it has to be lowercase - nothing else will be accepted

The same `Entity` has a `Configurable` aspect which has a `Type` attribute, and this is titlecase `Aether`.

When `onos-config` synchronizes from `onos-topo` it converts this `Entity` in to a `Device`, but it sets the `Type` of the `Device` to the `Kind` value, but should be the `Configurable/Type` value, so that it can match the `Modelplugin` type i.e. `Aether`

### DIFF
--- a/pkg/device/device.go
+++ b/pkg/device/device.go
@@ -230,7 +230,7 @@ func ToDevice(object *topo.Object) (*Device, error) {
 		ID:          ID(object.ID),
 		Revision:    object.Revision,
 		Protocols:   protocolStates,
-		Type:        typeKindID,
+		Type:        Type(configurable.Type),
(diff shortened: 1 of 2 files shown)


adibrastegarnia (approved):

adibrastegarnia:
LGTM

tomikazi (approved):

PR #1351, opened 2021-09-01
AETHER-2109 Forbid the setting of empty strings through gNMI

SeanCondon:
[AETHER-2109](https://jira.opennetworking.org/browse/AETHER-2109)

Empty strings can be a problem to validate, and are unnecessary anyway. 

It is better to delete an attribute than to set its value to an empty string. If it is not mandatory it can be deleted.

If an attribute is mandatory - it can't be deleted - but neither is it valid to set it to an empty string.

In either case an empty string would be of no value.

I am only forbidding it on Set(). If it comes up from the device in OpState - that is OK (but pointless).

The reason it is difficult to validate is that with an empty string it will be left out of the JSON tree that's passed to YGOT.
When the attribute has a constraint like minLength > 0 this can mean that it's not getting validated properly.


sbconsulting (approved):
Strongly in favor, but also wary that there's some corner case we haven't thought of where someone wants to differentiate between an empty string and an unset string. For example, seems like there are valid cases for integers differentiating between "0" and nil.

SeanCondon:
> Strongly in favor, but also wary that there's some corner case we haven't thought of where someone wants to differentiate between an empty string and an unset string. For example, seems like there are valid cases for integers differentiating between "0" and nil.

For sure I was careful keep this to strings (and byte arrays) - users are free to delete integer attributes but cannot set them to nil in anycase

PR #1375, opened 2021-11-05
metadata precondition failed

SeanCondon:
This is giving me error when I'm trying to update the `NetworkChangeStore`
```
"message": " rpc error: code = Unknown desc = Error updating a3783c74-3e53-11ec-87f9-46f3d506bafe with username. metadata precondition failed"
```


ray-milkey (approved):

kuujo (changes requested):
This is not an avoidable error. It’s any optimistic lock failure that indicates the `NetworkChange` was updated by another goroutine between the `Get` and `Update', and it simply has to be retried until the change can be read and modified atomically. 

For example, suppose the change when it’s read here has `Incarnation=1`, and this code increments it to `Incarnation=2`. Well, this error indicates another goroutine (probably the controller) also read and updated the same change at the same time. Suppose that controller read the same change and already incremented the `Incarnation` number to 2. If Atomix then allowed the update here (also `Incarnation=2`) to be stored, well this code wouldn’t have actually incremented the incarnation number at all. These two concurrent updates on the same state can conflict with each other. The `Conflict` error that’s returned by Atomix here indicates that this update conflicted with another update and needs to be retried.

Here’s an example of how to retry the update from E2T:
https://github.com/onosproject/onos-e2t/blob/master/pkg/northbound/e2/v1beta1/subscription.go#L429-L457

pkg/northbound/gnmi/set.go
```diff
@@ -187,15 +188,19 @@ func (s *Server) Set(ctx context.Context, req *gnmi.SetRequest) (*gnmi.SetRespon
 		return nil, status.Error(codes.Internal, errSet.Error())
 	}
 	if userName != "" {
+		s.mu.Lock()
```
kuujo, line 191: This lock is not necessary since there’s no volatile state being updated and the Atomix stores are already thread safe.

pkg/northbound/gnmi/set.go
```diff
@@ -187,15 +188,19 @@ func (s *Server) Set(ctx context.Context, req *gnmi.SetRequest) (*gnmi.SetRespon
...
 		ch, err := mgr.NetworkChangesStore.Get(change.ID)
 		if err != nil {
-			return nil, status.Error(codes.Internal, errSet.Error())
+			s.mu.Unlock()
+			return nil, errors.NewInternal("Error getting NetworkChange %s: %s", change.ID, err.Error())
```
kuujo, line 195: FYI the reason this `Internal` error is causing an `Unknown` error is because this is a gRPC server and the error is not being converted for gRPC. If you do `errors.Status(err).Err()` you’ll get a gRPC error that will return the correct error code to the client.

kuujo: Also, keep in mind Atomix returns a lot more rich error information now — 'NotFound` when a key is not in the store, `AlreadyExists' when a key is already in the store, `Conflict` when there’s an update conflict, `Internal` when there’s an error inside the state machine, etc — and when we wrap errors in `NewInternal` we’re losing that type information. We really are trying to be more precise in which errors we want to propagate up to the client and which we can handle internally, as you can see in the E2T snippet above. That said, if this _is_ an error we want to propagate up to the client, I think calling it an `Internal` error is fine.

SeanCondon: good to know - thanks

SeanCondon:
> This is not an avoidable error

OK understood now - I was not clear on what was thread safe and what was not.

For what I am trying to do I think I will take the simpler approach of just updating the `NetworkChange` with `username` inside `NewNetworkChange()` before it is added to the store.

I have added your comments to my folder marked "Pearls of wisdom from Jordan" ;-)

PR #1381, opened 2021-11-16
Added topo store; mostly identical from E2T

tomikazi:


pkg/store/topo/store.go
```diff
@@ -0,0 +1,167 @@
...
+	if err != nil {
+		return nil, errors.FromGRPC(err)
+	}
+
+	return listResponse.Objects, nil
+}
+
+// Delete deletes topology object using the given ID
+func (s *targetStore) Delete(ctx context.Context, object *topoapi.Object) error {
+	_, err := s.client.Delete(ctx, &topoapi.DeleteRequest{
```
adibrastegarnia, line 131: would be good to add a log here as well

pkg/store/topo/store_test.go
```diff
@@ -0,0 +1,183 @@
...
+	"github.com/onosproject/onos-lib-go/pkg/certs"
+	topomgr "github.com/onosproject/onos-topo/pkg/manager"
+	"github.com/stretchr/testify/assert"
+	"io/ioutil"
+	"os"
+	"testing"
+)
+
+const (
+	caCrtFile = "/tmp/onos-proxy.cacrt"
```
adibrastegarnia, line 32: any reasons that we call it onos-proxy?

adibrastegarnia (approved):

pkg/store/topo/store_test.go
tomikazi: Ooops... borrowed test code from proxy and forgot to rename.

PR #1395, opened 2021-12-07
Refactor manager start and creation of stores

ray-milkey:


cmd/onos-config/onos-config.go
```diff
@@ -78,59 +67,31 @@ func main() {
...
-		log.Fatal("Cannot load device atomix store ", err)
-	}
-
-	networkChangesStore, err := network.NewAtomixStore(atomixClient)
-	if err != nil {
-		log.Fatal("Cannot load network atomix store ", err)
-	}
-
-	networkSnapshotStore, err := networksnap.NewAtomixStore(atomixClient)
+	modelRegistry, err := modelregistry.NewModelRegistry(modelregistry.Config{})
```
adibrastegarnia, line 70: we should be able to move this to manager as well correct?

adibrastegarnia (changes requested):

cmd/onos-config/onos-config.go
ray-milkey: Can't move this unfortunately. Model registry must be initialized before the manager is created. The tests have a special implementation of this.

ray-milkey: hmm no. I think I can move it now. Let me try

cmd/onos-config/onos-config.go
```diff
@@ -78,59 +66,25 @@ func main() {
...
-		log.Fatal("Cannot load network atomix store ", err)
-	}
-
-	deviceStateStore, err := state.NewStore(networkChangesStore, deviceSnapshotStore)
-	if err != nil {
-		log.Fatal("Cannot load device store with address %s:", *topoEndpoint, err)
-	}
-	log.Infof("Topology service connected with endpoint %s", *topoEndpoint)
+	mgr := manager.NewManager(cfg)
+	log.Info("Manager created")
```
adibrastegarnia, line 79: Perhaps we can remove this log as well and just log it inside Run function

adibrastegarnia (approved):

PR #1396, opened 2021-12-10
Initial sidecar plugin registry implementation

tomikazi:
Start of the implementation of the registry responsible for discovering sidecar model plugins.


pkg/pluginregistry/registry.go
```diff
@@ -0,0 +1,151 @@
...
+	"github.com/onosproject/onos-lib-go/pkg/logging"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+	"sync"
+)
+
+var log = logging.GetLogger("registry")
+
+// ModelPlugin is a record of information compiled from the configuration model plugin
+type ModelPlugin struct {
```
adibrastegarnia, line 34: If you look at the current registry, you will see there are some util functions for Read only/ReadWrite paths which is integrated with model registry + some stratum specefic code part of the loading the plugin. Somehow part of my PR I need those util functions (jsonvalues under model registry) but they should be independent of model registry package. I will try to define them somewhere else but we should compare current model registry plugin with the new one and see how they used to be integrated.

tomikazi: Yes, the RO/RW paths addition is on my list to do along with the xpath additions that Sean did recently on master. Unclear what to do with the Stratum-specific stuff however.

adibrastegarnia (approved):

PR #1397, opened 2021-12-12
Refactor northbound servers so manager can create them

ray-milkey:
This PR moves creation and startup of the northbound servers in onos-config to the manager. These changes break many unit tests because the current tests use the Manager class and that creates a circular dependency. This needs to be resolved. Also there is still code in the manager that provides gnmi specific functionality that needs to be moved out.


pkg/manager/manager.go
```diff
@@ -166,17 +173,79 @@ func (m *Manager) initializeStores() {
...
 
 	m.initializeStores()
 	log.Info("Stores initialized")
+
 	if err := m.Start(); err != nil {
 		log.Fatal("Unable to run Manager", err)
 	}
 
-	log.Info("Manager Started")
+	if err := m.startNBServers(); err != nil {
```
adibrastegarnia, line 245: I think it is better to move this call to Start function. main calls Run and inside Run we should have one Start function which initializes stores, starts controllers, NB server, etc

adibrastegarnia (changes requested):

pkg/manager/manager.go
```diff
@@ -166,17 +173,79 @@ func (m *Manager) initializeStores() {
...
+			AuthorizationEnabled:  authorization,
+		}))
+
+	s.AddService(logging.Service{})
+
+	adminService := admin.NewService(mgr.NetworkChangesStore, mgr.NetworkSnapshotStore, mgr.DeviceSnapshotStore)
+	s.AddService(adminService)
+
+	diagService := diags.NewService(mgr.DeviceChangesStore,
+		mgr.DeviceCache,
```
adibrastegarnia, line 201: Since the code is moving toward injecting deps, we should remove instances in the manager datas structure like this https://github.com/onosproject/onos-e2t/blob/master/pkg/manager/manager.go#L67. We don't need to define store instances in the manager structure and the instance can be injected to any package from Start function.

adibrastegarnia (changes requested):

pkg/manager/manager.go
ray-milkey: It is here because the tests call Start() and they don't want to bring up the servers.

pkg/manager/manager.go
ray-milkey: Let's do this once we figure out how to get the tests working again.

adibrastegarnia: Sure.

PR #1398, opened 2021-12-13
Add connection controller

adibrastegarnia:


ray-milkey:
retest this please

adibrastegarnia:
retest this please

kuujo (changes requested):
I think it looks pretty close overall. Just some cosmetic changes needed to clean up the code now and we should be good to go. But I think I need to take another look at the logic again tomorrow just to double check.

pkg/southbound/gnmi/conn.go
```diff
@@ -38,12 +44,20 @@ type ConnID string
 type Conn interface {
 	Client
 	ID() ConnID
+	TargetID() topoapi.ID
+	GetClientConn() *grpc.ClientConn
```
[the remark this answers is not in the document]
adibrastegarnia: I wanted to expose just state then I found out we also need Connect to trigger state changes when the Connection goes to IDLE.

pkg/southbound/gnmi/conn_event.go
```diff
@@ -0,0 +1,39 @@
...
+package gnmi
+
+// ConnEvent connection event
+type ConnEvent struct {
+	Conn      Conn
+	EventType ConnEventType
+}
+
+// ConnEventType a connection event type
+type ConnEventType int
```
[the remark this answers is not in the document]
adibrastegarnia: This is the event for adding/removing connections. I guess we need both. First we need to find out when a SB connection is added to initiate a go routine to listen to connection state changes otherwise we don't know that.

pkg/controller/controlrelation/controller.go
```diff
@@ -0,0 +1,170 @@
...
+			if !errors.IsNotFound(err) {
+				log.Warnf("Failed to reconcile control relation for connection %s, %v", conn.ID(), err)
+				return false, err
+			}
+			return false, nil
+		}
+		return r.reconcileDeleteControlRelation(ctx, controlRelation)
+	} else if currentState == connectivity.Idle {
+		log.Debugf("Reconnecting to target:", conn.TargetID())
+		conn.Connect()
```
kuujo, line 141: I’m not sure we should really care if the connection state is `IDLE`. It seems like a lot of extra overhead that we can get just by using keep-alive instead. But even without keep-alive, the connection state doesn’t really matter until we are interested in actually sending a request, and the act of sending the request will detect a failure if one has occurred. Mastership can change when that happens. I think we should remove the `Connect` method/calls for now.

pkg/controller/controlrelation/controller.go
```diff
@@ -0,0 +1,170 @@
...
+
+func (r *Reconciler) reconcileControlRelation(ctx context.Context, targetID topoapi.ID) (bool, error) {
+	conn, err := r.conns.Get(ctx, targetID)
+	if err != nil {
+		if !errors.IsNotFound(err) {
+			log.Warnf("Failed to reconcile control relation for target %s, %v", targetID, err)
+			return false, err
+		}
+		// if the connection is not found, find the associated control relation and delete it from topo
+		controlRelationFilter := getControlRelationFilter(utils.GetOnosConfigID(), targetID)
```
kuujo, line 92: In a multi-node cluster, this will cause the nodes to delete each other’s control relations. The controls relations should only be deleted directly by the associated source node, and only indirectly by other nodes when they delete the source entity.

pkg/southbound/gnmi/conn_manager.go
```diff
@@ -38,28 +36,27 @@ import (
 type ConnManager interface {
 	Get(ctx context.Context, targetID topoapi.ID) (Conn, error)
 	List(ctx context.Context) ([]Conn, error)
-	Watch(ctx context.Context, ch chan<- Conn) error
+	Watch(ctx context.Context, ch chan<- ConnEvent) error
 	Connect(ctx context.Context, target *topoapi.Object) (Conn, error)
+	Remove(ctx context.Context, targetID topoapi.ID) error
```
kuujo, line 41: I think if we’re going to call one method `Connect`, this method should be called `Close`. Alternatively, we could exclude this method and just remove it from the manager when the `Conn`’s `Close` method is called. Either way, the package should only export one method for closing/removing a connection — either Conn’s `Close` method or ConnManager’s `Close` method.

pkg/southbound/gnmi/conn_manager.go
```diff
@@ -106,15 +103,21 @@ func (m *connManager) connect(ctx context.Context, target *topoapi.Object) (Conn
...
 	m.conns[target.ID] = conn
-	m.clientConns[target.ID] = clientConn
 	m.connsMu.Unlock()
-	m.eventCh <- conn
+	m.eventCh <- ConnEvent{
+		Conn:      conn,
+		EventType: Connected,
+	}
 
+	go conn.processConnStateEvents()
```
kuujo, line 120: This should really be done in the `newConn` function since it’s an implementation detail of `conn`.

pkg/southbound/gnmi/conn_event.go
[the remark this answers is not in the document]
kuujo: Yeah you’re right they should be kept separate

pkg/southbound/gnmi/conn.go
```diff
@@ -38,12 +44,21 @@ type ConnID string
...
 }
 
 // conn gNMI Connection
 type conn struct {
 	*client
-	id ConnID
+	clientConn          *grpc.ClientConn
+	id                  ConnID
+	targetID            topoapi.ID
+	connStateWatchers   []chan<- connectivity.State
```
kuujo, line 59: I would just call this `stateWatchers`. `conn` is redundant since that’s the name of the type it’s in already.

kuujo (changes requested):

pkg/controller/controlrelation/controller.go
kuujo: I’m sorry, I saw the relations are being filtered for this source. Disregard.

pkg/controller/controlrelation/controller.go
```diff
@@ -0,0 +1,170 @@
...
+		controlRelations, err := r.topo.List(ctx, controlRelationFilter)
+		if err != nil {
+			log.Warnf("Failed to reconcile control relation for target %s, %v", targetID, err)
+			return false, err
+		}
+		if len(controlRelations) == 0 {
+			return false, nil
+		}
+
+		return r.reconcileDeleteControlRelation(ctx, &controlRelations[0])
```
kuujo, line 102: Assuming there can only be 0 or 1 control relations is a bad idea. We know the name of the relation we’re managing. There’s no need to list the relations at all. Just try to `Get` the relation. If it exists then delete it.

pkg/southbound/gnmi/conn.go
```diff
@@ -132,4 +147,66 @@ func (c *conn) ID() ConnID {
...
+		c.connStateWatchers = connStateWatchers
+		c.connStateWatchersMu.Unlock()
+	}()
+	return nil
+
+}
+
+func (c *conn) processConnStateEvents() {
+	log.Infof("Starting processing of connection state events for connection: %s", c.id)
+	go func() {
```
kuujo, line 188: This goroutine is redundant since this function is also being called in a new goroutine already.

pkg/controller/controlrelation/controller.go
adibrastegarnia: Removed for now

pkg/southbound/gnmi/conn_manager.go
adibrastegarnia: Renamed to Close

pkg/southbound/gnmi/conn_manager.go
adibrastegarnia: I added a newConn function to return an instance of conn struct and inside that function, it starts this goroutine.

pkg/southbound/gnmi/conn.go
adibrastegarnia: Renamed

pkg/southbound/gnmi/conn.go
adibrastegarnia: Good catch.

pkg/controller/controlrelation/controller.go
adibrastegarnia: Please check this part. I added a loop to delete a control relation once at a time (however, it is one) and return. if there is no relation then it returns false, nil at the end (I removed checking the condition which checks the number of relations is 0 to return false,nil)

pkg/controller/controlrelation/controller.go
```diff
@@ -0,0 +1,165 @@
...
+		// if the connection is not found, find the associated control relation and delete it from topo
+		controlRelationFilter := getControlRelationFilter(utils.GetOnosConfigID(), targetID)
+		controlRelations, err := r.topo.List(ctx, controlRelationFilter)
+		if err != nil {
+			log.Warnf("Failed to reconcile control relation for target %s, %v", targetID, err)
+			return false, err
+		}
+
+		for _, controlRelation := range controlRelations {
+			return r.reconcileDeleteControlRelation(ctx, &controlRelation)
```
kuujo, line 100: This is returning regardless of whether something changed or not. It should only return if something changed, otherwise it will never get to the second item in a list with length > 1

kuujo (changes requested):

pkg/controller/controlrelation/controller.go
```diff
@@ -0,0 +1,165 @@
...
+		}
+		return false, nil
+	}
+
+	currentState := conn.State()
+	if currentState == connectivity.Ready {
+		log.Infof("Creating control relation for connection %s", conn.ID())
+		// creates a control relation
```
kuujo, line 108: This comment probably doesn’t add any information of value.

pkg/controller/controlrelation/controller.go
```diff
@@ -0,0 +1,165 @@
...
+			if !errors.IsNotFound(err) {
+				log.Warnf("Failed to reconcile control relation for connection %s, %v", conn.ID(), err)
+				return false, err
+			}
+			return false, nil
+		}
+		return r.reconcileDeleteControlRelation(ctx, controlRelation)
+	}
+
+	return true, nil
```
kuujo, line 140: Is this right? `true` indicates a state change occurred.

kuujo (approved):

PR #1399, opened 2021-12-14
Remove saving stores in the manager struct

ray-milkey:


pkg/manager/manager.go
```diff
@@ -308,7 +217,61 @@ func (m *Manager) Start() error {
 		return err
 	}
 
-	return nil
+	// Creates gRPC server and registers various services; then serves.
+	authorization := false
```
adibrastegarnia, line 221: You can define a function like this https://github.com/onosproject/onos-e2t/blob/master/pkg/manager/manager.go#L148 to start NB server and calls it from manager. We will do the same thing for controllers and SB

pkg/manager/manager.go
```diff
@@ -73,27 +68,10 @@ type Config struct {
...
-	networkSnapshotController *controller.Controller
-	deviceSnapshotController  *controller.Controller
-	ModelRegistry             *modelregistry.ModelRegistry
-	TopoChannel               chan *topodevice.ListResponse
-	OperationalStateChannel   chan events.OperationalStateEvent
-	SouthboundErrorChan       chan events.DeviceResponse
-	Dispatcher                *dispatcher.Dispatcher
-	OperationalStateCache     map[topodevice.ID]devicechange.TypedValueMap
-	OperationalStateCacheLock *sync.RWMutex
+	SouthboundErrorChan     chan events.DeviceResponse
```
adibrastegarnia, line 71: How these channels used? If we disable tests, I guess we can define them locally for now as well and inject into other packages correct which need them?

pkg/manager/manager.go
```diff
@@ -111,191 +89,122 @@ func (m *Manager) setTargetGenerator(targetGen func() southbound.TargetIf) {
...
+	if err := m.Start(); err != nil {
+		log.Fatal("Unable to run Manager", err)
+	}
+}
+
+// Start starts the manager
+func (m *Manager) Start() error {
 	opts, err := certs.HandleCertPaths(m.Config.CAPath, m.Config.KeyPath, m.Config.CertPath, true)
 	if err != nil {
 		log.Fatal(err)
```
adibrastegarnia, line 105: We should return error and log here. We already call Fatal when we call m.Start()

pkg/manager/manager.go
```diff
@@ -111,191 +89,122 @@ func (m *Manager) setTargetGenerator(targetGen func() southbound.TargetIf) {
...
-	m.OperationalStateCacheLock = &sync.RWMutex{}
+	dsp := dispatcher.NewDispatcher()
+	operationalStateCache := make(map[topodevice.ID]devicechange.TypedValueMap)
+	operationalStateCacheLock := &sync.RWMutex{}
 
 	// Start the NetworkChange controller
-	errNetworkCtrl := m.networkChangeController.Start()
+	errNetworkCtrl := networkChangeController.Start()
 	if errNetworkCtrl != nil {
 		log.Error("Can't start controller ", errNetworkCtrl)
```
adibrastegarnia, line 176: Same thing here for any calls inside Start, returns error to the caller

adibrastegarnia (changes requested):

ray-milkey:
retest this please

pkg/manager/manager.go
ray-milkey: Must have been there for the tests. Nothing uses it now.

pkg/manager/manager.go
```diff
@@ -73,27 +68,9 @@ type Config struct {
...
-	networkSnapshotController *controller.Controller
-	deviceSnapshotController  *controller.Controller
-	ModelRegistry             *modelregistry.ModelRegistry
-	TopoChannel               chan *topodevice.ListResponse
-	OperationalStateChannel   chan events.OperationalStateEvent
-	SouthboundErrorChan       chan events.DeviceResponse
-	Dispatcher                *dispatcher.Dispatcher
-	OperationalStateCache     map[topodevice.ID]devicechange.TypedValueMap
-	OperationalStateCacheLock *sync.RWMutex
+	TopoChannel             chan *topodevice.ListResponse
```
adibrastegarnia, line 71: What about topoChannel and Operational state channel?

adibrastegarnia (changes requested):

adibrastegarnia (approved):

PR #1401, opened 2021-12-14
Restore admin tests

ray-milkey:


ray-milkey:
retest this please

adibrastegarnia (approved):

ray-milkey:
pleas retest this

ray-milkey:
retest this please

pkg/northbound/admin/admin_test.go
```diff
@@ -0,0 +1,182 @@
...
+	return conn, client, s, deviceSnapshotStore
+}
+
+func Test_RollbackNetworkChange_BadName(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	conn, client, server, _ := setUpServer(ctx, t)
+	defer server.Stop()
+	defer conn.Close()
```
kuujo, line 99: These deferrals are unnecessary since the setup function is already calling them when the context is closed. We should only have to cancel the context, so we can actually remove these lines from the test functions. Then, we may not need to return the client and server from the setup function either. Deleting code is always good, and I guess this is one of the benefits of the context pattern…

PR #1407, opened 2022-01-06
Initial implementation of configuration controller

adibrastegarnia:


pkg/controller/configuration/watcher.go
```diff
@@ -0,0 +1,132 @@
...
+		for event := range eventCh {
+			if _, ok := event.Object.Obj.(*topoapi.Object_Entity); ok {
+				err = event.Object.GetAspect(&topoapi.Configurable{})
+				if err == nil {
+					configurations, err := w.configurations.List(ctx)
+					if err != nil {
+						log.Warn(err)
+						continue
+					}
+					for _, config := range configurations {
```
adibrastegarnia, line 109: Here If we use any IDs for configuration (i.e. not target ID), we need to find all the configuration that is associated with the target ID of the event.

pkg/controller/configuration/controller.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+
+	if len(root.Notification) == 0 {
+		return false, errors.NewInvalid("notification list is empty")
+	}
+
+	var currentConfigValues []*configapi.PathValue
+	for _, notification := range root.Notification {
+		for _, update := range notification.Update {
+			configValues, err := jsonvalues.DecomposeJSONWithPaths("", update.GetVal().GetJsonIetfVal(), nil, nil)
```
adibrastegarnia, line 130: This utility function with a different name will be called  using the plugin interface

PR #1410, opened 2022-01-12
Add initial transaction controller

adibrastegarnia:


pkg/controller/transaction/controller.go
```diff
@@ -0,0 +1,228 @@
...
+	return controller.Result{}, nil
+}
+
+func (r *Reconciler) reconcileTransaction(ctx context.Context, transaction *configapi.Transaction) (bool, error) {
+
+	log.Debugf("Reconciling transaction %s in %s state", transaction.ID, transaction.Status.State)
+	// If the transaction is Pending, begin validation if the prior transaction
+	// has already been applied. This simplifies concurrency control in the controller
+	// and guarantees transactions are applied to the configurations in sequential order.
+	if transaction.Status.State == configapi.TransactionState_TRANSACTION_PENDING {
```
kuujo, line 105: I think we should just do a switch and call a method for each state:

```go
switch transaction.Status.State {
case configapi.TransactionState_TRANSACTION_PENDING:
    return r.reconcileTransactionPending(transaction)
…
}
```

pkg/controller/transaction/controller.go
```diff
@@ -0,0 +1,228 @@
...
+				if err != nil {
+					if !errors.IsConflict(err) || !errors.IsNotFound(err) {
+						return false, err
+					}
+					return false, nil
+				}
+
+				return true, nil
+			}
+		}
```
kuujo, line 127: There’s something missing here: this covers if the transaction index is > 1, but what if it’s 1? Then we can proceed to validate and apply it.

But also we probably shouldn’t use the index 1, but rather consider this the first transaction in the log if there’s no prior transaction. Once we introduce log compaction, the first index in the log will not necessarily be `1`.

pkg/controller/transaction/controller.go
```diff
@@ -0,0 +1,228 @@
...
+
```
kuujo, line 128: I think we can return `true` here once the transaction is moved to `VALIDATING`

pkg/controller/transaction/controller.go
```diff
@@ -0,0 +1,228 @@
...
+		} else {
+			for _, transactionChange := range transaction.Changes {
+				config, err := r.configurations.Get(ctx, transactionChange.TargetID)
+				if err != nil {
+					if !errors.IsNotFound(err) {
+						return false, err
+					}
+					return false, nil
+				}
+				if config.Index < transaction.Index {
```
kuujo, line 198: This condition is not needed any more since we’re applying transactions in strict sequential order. We know the transaction index will always be greater than the configuration’s index

pkg/controller/transaction/controller.go
```diff
@@ -0,0 +1,228 @@
...
+				}
+				config.Status.State = configapi.ConfigurationState_CONFIGURATION_PENDING
+				config.Status.TransactionIndex = transaction.Index
+				err = r.configurations.Update(ctx, config)
+				if err != nil {
+					if !errors.IsConflict(err) && !errors.IsNotFound(err) {
+						return false, err
+					}
+					return false, nil
+				}
```
kuujo, line 214: Return `true` here too. This will become unnecessary if we move each state into a separate method though

kuujo (changes requested):

pkg/controller/transaction/controller.go
```diff
@@ -0,0 +1,246 @@
...
+		// target and Complete the transaction.
+	case configapi.TransactionState_TRANSACTION_APPLYING:
+		return r.reconcileTransactionApplying(ctx, transaction)
+
+	}
+	return true, nil
+}
+
+func (r *Reconciler) reconcileTransactionPending(ctx context.Context, transaction *configapi.Transaction) (bool, error) {
+	if transaction.Index > 1 {
```
kuujo, line 121: Need to remove this condition. Checking the previous transaction is what we want, and this will prevent that from being done on the first transaction.

pkg/controller/transaction/controller.go
```diff
@@ -0,0 +1,246 @@
...
+		prevTransaction, err := r.transactions.GetByIndex(ctx, transaction.Index-1)
+		if err != nil {
+			if !errors.IsNotFound(err) {
+				return false, err
+			}
+			return false, nil
```
kuujo, line 127: We should not return on NotFound. If the previous transaction is not found then this is the first transaction in the log. Proceed to VALIDATING

adibrastegarnia:
retest this please

PR #1412, opened 2022-01-13
[AETHER-2815] Adding CLI config to use the new plugin registry, defau…

teone:
…lts to false

### DIFF
--- a/pkg/northbound/utils/setconfig.go
+++ b/pkg/northbound/utils/setconfig.go
@@ -35,9 +35,6 @@ import (
 // SetConfigAlreadyApplied is a string constant for "Already applied:"
 const SetConfigAlreadyApplied = "Already applied:"
 
-// FIXME: temporary flag to enable new plugin validation - setting to true will presently break unit tests
-var testNewPlugin = false
-
 // ValidateNetworkConfig validates the given updates and deletes, according to the path on the configuration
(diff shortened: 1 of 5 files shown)


pkg/northbound/utils/setconfig.go
```diff
@@ -35,23 +35,20 @@ import (
...
 	pluginRegistry *pluginregistry.PluginRegistry,
-	deviceStateStore state.Store, allowUnvalidatedConfig bool) error {
+	deviceStateStore state.Store, allowUnvalidatedConfig bool, usePluginRegistry bool) error {
 
 	modelName := utils.ToModelName(deviceType, version)
 
 	// TODO: permanently enable the new config plugin lookup when ready
 	var configPlugin *pluginregistry.ModelPlugin
-	if testNewPlugin {
+	if usePluginRegistry {
```
adibrastegarnia, line 51: We are going to remove this hopefully next week anyway but if you need it, go for it

teone: yes, this will be removed once we fully migrate to the new plugin registry, but in the meantime we won't have to rebuild the `onos-config` container to test them (that was the idea when discussing it with @tomikazi )

adibrastegarnia (approved):

teone:
retest this please

PR #1415, opened 2022-01-19
Adding ability to get model plugin capabilities and cache path maps

tomikazi:


pkg/pluginregistry/registry.go
```diff
@@ -102,6 +112,24 @@ func (r *PluginRegistry) discoverPlugin(port uint) {
 	log.Infof("Configuration model plugin %s discovered on port %d", plugin.ID, port)
 }
 
+func getRoPathMap(resp *api.ModelInfoResponse) path.ReadOnlyPathMap {
+	pm := make(map[string]path.ReadOnlySubPathMap)
+	for _, pe := range resp.ModelInfo.ReadOnlyPath {
+		// TODO: Implement conversion
```
tomikazi, line 118: I thought you had code that did this. It's not clear to me what the logic here should be... flattening map/map to array was easy, but if you have the code to reconstitute the map/map, there's no need to redo it.

adibrastegarnia: No I don't have but I don't think I need it as well at the moment. I think I just need RW paths.

adibrastegarnia (approved):

adibrastegarnia:
retest this please

adibrastegarnia:
retest this please

PR #1420, opened 2022-01-21
Topology test

ray-milkey:


ray-milkey:
retest this please

test/topo/topotest.go
```diff
@@ -0,0 +1,134 @@
...
+		relationObject topo.Object
+	)
+
+	// Get a topology API client
+	client, err := NewTopoClient()
+	assert.NoError(t, err)
+	assert.NotNil(t, client)
+
+	// Get the list of topo objects
+	objs, err := client.List(context.Background())
```
adibrastegarnia, line 56: You can use filters to get entities or relations rather than listing all of the objects.

test/topo/topotest.go
```diff
@@ -0,0 +1,134 @@
...
+
+	// Check onos-config aspects
+	assert.Len(t, configObject.Aspects, 1)
+	aspect, ok := configObject.Aspects["onos.topo.Lease"]
+	assert.True(t, ok)
+	assert.NotNil(t, aspect)
+
+	// Check the target simulator aspects
+	assert.Greater(t, len(targetObject.Aspects), 1)
+	aspect, ok = targetObject.Aspects["onos.topo.Configurable"]
```
adibrastegarnia, line 112: Each object has a method called GetAspect you should use because it returns error and if the error is not nil then the aspect is not set.

test/topo/topotest.go
adibrastegarnia: like this https://github.com/onosproject/onos-e2t/blob/master/test/utils/topo.go#L275

test/topo/topotest.go
```diff
@@ -0,0 +1,134 @@
...
+	"github.com/onosproject/onos-api/go/onos/topo"
+	"github.com/onosproject/onos-config/test/utils/gnmi"
+	toposdk "github.com/onosproject/onos-ric-sdk-go/pkg/topo"
+	"github.com/stretchr/testify/assert"
+	"testing"
+	"time"
+)
+
+// NewTopoClient creates a new topo SDK client
+func NewTopoClient() (toposdk.Client, error) {
```
adibrastegarnia, line 28: I am not we need a different suite for this test. Lets have a common util functions for topo like e2t that we can use it for other tests as well.  We can add this test to gnmi suite. Perhaps gnmi should be renamed to config tests to include tests related to Set/Get operations + topo integration.

adibrastegarnia: technically, this test tests connectivity in SB of onos-config + topo integration so we testing core of onos-config

test/topo/topotest.go
```diff
@@ -0,0 +1,134 @@
...
+	assert.True(t, ok)
+	assert.NotNil(t, aspect)
+	mastershipState := &topo.MastershipState{}
+	err = targetObject.GetAspect(mastershipState)
+	assert.NoError(t, err)
+	assert.Equal(t, relationObject.ID, topo.ID(mastershipState.NodeId))
+}
+
+// TestTopoEntries checks that the correct topology entities and relations are created
+func (s *TestSuite) TestTopoEntries(t *testing.T) {
```
adibrastegarnia, line 125: Perhaps we should give it a better name. Perhaps TestTopoIntegration

adibrastegarnia (approved):

test/gnmi/topotest.go
```diff
@@ -0,0 +1,181 @@
...
+
+// getRelationOrDie queries a single control relation between the config and target, and fails if not found
+func getRelationOrDie(ctx context.Context, t *testing.T, configID topo.ID, targetID topo.ID, c toposdk.Client) topo.Object {
+	filter := &topo.Filters{
+		RelationFilter: &topo.RelationFilter{
+			SrcId:    string(configID),
+			TargetId: string(targetID),
+			Scope:    topo.RelationFilterScope_ALL,
+		},
+		KindFilter: &topo.Filter{
```
tomikazi, line 66: I don't think you need this. Simply set RelationKind field int the RelationFilter to topo.CONTROLS.

PR #1422, opened 2022-01-21
NB implementation of the admin ConfigurationService

tomikazi:


pkg/manager/manager.go
```diff
@@ -133,7 +136,7 @@ func (m *Manager) startNorthboundServer(
 
 	s.AddService(logging.Service{})
 
-	adminService := admin.NewService(networkChangesStore, networkSnapshotStore, deviceSnapshotStore, pluginRegistry)
+	adminService := admin.NewService(transactionsStore, configurationsStore, pluginRegistry, networkChangesStore, networkSnapshotStore, deviceSnapshotStore)
```
adibrastegarnia, line 139: I think we should remove networkChangeStore, snaphort store, deviceStore from the arguments and clean up admin service.

pkg/manager/manager.go
```diff
@@ -278,6 +281,9 @@ func (m *Manager) Start() error {
 		return err
 	}
 
+	// TODO: placeholder for creation of the transactions store
+	var transactions transaction.Store
```
adibrastegarnia, line 285: Go ahead and create a store like Configuration store and start it. You don't need to put a place holder.

pkg/northbound/admin/configuration.go
```diff
@@ -0,0 +1,84 @@
...
+		if err != nil {
+			log.Warnf("ListConfigurations %+v failed: %v", req, err)
+			return errors.Status(err).Err()
+		}
+	}
+	return nil
+}
+
+// WatchConfigurations provides stream with events representing configuration changes
+func (s Server) WatchConfigurations(req *admin.WatchConfigurationsRequest, stream admin.ConfigurationService_WatchConfigurationsServer) error {
```
adibrastegarnia, line 55: Why some of the server instances are pointer some of them are not?

adibrastegarnia (changes requested):

pkg/manager/manager.go
adibrastegarnia: I will rebase

pkg/northbound/admin/configuration.go
tomikazi: Fixed.

pkg/manager/manager.go
tomikazi: Cleaned up as much as possible at this point in time. Left notes in other places.

pkg/manager/manager.go
tomikazi: Done.

pkg/northbound/admin/transaction.go
```diff
@@ -0,0 +1,84 @@
...
+		watchOpts = append(watchOpts, transaction.WithReplay())
+	}
+
+	ch := make(chan configapi.TransactionEvent)
+	if err := s.transactionsStore.Watch(stream.Context(), ch, watchOpts...); err != nil {
+		log.Warnf("WatchTransactionsRequest %+v failed: %v", req, err)
+		return errors.Status(err).Err()
+	}
+
+	return s.streamTransactions(stream, ch)
```
adibrastegarnia, line 68: I think we should return the error here and convert it to status before returning it back.

pkg/northbound/admin/configuration.go
```diff
@@ -0,0 +1,84 @@
...
+		watchOpts = append(watchOpts, configuration.WithReplay())
+	}
+
+	ch := make(chan configapi.ConfigurationEvent)
+	if err := s.configurationsStore.Watch(stream.Context(), ch, watchOpts...); err != nil {
+		log.Warnf("WatchConfigurationsRequest %+v failed: %v", req, err)
+		return errors.Status(err).Err()
+	}
+
+	return s.streamConfigurations(stream, ch)
```
adibrastegarnia, line 68: I think we should return the error here and convert it to status before returning it back.

adibrastegarnia (changes requested):

adibrastegarnia (approved):

pkg/northbound/admin/transaction.go
tomikazi: Fixed

pkg/northbound/admin/configuration.go
tomikazi: Fixed

PR #1424, opened 2022-01-22
[SDRAN-1426] Fixing the Plugin Registry

teone:
Fixing the Plugin Registry lookup method so that we can discover a plugin from a gNMI Set request.

Plugins are now stored using the `Name` (as provided via CLI params) as the Index, but the plugin ID is generated from the `ModelInfo` `Name` and `Version` so that we can use the correct plugin to validate a gNMI request.


pkg/pluginregistry/registry.go
```diff
@@ -139,14 +150,15 @@ func (r *PluginRegistry) discoverPlugin(endpoint string) {
 	}
 	plugin.Status = loaded
 	plugin.Info = *resp.ModelInfo
+	plugin.ID = strings.ToLower(fmt.Sprintf("%s-%s", resp.ModelInfo.Name, resp.ModelInfo.Version))
```
tomikazi, line 154: I think that once we find the plugin, we should delete the entry under plugin.Name and re-add under plugin.ID. The name is nothing but a provisional ID.

pkg/pluginregistry/registry.go
```diff
@@ -204,11 +216,18 @@ func newClient(port uint) (api.ModelPluginServiceClient, error) {
...
 // GetPlugin returns the plugin with the specified ID
+// NOTE this method might get slow if we have a lot of plugins loaded. Do we see that as a possibility?
+// If so we might need to re-work the registry and index the plugins using the ID while using an autogenerated ID
+// for the ones that fail to load (eg: unknown-fee754)
 func (r *PluginRegistry) GetPlugin(id string) (*ModelPlugin, bool) {
 	r.lock.RLock()
 	defer r.lock.RUnlock()
-	p, ok := r.plugins[id]
-	return p, ok
+	for _, p := range r.plugins {
```
tomikazi, line 225: We can continue to do a direct lookup if we do the swap on discovery.

pkg/pluginregistry/registry.go
teone: I was going down a similar path yesterday,
thinking that we can ditch the `Port` field in favor of an `Endpoint` field (might get useful if we decide to move away from Sidecars, as no changes would be required), and use this `Endpoint` field as `ID` before we successfully load the plugin info, then update id with the received information.

adibrastegarnia:
Go mod tidy to pass the build ?

adibrastegarnia (approved):

PR #1426, opened 2022-01-23
Fix transaction controller to ensure changes are applied to configuration

kuujo:


pkg/controller/transaction/controller.go
```diff
@@ -86,30 +86,38 @@ func (r *Reconciler) Reconcile(id controller.ID) (controller.Result, error) {
...
+		transaction.Status.State == configapi.TransactionState_TRANSACTION_FAILED {
+		nextTransaction, err := r.transactions.GetByIndex(ctx, transaction.Index+1)
+		if err != nil {
+			if !errors.IsNotFound(err) {
+				return controller.Result{}, err
+			}
 			return controller.Result{}, nil
 		}
+		log.Debugf("Queueing next transaction %d", nextTransaction.Index)
+		return controller.Result{Requeue: controller.ID{Value: nextTransaction.ID}}, nil
```
kuujo, line 101: If one transaction is complete, we should always reconcile the next transaction in case completing this transaction unblocked the next one.

pkg/controller/transaction/controller.go
```diff
@@ -202,54 +208,78 @@ func (r *Reconciler) reconcileTransactionChangeValidating(ctx context.Context, t
...
-			config.Status.TransactionIndex = transaction.Index
-			err = r.configurations.Update(ctx, config)
-			if err != nil {
-				if !errors.IsConflict(err) && !errors.IsNotFound(err) {
-					return false, err
-				}
-				return false, nil
+	// The source configurations must be stored prior to updating the target configurations
+	// otherwise they will be lost.
+	err := r.transactions.Update(ctx, transaction)
```
kuujo, line 240: We have to be really careful to make sure the information we’re storing about the source we’re changing is stored before we begin updating the target configurations. Otherwise, if we update a configuration and then crash, when we recover from the crash we’d be storing the updated target configuration as the source.

adibrastegarnia: Makes sense.

adibrastegarnia (approved):

PR #1428, opened 2022-01-24
Refactoring of integration tests for terminology changes

ray-milkey:


test/config/createremovetargettest.go
```diff
@@ -104,17 +104,17 @@ func (s *TestSuite) TestCreatedRemovedDevice(t *testing.T) {
...
 	// Wait for config to connect to the device
-	gnmi.WaitForDeviceAvailable(t, createRemoveDeviceModDeviceName, 2*time.Minute)
+	gnmi.WaitForTargetAvailable(t, createRemoveDeviceModDeviceName, 2*time.Minute)
 
 	// Check that the value was set correctly
 	gnmi.CheckGNMIValue(t, c, devicePath, createRemoveDeviceModValue2, 0, "Query after set 2 returns wrong value")
 
 	// Check that the network change has completed
-	gnmi.WaitForNetworkChangeComplete(t, networkChangeID2, 10*time.Second)
+	gnmi.WaitForTransactionComplete(t, networkChangeID2, 10*time.Second)
```
adibrastegarnia, line 113: there is no network change ID thing. it will be transaction ID

test/config/offlinetargettest.go
```diff
@@ -51,19 +51,19 @@ func (s *TestSuite) TestOfflineDevice(t *testing.T) {
...
+	devicePath := gnmi.GetTargetPathWithValue(offlineDeviceName, modPath, modValue, proto.StringVal)
 	networkChangeID := gnmi.SetGNMIValueOrFail(t, gnmiClient, devicePath, gnmi.NoPaths, extensions)
 
 	// Bring device online
 	simulator := gnmi.CreateSimulatorWithName(t, offlineDeviceName)
 	defer gnmi.DeleteSimulator(t, simulator)
 
 	// Wait for config to connect to the device
-	gnmi.WaitForDeviceAvailable(t, device.ID(simulator.Name()), time.Minute)
+	gnmi.WaitForTargetAvailable(t, device.ID(simulator.Name()), time.Minute)
```
adibrastegarnia, line 62: there is no device ID anymore. Under onos-api/go/v2 you will be find these types that you can use for the tests: https://github.com/onosproject/onos-api/blob/master/go/onos/config/v2/types.go#L53

adibrastegarnia (changes requested):

test/config/multiplesettest.go
```diff
@@ -50,7 +50,7 @@ func (s *TestSuite) TestMultipleSet(t *testing.T) {
 
 		msValue := generateTimezoneName()
 
-		devicePath := gnmi.GetDevicePathWithValue(simulator.Name(), tzPath, msValue, proto.StringVal)
+		devicePath := gnmi.GetTargetPathWithValue(simulator.Name(), tzPath, msValue, proto.StringVal)
```
adibrastegarnia, line 53: Perhaps change the name of variables as well to target to avoid using two different terminology

adibrastegarnia (changes requested):

test/config/offlinetargettest.go
ray-milkey: These are still using the "device shim" and will be changed when that is eliminated.

adibrastegarnia: I think you can go ahead and update it  unless you want to do it separately which is fine.

PR #1429, opened 2022-01-24
Fix bugs in controllers, NB, etc 

adibrastegarnia:


pkg/controller/configuration/controller.go
```diff
@@ -89,6 +89,38 @@ func (r *Reconciler) Reconcile(id controller.ID) (controller.Result, error) {
...
+			return controller.Result{}, nil
+		}
+
+		configurableAspect := &topoapi.Configurable{}
+		err = target.GetAspect(configurableAspect)
+		if err != nil {
+			return controller.Result{}, err
+		}
+
+		initConfig := &configapi.Configuration{
```
adibrastegarnia, line 107: I think we should create an initial configuration entry when the target is added otherwise there is no way to find the config in the transaction controller to update source config in transaction.

kuujo: Yeah this is definitely something I overlooked... but I think this should actually be done by the transaction controller if it does not find a `Configuration` when applying a `Transaction`.

kuujo (changes requested):

pkg/controller/configuration/controller.go
```diff
@@ -118,6 +150,10 @@ func (r *Reconciler) reconcileConfiguration(ctx context.Context, config *configa
 	_ = target.GetAspect(&mastership)
 	targetMastershipTerm := configapi.MastershipTerm(mastership.Term)
 
+	if config.Status.State == configapi.ConfigurationState_CONFIGURATION_COMPLETE {
```
kuujo, line 153: Should this be `|| config.Status.State == configapi.ConfigurationState_CONFIGURATION_FAILED`?

kuujo (approved):

PR #1436, opened 2022-01-25
Use v2 of the onos-config NB API in tests

ray-milkey:


test/config/multiplesettest.go
```diff
@@ -44,18 +41,19 @@ func (s *TestSuite) TestMultipleSet(t *testing.T) {
 	// Make a GNMI client to use for requests
 	gnmiClient := gnmi.GetGNMIClientOrFail(t)
 
-	var transactionIDs []network.ID
+	//var transactionIDs []network.ID
```
adibrastegarnia, line 44: Do you want to remove these ?

adibrastegarnia (approved):

PR #1437, opened 2022-01-26
Update onos-api dependency to use maps for transaction targets/changes

kuujo:


pkg/controller/configuration/controller.go
```diff
@@ -241,9 +241,9 @@ func (r *Reconciler) reconcileConfiguration(ctx context.Context, config *configa
 
 		desiredConfigValues := config.Values
 		log.Debugf("Desired config values to reconcile: %v", desiredConfigValues)
-		for _, desiredConfigValue := range desiredConfigValues {
+		for path, desiredConfigValue := range desiredConfigValues {
```
adibrastegarnia, line 244: I think the key also can be path instead of desiredConfigValue.Path

adibrastegarnia (approved):

PR #1445, opened 2022-01-26
Add missing watchers to configuration controller

kuujo:
Because reconciliation in the `Configuration` controller depends on the master relation and connection, the configuration controller should be triggered on changes to controls relations and connections.


kuujo:
@adibrastegarnia be sure to double check my code please... I'm not sure I did it right

adibrastegarnia (approved):

PR #1450, opened 2022-01-27
Changed targetID to configurationID for configuration lookup

tomikazi:


pkg/northbound/admin/configuration.go
```diff
@@ -25,8 +25,7 @@ import (
 // GetConfiguration returns response with the requested configuration
 func (s Server) GetConfiguration(ctx context.Context, req *admin.GetConfigurationRequest) (*admin.GetConfigurationResponse, error) {
 	log.Infof("Received GetConfiguration request: %+v", req)
-	// TODO: Add target type and version to configuration ID
-	conf, err := s.configurationsStore.Get(ctx, configuration.NewID(req.TargetID, "", ""))
+	conf, err := s.configurationsStore.Get(ctx, req.ConfigurationID)
```
adibrastegarnia, line 28: Add Watch option for WithConfigurationID thing if in the request is available

adibrastegarnia (changes requested):

pkg/northbound/admin/configuration.go
adibrastegarnia: There is no replay, we should add one more

adibrastegarnia (approved):

PR #1452, opened 2022-01-27
fix tests, clean up 

adibrastegarnia:


test/config/transactiontest.go
```diff
@@ -66,24 +67,17 @@ func (s *TestSuite) TestTransaction(t *testing.T) {
 
 	// Set initial values
 	targetPathsForInit := gnmi.GetTargetPathsWithValues(targets, paths, initialValues)
-	initialTransactionID, initialTransactionIndex := gnmi.SetGNMIValueOrFail(t, gnmiClient, targetPathsForInit, gnmi.NoPaths, gnmi.NoExtensions)
-	complete := gnmi.WaitForTransactionComplete(t, initialTransactionID, initialTransactionIndex, 10*time.Second)
-	assert.True(t, complete, "Set never completed")
+	gnmi.SetGNMIValueOrFail(t, gnmiClient, targetPathsForInit, gnmi.NoPaths, gnmi.NoExtensions)
```
ray-milkey, line 70: Is there ever a need for the transaction ID and Index now? Can we change the signature of this method to remove them?

tomikazi: I would keep the return values. We may just not need them in the tests; although it seems it should be `_, _ = ...`

adibrastegarnia: ok. let me do that.

tomikazi (approved):

PR #1453, opened 2022-01-27
Apply validation on the whole config tree path/values

adibrastegarnia:


ray-milkey (approved):

pkg/controller/transaction/controller.go
```diff
@@ -171,14 +171,31 @@ func (r *Reconciler) reconcileTransactionChangeValidating(ctx context.Context, t
...
 		for path, changeValue := range change.Values {
-			pathValue := &configapi.PathValue{
+			currentConfigValues[path] = &configapi.PathValue{
 				Path:    path,
 				Value:   changeValue.Value,
 				Deleted: changeValue.Delete,
-				Index:   transaction.Index,
 			}
+		}
+		for _, pathValue := range currentConfigValues {
```
kuujo, line 198: I would move the definition of `pathValues` down here

kuujo (changes requested):

pkg/controller/transaction/controller.go
```diff
@@ -171,14 +171,31 @@ func (r *Reconciler) reconcileTransactionChangeValidating(ctx context.Context, t
...
+		config, err := r.configurations.Get(ctx, configID)
+		if err != nil {
+			if !errors.IsNotFound(err) {
+				log.Errorf("Failed applying Transaction %d to target '%s'", transaction.Index, targetID, err)
+				return controller.Result{}, err
+			}
+		}
 		pathValues := make([]*configapi.PathValue, 0, len(change.Values))
+		currentConfigValues := make(map[string]*configapi.PathValue)
+		if config.GetValues() != nil {
```
kuujo, line 187: I think you can/should just use `config.Values`. The `GetValues()` method is really just there so you can define interfaces for your Protobuf generated types to implement.

pkg/controller/transaction/controller.go
```diff
@@ -171,14 +171,31 @@ func (r *Reconciler) reconcileTransactionChangeValidating(ctx context.Context, t
...
+			currentConfigValues = config.Values
+		}
+
 		for path, changeValue := range change.Values {
-			pathValue := &configapi.PathValue{
+			currentConfigValues[path] = &configapi.PathValue{
```
kuujo, line 192: I think this should just be called `configValues` because it’s not just current values… it’s the current values plus we’re adding the new ones here.

kuujo: More descriptive might be to call this `pathValues` (since it’s a path to value mapping) and the other one just `values`.

kuujo (approved):

PR #1455, opened 2022-01-27
Fixed TestTransaction and TestDeleteAndRollback

tomikazi:


test/config/deletetest.go
```diff
@@ -68,12 +72,13 @@ func (s *TestSuite) TestDelete(t *testing.T) {
 	// Now rollback the change
 	adminClient, err := gnmi.NewAdminServiceClient()
 	assert.NoError(t, err)
-	rollbackResponse, rollbackError := adminClient.RollbackNetworkChange(
-		context.Background(), &admin.RollbackRequest{Name: string(transactionID)})
+	rollbackResponse, rollbackError := adminClient.RollbackNetworkChange(context.Background(), &admin.RollbackRequest{Index: transactionIndex})
```
adibrastegarnia, line 75: Is it possible to rename this API to rollbackTransaction?

adibrastegarnia (changes requested):

test/utils/gnmi/utils.go
```diff
@@ -277,6 +277,41 @@ func WaitForConfigurationCompleteOrFail(t *testing.T, configurationID configapi.
...
+		resp, err := stream.Recv()
+		if err != nil {
+			return false
+		}
+		assert.NotNil(t, resp)
+		fmt.Printf("%v\n", resp.TransactionEvent)
+
+		t := resp.TransactionEvent.Transaction
+		if rt := t.GetRollback(); rt != nil {
+			if rt.Index == transactionIndex {
```
adibrastegarnia, line 304: A rollback transaction can go through different stages ( like a normal transaction. What status this function is interested in?

adibrastegarnia: You can also watch a specefic transaction rather than comparing it. Send WithTransactionID to the watch. Look at WatchConfiguration as example.

adibrastegarnia (approved):

PR #1461, opened 2022-01-29
update failure status in transaction controller

adibrastegarnia:


pkg/controller/transaction/controller.go
```diff
@@ -382,6 +386,10 @@ func (r *Reconciler) reconcileTransactionRollbackValidating(ctx context.Context,
 		}
 		log.Warnf("Rollback Transaction %d failed: target Transaction %d not found", transaction.Index, rollback.Index)
 		transaction.Status.State = configapi.TransactionState_TRANSACTION_FAILED
+		transaction.Status.Failure = &configapi.Failure{
+			Type:        configapi.Failure_UNAVAILABLE,
```
kuujo, line 390: Actually, this should be `NotFound`. `Unavailable` will be indistinguishable from a connection failure by the client.

PR #1470, opened 2022-01-30
Add supported models to the capabilities response

ray-milkey:


pkg/northbound/gnmi/v2/service.go
```diff
@@ -79,15 +79,17 @@ type Server struct {
...
-	// TODO adds capability
-	/*capabilities, err := s.modelRegistry.Capabilities()
-	if err != nil {
-		return nil, err
-	}*/
+	plugins := s.pluginRegistry.GetPlugins()
+
+	supportedModels := make([]*gnmi.ModelData, 0)
+	for _, plugin := range plugins {
+		capabilities := plugin.Capabilities(ctx, nil)
```
adibrastegarnia, line 86: What is the second argument here which can be nil?

ray-milkey: Not sure what the intention is - in the function itself the parameter is unused

adibrastegarnia: OK perhaps we should look into what it was before and remove or complete it if it is needed. Going to approve it that you can merge.

adibrastegarnia (approved):

pkg/northbound/gnmi/v2/service.go
ray-milkey: OK. I'll find out of we can remove that parameter before I merge it.

tomikazi (approved):

PR #1472, opened 2022-01-31
Simplifying tests by relying on sync transaction mode extension

tomikazi:


test/config/multiplesettest.go
```diff
@@ -50,20 +46,14 @@ func (s *TestSuite) TestMultipleSet(t *testing.T) {
...
 
 		// Remove the path we added
-		gnmi.SetGNMIValueOrFail(t, gnmiClient, gnmi.NoPaths, targetPath, gnmi.NoExtensions)
-
-		err = gnmi.WaitForConfigurationCompleteOrFail(t, configapi.ConfigurationID(simulator.Name()), time.Minute)
-		assert.NoError(t, err)
+		gnmi.SetGNMIValueOrFail(t, gnmiClient, gnmi.NoPaths, targetPath, gnmi.SyncExtension(t))
 
 		//  Make sure it got removed
 		gnmi.CheckGNMIValue(t, gnmiClient, targetPath, "", 0, "incorrect value found for path /system/clock/config/timezone-name after delete")
```
adibrastegarnia, line 59: It is not clear to me this function is checking onos-config or target. If it is onos-config is good but perhaps we should check the target as well to make sure the value is updated properly.

test/config/offlinetargettest.go
```diff
@@ -47,7 +46,7 @@ func (s *TestSuite) TestOfflineTarget(t *testing.T) {
 
 	// Sends a set request using onos-config NB
 	targetPath := gnmiutils.GetTargetPathWithValue(offlineTargetName, modPath, modValue, proto.StringVal)
-	gnmiutils.SetGNMIValueOrFail(t, gnmiClient, targetPath, gnmiutils.NoPaths, []*gnmi_ext.Extension{})
+	gnmiutils.SetGNMIValueOrFail(t, gnmiClient, targetPath, gnmiutils.NoPaths, gnmiutils.NoExtensions)
```
adibrastegarnia, line 49: For all of the sets, we should enable sync to make sure it is applied before checking the value on target if this test is doing it.

test/config/singlepathtest.go
```diff
@@ -43,19 +38,13 @@ func (s *TestSuite) TestSinglePath(t *testing.T) {
...
-	assert.NoError(t, err)
+	gnmi.SetGNMIValueOrFail(t, gnmiClient, devicePath, gnmi.NoPaths, gnmi.SyncExtension(t))
 
 	// Check that the value was set correctly
 	gnmi.CheckGNMIValue(t, gnmiClient, devicePath, tzValue, 0, "Query after set returned the wrong value")
 
 	// Remove the path we added
-	gnmi.SetGNMIValueOrFail(t, gnmiClient, gnmi.NoPaths, devicePath, gnmi.NoExtensions)
-
-	err = gnmi.WaitForConfigurationCompleteOrFail(t, configapi.ConfigurationID(simulator.Name()), time.Minute)
```
adibrastegarnia, line 57: If the function is still in utils we should remove it.

test/config/singlepathtest.go
```diff
@@ -43,19 +38,13 @@ func (s *TestSuite) TestSinglePath(t *testing.T) {
 	devicePath := gnmi.GetTargetPathWithValue(simulator.Name(), tzPath, tzValue, proto.StringVal)
 
 	// Set a value using gNMI client
-	gnmi.SetGNMIValueOrFail(t, gnmiClient, devicePath, gnmi.NoPaths, gnmi.NoExtensions)
-
-	err := gnmi.WaitForConfigurationCompleteOrFail(t, configapi.ConfigurationID(simulator.Name()), time.Minute)
-	assert.NoError(t, err)
+	gnmi.SetGNMIValueOrFail(t, gnmiClient, devicePath, gnmi.NoPaths, gnmi.SyncExtension(t))
```
adibrastegarnia, line 41: this import in some of the tests called gnmi in some others gnmiutils. So I think we should use gnmiutils to be more clear.

adibrastegarnia (changes requested):

test/config/multiplesettest.go
ray-milkey: This test does not interrogate the target

ray-milkey: That would be a good thing to add.

adibrastegarnia: My point was we should no? It is a multi set thing and we should make sure it is applied to the target eventually. if onos-config has the change doesn't mean it is on the target.

tomikazi:
Let me address checking the target vs checking the config in a separate PR. I think this stage of clean-up is good enough to stand on its own.

adibrastegarnia (approved):

PR #1474, opened 2022-02-01
Adding more checks to tests

tomikazi:
This adds checks to validate that the target was properly affected. Note that delay is required; otherwise the checks will fail. This was marked as FIXME.

### DIFF
--- a/test/config/multiplesettest.go
+++ b/test/config/multiplesettest.go
@@ -20,6 +20,7 @@ import (
 	"github.com/onosproject/onos-config/test/utils/proto"
 	"github.com/stretchr/testify/assert"
 	"testing"
+	"time"
 )
 
 func generateTimezoneName() string {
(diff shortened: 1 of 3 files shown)


test/config/multiplesettest.go
```diff
@@ -39,23 +40,31 @@ func (s *TestSuite) TestMultipleSet(t *testing.T) {
...
+
 		// Remove the path we added
 		gnmiutils.SetGNMIValueOrFail(t, gnmiClient, gnmiutils.NoPaths, targetPath, gnmiutils.SyncExtension(t))
 
-		//  Make sure it got removed
+		//  Make sure it got removed, both from onos-config and the target
 		gnmiutils.CheckGNMIValue(t, gnmiClient, targetPath, "", 0, "incorrect value found for path /system/clock/config/timezone-name after delete")
+
+		// FIXME: Allow some time for target to commit the change; this should not be necessary
+		time.Sleep(2 * time.Second)
```
adibrastegarnia, line 67: Perhaps you check the status of configuration when it gets to this point

PR #1476, opened 2022-02-02
Fix NB Get problems

adibrastegarnia:


pkg/northbound/gnmi/v2/get.go
```diff
@@ -191,5 +231,10 @@ func (s *Server) checkOpaAllowed(ctx context.Context, targetInfo targetInfo, con
 	}
 
 	log.Debugf("body text of response from OPA:\n%s", bodyText)
+	modelPlugin, err := s.getModelPlugin(ctx, topoapi.ID(targetInfo.targetID))
```
adibrastegarnia, line 234: I am wondering If store an instance of model plugin in target info as well to avoid this call.

PR #1477, opened 2022-02-02
Add onos-config restart tests

ray-milkey:


test/config/configrestarttest.go
```diff
@@ -0,0 +1,84 @@
...
+	// Make a GNMI client to use for onos-config requests
+	gnmiClient := gnmiutils.GetGNMIClientWithContextOrFail(context.Background(), t)
+
+	targetPath := gnmiutils.GetTargetPathWithValue(simulator.Name(), restartTzPath, restartTzValue, proto.StringVal)
+
+	// Set a value using onos-config
+	gnmiutils.SetGNMIValueOrFail(t, gnmiClient, targetPath, gnmiutils.NoPaths, gnmiutils.SyncExtension(t))
+
+	// Check that the value was set correctly
+	gnmiutils.CheckGNMIValue(t, gnmiClient, targetPath, restartTzValue, 0, "Query after set returned the wrong value")
```
adibrastegarnia, line 45: Can we check the target is also has the value. If the sync extension works then it should be good to have this check for all the tests.

test/config/configrestarttest.go
```diff
@@ -0,0 +1,84 @@
...
+	simulator := gnmiutils.CreateSimulator(t)
+	defer gnmiutils.DeleteSimulator(t, simulator)
+
+	// Make a GNMI client to use for onos-config requests
+	gnmiClient := gnmiutils.GetGNMIClientWithContextOrFail(context.Background(), t)
+
+	targetPath := gnmiutils.GetTargetPathWithValue(simulator.Name(), restartTzPath, restartTzValue, proto.StringVal)
+
+	// Restart onos-config
+	configPod := hautils.FindPodWithPrefix(t, "onos-config")
```
adibrastegarnia, line 72: Since we run tests on a single instance, this doesn't make a difference but it is better to find master relation for the target that we are interested in from mastership state in target entity and use the source of that relation in order to find onos-config pod and delete it.

ray-milkey: Will address this at a later time.

PR #1481, opened 2022-02-03
Moving the plugin ID computation in the registry package

teone:


pkg/utils/idUtils.go
```diff
@@ -19,19 +19,8 @@ import (
...
-	return fmt.Sprintf("%s-%s", name, version)
-}
-
-// ToModelName simply joins together model name and version in a consistent way
-func ToModelName(name devicetype.Type, version devicetype.Version) string {
-	return fmt.Sprintf("%s-%s", name, version)
-}
-
 // ToConfigName simply joins together device ID and version in a consistent way
 func ToConfigName(deviceID devicetype.ID, version devicetype.Version) string {
```
adibrastegarnia, line 25: We can remove this as well.

adibrastegarnia (changes requested):

pkg/utils/idUtils.go
teone: Done

adibrastegarnia (approved):

PR #1482, opened 2022-02-03
Converted ModelPlugin to interface for testability; added mock entiti…

tomikazi:
…es to get tests


pkg/northbound/gnmi/v2/get.go
```diff
@@ -154,8 +154,8 @@ func (s *Server) checkOpaAllowed(ctx context.Context, targetInfo targetInfo, con
 		return nil, err
 	}
 
-	targetVersion := filterTargetForURL(modelPlugin.Info.Version)
-	targetType := filterTargetForURL(modelPlugin.Info.Name)
+	targetVersion := filterTargetForURL(modelPlugin.GetInfo().Info.Version)
```
adibrastegarnia, line 157: I think this function somehow redundant isn't it? GetInfo then info?

pkg/northbound/gnmi/v2/get_test.go
```diff
@@ -31,16 +31,18 @@ import (
...
 	return &Server{
 		mu:             sync.RWMutex{},
-		pluginRegistry: gnmitest.NewMockPluginRegistry(mctl),
-		topo:           gnmitest.NewMockStore(mctl),
+		pluginRegistry: registryMock,
+		topo:           topoMock,
 		transactions:   txStore,
 		configurations: cfgStore,
-	}, mctl, test
+	}, mctl, test, topoMock, registryMock
```
adibrastegarnia, line 45: I would suggest to define an struct and include all these mocks and information that should be returned to the caller.

teone (approved):

pkg/northbound/gnmi/v2/get.go
tomikazi: No... GetInfo returns ModelPluginInfo, which is a record (previously named ModelPlugin) which contains another structure defined by API called ModelInfo. Yes, it looks somewhat odd, but I think we should leave it as is.

pkg/northbound/gnmi/v2/get_test.go
tomikazi: Yeah, I was already working towards that in the current PR.

PR #1483, opened 2022-02-03
Updating tests to use the correct targetType for new plugins

teone:


Makefile
```diff
@@ -21,7 +21,7 @@ jenkins-test: build deps license_check linters
 	TEST_PACKAGES=github.com/onosproject/onos-config/... ./build/build-tools/build/jenkins/make-unit
 
 helmit-config: integration-test-namespace # @HELP run helmit gnmi tests locally
-	helmit test -n test ./cmd/onos-config-tests --suite config
+	helmit test -n test ./cmd/onos-config-tests --suite config -c ../../onos-helm-charts
```
adibrastegarnia, line 24: I think for integration tests we use the published charts. Any reason we should use local charts in the make file?

test/utils/charts/umbrella.go
```diff
@@ -16,12 +16,13 @@ package charts
 
 import (
 	"github.com/onosproject/helmit/pkg/helm"
-	"github.com/onosproject/onos-test/pkg/onostest"
 )
 
+const localRepo = "/home/teone/Sites/onos-helm-charts"
```
adibrastegarnia, line 21: Yeah you should remove this

Makefile
teone: my bad, didn't intend to commit that

test/utils/charts/umbrella.go
teone: as for the other comment

adibrastegarnia (approved):

PR #1488, opened 2022-02-04
Using locally vendored modules in the build

teone:
This is an hybrid approach in which `modules` are not committed in the repo, but downloaded on the host file-system and then mounted in the container.

* Modules are not stored in the repo, they are downloaded before the Docker build (or the regular build)
* After modules are downloaded, if LOCAL_PROTOS is set, the onos-api library is replaced with a local copy
* Only the files required for the build are mounted in the Docker container to optimize the cache usage

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,16 +18,11 @@ mod-lint: mod-update # @HELP ensure that the required dependencies are in place
 	# dependencies are vendored, but not committed, go.sum is the only thing we need to check
 	bash -c "diff -u <(echo -n) <(git diff go.sum)"
 
-local-protos: # @HELP Copies a local version of the onos-api dependency into the vendor directory
-ifdef LOCAL_PROTOS
-	rm -rf vendor/github.com/onosproject/onos-api/go
-	mkdir -p vendor/github.com/onosproject/onos-api/go
-	cp -r ${LOCAL_PROTOS}/go/* vendor/github.com/onosproject/onos-api/go
-	rm -rf vendor/github.com/onosproject/onos-api/go/vendor
-endif
+local-deps: # @HELP imports local deps in the vendor folder
+local-deps: local-helmit local-onos-api local-onos-lib-go local-onos-ric-sdk-go local-onos-test local-onos-topo
 
 build: # @HELP build the Go binaries and run all validations (default)
-build: mod-update local-protos
+build: mod-update local-deps
 	go build -mod=vendor -o build/_output/onos-config ./cmd/onos-config
 
 test: # @HELP run the unit tests and source code validation producing a golang style report
@@ -46,7 +41,7 @@ helmit-rbac: integration-test-namespace # @HELP run helmit gnmi tests locally
 
 integration-tests: helmit-config helmit-rbac # @HELP run helmit integration tests locally
 
-onos-config-docker: mod-update local-protos # @HELP build onos-config base Docker image
+onos-config-docker: mod-update local-deps # @HELP build onos-config base Docker image
 	docker build . -f build/onos-config/Dockerfile \
 		-t ${DOCKER_REPOSITORY}onos-config:${ONOS_CONFIG_VERSION}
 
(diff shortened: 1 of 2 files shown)


tomikazi (commented):
This looks OK to me. The only concern I have is that this represents a departure from what we do in the other repos. Should this be done for the other repos as well? It would be good to get comments from Ray and Jordan here.

teone:
I've been thinking a bit about this lately, and looks like every project use different standards... I wonder if it's the case to have a wider discussion and standardize things...

tomikazi:
> I've been thinking a bit about this lately, and looks like every project use different standards... I wonder if it's the case to have a wider discussion and standardize things...

I'm less concerned about different projects, and more about consistency between repos in the same project at the moment. I'm not opposed to the wider standardization, but that can veer into a territory with conflicting priorities. Having the different repos in the same project do things in the same manner should be quite doable in my opinion.

teone:
@tomikazi It's definitely easier than a cross-project solution!

I'll let this PR sit over the weekend to get feedback from @ray-milkey @adibrastegarnia and @kuujo, better make sure we're in agreement before making changes to multiple repos...

Makefile
```diff
@@ -8,16 +10,32 @@ ONOS_CONFIG_VERSION ?= latest
...
 
+mod-update: # @HELP Download the dependencies to the vendor folder
+	go mod tidy
+	go mod vendor
+
+mod-lint: mod-update # @HELP ensure that the required dependencies are in place
+	# dependencies are vendored, but not committed, go.sum is the only thing we need to check
+	bash -c "diff -u <(echo -n) <(git diff go.sum)"
+
+local-protos: # @HELP Copies a local version of the onos-api dependency into the vendor directory
```
adibrastegarnia, line 21: What about other deps? I mean can we replace onos-lib-go, sdk, etc without any problems with this PR or it is just onos-api

teone: As of now there is support for `onos-api` only, let me update to support the others as well

adibrastegarnia:
retest this please

Makefile
```diff
@@ -61,4 +73,44 @@ clean:: # @HELP remove all the build artifacts
...
+endif
+
+local-onos-lib-go: # @HELP Copies a local version of the onos-lib-go dependency into the vendor directory
+ifdef LOCAL_ONOS_LIB_GO
+	rm -rf vendor/github.com/onosproject/onos-lib-go
+	mkdir -p vendor/github.com/onosproject/onos-lib-go
+	cp -r ${LOCAL_ONOS_LIB_GO}/* vendor/github.com/onosproject/onos-lib-go
+endif
+
+local-onos-ric-sdk-go: # @HELP Copies a local version of the onos-ric-sdk-go dependency into the vendor directory
```
adibrastegarnia, line 97: I am not sure this is a good idea if we want to be consistent in all repos in this organization. The problem is different repos have different dep to different subsystems and each of those dep can be replaced. Previously, it was not needed to us any of these things to get local dep working.

teone: That's consistent with the way VOLTHA handles that: https://github.com/opencord/voltha-go/blob/master/Makefile#L70-L84

It's not required to have those "helpers" in the makefile, as long as the local `vendor` folder is mounted in the container the dependency replacement will work. I found that useful in the past so that the workflow looks like:

```
// make changes in the onos-api repo (wherever it is on the local fs)

// build the onos-config container wherever those changes are
LOCAL_ONOS_API=<path-to-onos-api-repo> make kind
```
If everything works commit the changes in the `onos-api` repo, no copy/paste needed from the `vendor` folder, once release update `go.mod` in `onos-config`, run `make mod-update` and commit.

adibrastegarnia: Good. If those helpers are optional and vendor folder is already taken care of that replace thing then it is great. I thought we just support those deps for local replace.

teone: Ah no, my bad for not explaining :)

You can replace anything you want in the `vendor` folder and it will work fine, these helpers are good for dependencies we manage as often we work on both in parallel, this way you don't have to copy them around manually (reducing the margin for errors).

If we adopt this approach on all the repos and you're testing a change that spans Config, Topo and CLI, you can simply run the same `LOCAL_ONOS_API=<path-to-onos-api-repo> make kind` in the 3 repos to have the image with your changes in `onos-api`

teone:
@tomikazi @adibrastegarnia @ray-milkey @kuujo are we ok with merging this?

ray-milkey (approved):

adibrastegarnia (approved):

PR #1489, opened 2022-02-04
Don't use retries on the client fir RBAC tests

ray-milkey:


teone (approved):
Looks good to me, but someone else more hands on should confirm

tomikazi (approved):

PR #1492, opened 2022-02-05
Add Proposal controller

kuujo:
This PR adds a new `Proposal` controller for managing the transactional changes/rollbacks associated with a specific target. The `Transaction` controller creates a `Proposal` for each target it affects, and `Proposal`s for each target are processed in transaction order. Both `Transaction`s and `Proposal`s progress through four phases:
* initialize - sets up the `Transaction` and `Proposal`s, e.g. linking neighboring `Proposal`s for navigation
* validate - validates the `Tansaction` and its `Proposal`s using RBAC and model plugins to determine whether the `Transaction` can be committed
* commit - commits the changes/rollbacks to associated target `Configuration`s
* apply - applies changes/rollbacks to targets via gNMI `Set`

The `Configuration` controller is no longer responsible for synchronizing changes as they're committed to the `Transaction` log. Instead, the `Configuration` controller is only responsible for recovering applied changes on ephemeral targets after a restart.


kuujo:
retest this please

adibrastegarnia (approved):

SeanCondon:
> This PR adds...

This is a great description - I hope this will be added to the documentation

kuujo:
@SeanCondon I'm working on the documentation right now

SeanCondon:
> 

:+1:

PR #1497, opened 2022-02-07
reconcile abort phase in transaction and proposal controller

adibrastegarnia:


pkg/controller/transaction/controller.go
```diff
@@ -510,6 +513,56 @@ func (r *Reconciler) reconcileCommit(ctx context.Context, transaction *configapi
...
+			transaction.Status.State = configapi.TransactionStatus_PENDING
+			transaction.Status.Phases.Abort.State = configapi.TransactionAbortPhase_ABORTED
+			transaction.Status.Phases.Abort.End = getCurrentTimestamp()
+			if err := r.updateTransactionStatus(ctx, transaction); err != nil {
+				return controller.Result{}, err
+			}
+			return controller.Result{}, nil
+		}
+
+	case configapi.TransactionAbortPhase_ABORTED:
```
kuujo, line 560: I think we can just exclude this state.

pkg/controller/proposal/controller.go
```diff
@@ -421,6 +399,55 @@ func (r *Reconciler) reconcileValidate(ctx context.Context, proposal *configapi.
...
+				log.Errorf("Failed reconciling Transaction %d Proposal to target '%s'", proposal.TransactionIndex, proposal.TargetID, err)
+				return controller.Result{}, err
+			}
+			proposal.Status.Phases.Abort.End = getCurrentTimestamp()
+			proposal.Status.Phases.Abort.State = configapi.ProposalAbortPhase_ABORTED
+			if err := r.updateProposalStatus(ctx, proposal); err != nil {
+				return controller.Result{}, err
+			}
+			return controller.Result{}, nil
+		} else if config.Status.Committed.Index == proposal.Status.PrevIndex &&
```
kuujo, line 433: This should be the first condition, otherwise we're always going to (unnecessarily) update the configuration twice.

PR #1501, opened 2022-02-09
Support NB synchronous Get 

adibrastegarnia:


pkg/northbound/gnmi/v2/get.go
```diff
@@ -124,6 +134,30 @@ func (s *Server) Get(ctx context.Context, req *gnmi.GetRequest) (*gnmi.GetRespon
 		}
 	}
 
+	if transactionStrategy.Synchronicity.String() == configapi.TransactionStrategy_SYNCHRONOUS.String() {
```
adibrastegarnia, line 137: should I use switch case here?

PR #1508, opened 2022-02-11
Refactor client creation code

ray-milkey:
- No need to query k8s to find config nodes, use the name onos-config and let DNS find it
- Consolidate and share code to create connections to onos-config
- Rename client creation functions to make it clearer what they are doing


ray-milkey:
retest this please

adibrastegarnia (approved):
That is great

PR #1511, opened 2022-02-11
[AETHER-3138] Supporting multiple gNMI extensions

teone:
It is possible that the client sets multiple gNMI extensions in a request.
Iterate over all of them to find the value.

**NOTE**
As of now when looking for a

### DIFF
--- a/pkg/northbound/gnmi/v2/extensions.go
+++ b/pkg/northbound/gnmi/v2/extensions.go
@@ -25,29 +25,44 @@ import (
 
 var log = logging.GetLogger("northbound", "gnmi")
 
-func processExtension(ext *gnmi_ext.Extension) (configapi.TransactionStrategy, error) {
-	var transactionStrategy configapi.TransactionStrategy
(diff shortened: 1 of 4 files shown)


pkg/northbound/gnmi/v2/extensions.go
```diff
@@ -25,29 +25,44 @@ import (
...
+			// otherwise just return the values
+			return regExt.RegisteredExt.Msg, nil
 		}
 	}
-	return transactionStrategy, nil
+	return extType, nil
 }
 
-func getExtensions(request interface{}) (configapi.TransactionStrategy, error) {
+func getTransactionStrategy(request interface{}) (*configapi.TransactionStrategy, error) {
```
adibrastegarnia, line 49: I think we can just return the value. I don't think we need a pointer to that since we don't want to change anything.

adibrastegarnia (approved):

pkg/northbound/gnmi/v2/extensions.go
teone: good point, done

SeanCondon (approved):

PR #1512, opened 2022-02-12
Add test for deleted paths

ray-milkey:
Added a check that when a path is deleted, it doesn't show up in queries anymore.

### DIFF
--- a/test/config/treepathtest.go
+++ b/test/config/treepathtest.go
@@ -40,7 +40,7 @@ func (s *TestSuite) TestTreePath(t *testing.T) {
 
 	// Make a simulated device
 	simulator := gnmiutils.CreateSimulator(ctx, t)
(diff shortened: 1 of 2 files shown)


test/config/treepathtest.go
```diff
@@ -70,4 +74,13 @@ func (s *TestSuite) TestTreePath(t *testing.T) {
 
 	//  Make sure new root got removed
 	gnmiutils.CheckGNMIValue(ctx, t, gnmiClient, getPath, gnmiutils.NoExtensions, "", 0, "New root was not removed")
+
+	// Make sure path got removed
+	interfacesPath := gnmiutils.GetTargetPath(simulator.Name(), newConfigPath)
```
SeanCondon, line 79: I think you should be checking the root path - to see if this interface got deleted
In gnmi it's like having no elem in the query - so I guess in test it would be like the path `/`

PR #1516, opened 2022-02-14
Implementing PrunePathValues function

tomikazi:


adibrastegarnia (approved):

teone (approved):

tomikazi:
retest this please

pkg/controller/proposal/controller.go
```diff
@@ -480,6 +480,8 @@ func (r *Reconciler) reconcileCommit(ctx context.Context, proposal *configapi.Pr
 			for path, changeValue := range changeValues {
 				config.Values[path] = changeValue
 			}
+			config.Values = tree.PrunePathMap(config.Values, true)
```
adibrastegarnia, line 483: I think you can operate on change values first before putting it in the config values?

tomikazi: I'm not sure I can. The change values could contain just the delta no? We want the prune the entire config as a whole.

adibrastegarnia: that is right,

kuujo (approved):

PR #1517, opened 2022-02-15
Fix Get state test race condition 

adibrastegarnia:


pkg/northbound/gnmi/v2/get.go
```diff
@@ -222,7 +222,7 @@ func (s *Server) processStateOrOperationalRequest(ctx context.Context, req *gnmi
 
 		conn, err := s.conns.GetByTarget(ctx, topoapi.ID(targetID))
 		if err != nil {
-			return nil, errors.Status(err).Err()
```
kuujo, line 225: We should check for the specific `NotFound` error here rather than just returning `Unavailable` for every possible error. What if we decide to refactor the connection manager and add another type of error to it, one that shouldn’t necessarily be returned as an `Unavailable` error to the client?

kuujo (approved):

pkg/northbound/gnmi/v2/get.go
kuujo: Actually, why was `return nil, errors.Status(err).Err()` being returned as an `Internal` error anyways?

PR #1520, opened 2022-02-15
Refactor tests to use new set and get utilities

ray-milkey:


benchmark/gnmi/getbench.go
```diff
@@ -27,6 +27,14 @@ func (s *BenchmarkSuite) BenchmarkGet(b *benchmark.Benchmark) error {
...
 	ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
 	defer cancel()
-	_, _, err := gnmiutils.GetGNMIValue(ctx, s.client, devicePath, gnmiutils.NoExtensions, gbp.Encoding_PROTO)
+
+	var onosConfigGetReq = &gnmiutils.GetRequest{
+		Ctx:      ctx,
+		Client:   s.client,
+		Paths:    devicePath,
+		Encoding: gbp.Encoding_PROTO,
+		DataType: gbp.GetRequest_CONFIG,
```
adibrastegarnia, line 36: For now, lets don't specify this to avoid confusion. The default is ALL which is fine.

test/config/createremovetargettest.go
```diff
@@ -16,6 +16,7 @@
 package config
 
 import (
+	protognmi "github.com/openconfig/gnmi/proto/gnmi"
```
adibrastegarnia, line 19: you can also call this gnmiapi.

test/config/updatedeletetest.go
```diff
@@ -17,6 +17,7 @@ package config
 import (
 	gnmiutils "github.com/onosproject/onos-config/test/utils/gnmi"
 	"github.com/onosproject/onos-config/test/utils/proto"
+	gbp "github.com/openconfig/gnmi/proto/gnmi"
```
adibrastegarnia, line 20: we should try to use the same alias for this gnmi thing. I guess gnmiapi should be good.

test/config/createremovetargettest.go
```diff
@@ -64,7 +85,9 @@ func (s *TestSuite) TestCreatedRemovedTarget(t *testing.T) {
 	// Set a value using gNMI client - target is down
 	setPath2 := gnmiutils.GetTargetPathWithValue(createRemoveTargetModTargetName, createRemoveTargetModPath, createRemoveTargetModValue2, proto.StringVal)
 
-	_, _ = gnmiutils.SetGNMIValueOrFail(ctx, t, c, setPath2, gnmiutils.NoPaths, gnmiutils.NoExtensions)
+	setReq.UpdatePaths = setPath2
+	setReq.Extensions = nil
```
adibrastegarnia, line 89: I think the default should be nil so you don't need to set it to nil.

test/config/modelstest.go
```diff
@@ -15,6 +15,7 @@
 package config
 
 import (
+	gbp "github.com/openconfig/gnmi/proto/gnmi"
```
adibrastegarnia, line 18: same thing here. consistent naming.

test/config/createremovetargettest.go
ray-milkey: setReq is reused, so the value is overwritten

PR #1522, opened 2022-02-15
[SDRAN-1521] WIP Testing ExtractPaths method

teone:
Starting to look at the `ExtractPaths` method.

There is something that does not adds up in a basic test I added: `TestExtractPathsNested`
```
// read-only paths
/cont1a/list2a[name=*]
	/name: [IsAKey: false, ValueType: STRING, TypeOpts: [], Description: ]
	
// read-write paths
 /cont1a/list2a[name=*]/name
	[IsAKey: true, ValueType: STRING, TypeOpts: [], Description: ]
```

It seems strange to me that the same path is both `read-only` and `read-write`. My understanding is that it should be a `rw` path, is that correct @SeanCondon ?

### DIFF
--- a/pkg/utils/path/path.go
+++ b/pkg/utils/path/path.go
@@ -52,11 +52,29 @@ type ReadOnlyAttrib struct {
 	AttrName    string
 }
 
+func (ro ReadOnlyAttrib) String() string {
+	// TODO add all fields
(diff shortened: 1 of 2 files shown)


gab-arrobo:
@teone, is this PR still WIP or should it be closed?

PR #1523, opened 2022-02-16
Implement caching for configuration store

adibrastegarnia:


pkg/store/configuration/watchers.go
```diff
@@ -0,0 +1,69 @@
...
+// newWatchers creates watchers
+func newWatchers() *watchers {
+	return &watchers{
+		watchers: make(map[uuid.UUID]watcher),
+	}
+}
+
+// sendAll sends a configuration event for all registered watchers
+func (ws *watchers) sendAll(event configapi.ConfigurationEvent) {
+	ws.rm.RLock()
```
kuujo, line 34: This read lock will have no effect because the actual read is being don’t in a different goroutine. It will be released before the memory is even read inside the new goroutine. The lock needs to be acquired by the goroutine that’s reading the memory.

pkg/store/configuration/configuration.go
```diff
@@ -258,42 +275,99 @@ func (s *configurationStore) List(ctx context.Context) ([]*configapi.Configurati
...
+		for event := range eventCh {
+			ch <- event
+		}
 	}()
+
+	go func() {
+		<-ctx.Done()
+		err := s.watchers.remove(watcherID)
+		if err != nil {
+			return
```
kuujo, line 325: If we return here then the channel will never be closed.

pkg/store/configuration/watchers.go
```diff
@@ -0,0 +1,69 @@
...
+
+}
+
+// remove removes a watcher
+func (ws *watchers) remove(id uuid.UUID) error {
+	ws.rm.Lock()
+	watchers := make(map[uuid.UUID]watcher, len(ws.watchers)-1)
+	for watcherID, watcher := range ws.watchers {
+		if watcherID != id {
+			watchers[id] = watcher
```
kuujo, line 62: This is unnecessary. `watchers` is already a map so just acquire the write lock and add the channel to the map. Also this condition basically means if the first watcher ID does not match the UUID then add the channel to the map.

pkg/store/configuration/configuration.go
```diff
@@ -70,13 +73,27 @@ func NewAtomixStore(client atomix.Client) (Store, error) {
...
+
+	go configStore.watchConfigurationEvent(eventCh)
+	return configStore, nil
 }
 
 type configurationStore struct {
 	configurations _map.Map
+	watchers       *watchers
+	cache          map[configapi.ConfigurationID]configapi.Configuration
+	watchersMu     sync.RWMutex
```
kuujo, line 95: I don’t think this mutex is being used, is it?

kuujo (changes requested):

pkg/store/configuration/configuration.go
```diff
@@ -258,42 +275,99 @@ func (s *configurationStore) List(ctx context.Context) ([]*configapi.Configurati
...
+		}
 	}()
+
+	go func() {
+		<-ctx.Done()
+		err := s.watchers.remove(watcherID)
+		if err != nil {
+			return
+		}
+		close(ch)
```
kuujo, line 327: Actually, we should not close this channel here. Since this channel is dependent on an upstream channel (`eventCh`), it should be closed after the upstream channel is closed. Otherwise, `ch` can be closed before `eventCh` is read, and the for loop reading `eventCh` will try to write to a closed channel and cause a panic.

pkg/store/configuration/watchers.go
```diff
@@ -0,0 +1,69 @@
...
+func newWatchers() *watchers {
+	return &watchers{
+		watchers: make(map[uuid.UUID]watcher),
+	}
+}
+
+// sendAll sends a configuration event for all registered watchers
+func (ws *watchers) sendAll(event configapi.ConfigurationEvent) {
+	ws.rm.RLock()
+	go func() {
```
kuujo, line 35: We need to avoid creating a new goroutine for each event because it can cause the events to be reordered. Even though the data structure does not guarantee any order across keys, events _are_ ordered with respect to each key. Two concurrent updates to the same key will produce two events, and the events will be received in the order in which those updates were applied to the store. Handling those events on two different goroutines can cause them to be reordered.

kuujo:
Superseded by #1533

PR #1525, opened 2022-02-16
Re-instating ability to retrieve list of all target IDs via path.Targ…

tomikazi:
…et=*


pkg/northbound/gnmi/v2/get.go
```diff
@@ -344,3 +349,47 @@ func (s *Server) checkOpaAllowed(ctx context.Context, targetInfo *targetInfo, co
...
+
+	// Distill the list of configurable entities into their corresponding target IDs
+	targets := make([]string, 0, len(targetEntities))
+	for _, targetEntity := range targetEntities {
+		targets = append(targets, string(targetEntity.ID))
+	}
+
+	// Produce an appropriately encoded path value with all target IDs
+	var allDevicesPathElem = make([]*gnmi.PathElem, 0)
+	allDevicesPathElem = append(allDevicesPathElem, &gnmi.PathElem{Name: "all-devices"})
```
adibrastegarnia, line 371: Should we use target terminology instead of device terminology ?

tomikazi: Yes, we should! Good catch! :)

adibrastegarnia (approved):

PR #1529, opened 2022-02-17
Added HA crashed target integration test

tomikazi:


tomikazi:
This still has a small race condition where the test may check the target configuration before onos-config has had a chance to fully re-apply it after the target comes back up anew.

test/config/crashedtarget.go
```diff
@@ -0,0 +1,114 @@
...
+	checkTarget(ctx, t, target, targetPathsForGet)
+
+	// Kill the target simulator
+	gnmiutils.DeleteSimulator(t, target)
+
+	// Re-create the target simulator with the same name and wait for it to become available
+	target = gnmiutils.CreateSimulatorWithName(ctx, t, target.Name(), false)
+	defer gnmiutils.DeleteSimulator(t, target)
+	gnmiutils.WaitForTargetAvailable(ctx, t, topo.ID(target.Name()), time.Minute)
+
```
adibrastegarnia, line 96: From the time that ransim is available until onos-config applies the change, it takes some time. Perhaps a Get on NB of onos-config with SyncExtension helps to make sure it is applied before checking the target.

test/config/crashedtarget.go
```diff
@@ -0,0 +1,114 @@
...
+	gnmiapi "github.com/openconfig/gnmi/proto/gnmi"
+	"testing"
+	"time"
+
+	"github.com/onosproject/onos-api/go/onos/topo"
+
+	gnmiutils "github.com/onosproject/onos-config/test/utils/gnmi"
+)
+
+const (
```
ray-milkey, line 30: Because of go scoping rules, these constants and variables will end up in the namespace of every test. Putting these inside the test function is the best approach. If that's not feasible, then name these so they are unique for the test.

adibrastegarnia (approved):

ray-milkey (approved):

PR #1530, opened 2022-02-17
Fix deadlock and race condition in synchronous Get

kuujo:
This PR fixes a race condition and a potential deadlock in the NB API's handling of `SYNCHRONOUS` `Get`s. I will comment on some of the code changes to explain...

### DIFF
--- a/pkg/northbound/gnmi/v2/get.go
+++ b/pkg/northbound/gnmi/v2/get.go
@@ -171,21 +171,34 @@ func (s *Server) processRequest(ctx context.Context, req *gnmi.GetRequest, group
 				ch := make(chan configapi.ConfigurationEvent)
 				err := s.configurations.Watch(ctx, ch, configuration.WithConfigurationID(configuration.NewID(target.targetID)), configuration.WithReplay())
 				if err != nil {
-					return nil, err
+					return nil, errors.Status(err).Err()
 				}
 				wg.Add(1)
-				go func(id configapi.ConfigurationID) {
+				go func(target *targetInfo) {
+					defer wg.Done()
 					for event := range ch {
 						if event.Configuration.Status.Applied.Index >= target.configuration.Status.Committed.Index &&
 							event.Configuration.Status.Applied.Term == event.Configuration.Status.Term {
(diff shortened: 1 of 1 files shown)


pkg/northbound/gnmi/v2/get.go
```diff
@@ -171,21 +171,34 @@ func (s *Server) processRequest(ctx context.Context, req *gnmi.GetRequest, group
...
+					defer wg.Done()
 					for event := range ch {
 						if event.Configuration.Status.Applied.Index >= target.configuration.Status.Committed.Index &&
 							event.Configuration.Status.Applied.Term == event.Configuration.Status.Term {
-							wg.Done()
 							return
 						}
 					}
-				}(target.configuration.ID)
+				}(target)
```
kuujo, line 185: Because of the way Go's `for` loops work, we have to pass the target info into the goroutine, otherwise the `target.configuration.Status.Committed.Index` used in the condition above will basically just use the last target in the loop. I doubt this would have worked with multiple targets.

pkg/northbound/gnmi/v2/get.go
```diff
@@ -171,21 +171,34 @@ func (s *Server) processRequest(ctx context.Context, req *gnmi.GetRequest, group
...
+		doneCh := make(chan struct{})
+		go func() {
+			wg.Wait()
+			close(doneCh)
+		}()
+
+		// If the context is canceled by the client, return the context error.
+		select {
+		case <-doneCh:
+		case <-ctx.Done():
```
kuujo, line 199: We have to account for the scenario in which the client canceled the context while waiting for the configuration to be propagated, e.g. due to a timeout.

pkg/northbound/gnmi/v2/get.go
adibrastegarnia: yep good catch.

kuujo:
I'm running the integration tests. Once they all pass I think we can merge this when you're ready @adibrastegarnia

adibrastegarnia (approved):

kuujo:
passed

PR #1534, opened 2022-02-18
Implement caching and watch multiplexing in proposal store

kuujo:
This PR implements caching and watch multiplexing for the `Proposal` store using the same patterns used for the `Transaction` store in #1532


kuujo:
recheck this please

PR #1544, opened 2022-02-24
Addressing delete/create issue

tomikazi:


pkg/controller/proposal/controller.go
```diff
@@ -509,6 +513,22 @@ func (r *Reconciler) reconcileCommit(ctx context.Context, proposal *configapi.Pr
 	}
 }
 
+func applyChangeToConfig(values map[string]*configapi.PathValue, path string, value *configapi.PathValue) (string, *configapi.PathValue) {
+	values[path] = value
```
kuujo, line 517: What’s the purpose of setting this key here?

pkg/controller/proposal/controller.go
```diff
@@ -509,6 +513,22 @@ func (r *Reconciler) reconcileCommit(ctx context.Context, proposal *configapi.Pr
...
+
+	// Walk up the path and make sure that there are no parents marked as deleted in the given map, if so, remove them
+	parent := pathutils.GetParentPath(path)
+	for len(parent) > 0 {
```
kuujo, line 521: I think `parent != “”` would be clearer

kuujo (commented):
Don’t we have the same problem with the `Apply` phase when we update the configuration’s applied values?

pkg/controller/proposal/controller.go
tomikazi: Sure.

pkg/controller/proposal/controller.go
tomikazi: The core of the apply change function was to update the map (as was done before, but outside this function) and then to process the path parents to clear-out any prior deletions.

kuujo (approved):

PR #1547, opened 2022-02-28
Changed discovery of plugins to happen synchronously on startup

tomikazi:


ray-milkey (approved):

teone (approved):

pkg/pluginregistry/registry.go
```diff
@@ -118,7 +118,8 @@ func NewPluginRegistry(endpoints ...string) PluginRegistry {
 
 // Start the plugin registry
 func (r *pluginRegistry) Start() {
-	go r.discoverPlugins()
+	// Discover plugins synchronously on start-up.
+	r.discoverPlugins()
```
teone, line 122: How about using a `WaitGroup`? that way we can load plugins in parallel (not a big concern as I don't think we'll have many for now and it's only a startup operation)

tomikazi: We currently don't load them in parallel, so I think this is fine and simpler.

teone: Good point, we might add that to `discoverPlugins() {` if needed in the future

PR #1567, opened 2022-04-28
Subscription pass-through implementation and integration test

tomikazi:
Still WIP:
- more tests
- adding support for poll


pkg/northbound/gnmi/v2/subscribe.go
```diff
@@ -1,14 +1,182 @@
...
+	"github.com/onosproject/onos-api/go/onos/topo"
+	"github.com/onosproject/onos-lib-go/pkg/errors"
+	baseClient "github.com/openconfig/gnmi/client"
 	"github.com/openconfig/gnmi/proto/gnmi"
+	"io"
+	"strings"
 )
 
+type subContext struct {
+	ctx       context.Context
```
adibrastegarnia, line 21: we should not store context in struct. However it should be passed from one function to another and should flow through your program

FYI https://go.dev/blog/context-and-structs

pkg/southbound/gnmi/client.go
```diff
@@ -31,15 +32,35 @@ type Client interface {
...
 }
 
-// Subscribe calls gNMI subscription based on a given query
+// Subscribe calls gNMI subscription bacc
+//sed on a given query
 func (c *client) Subscribe(ctx context.Context, q baseClient.Query) error {
 	err := c.client.Subscribe(ctx, q)
+	c.lock.Lock()
+	defer c.lock.Unlock()
+	if !c.running {
```
adibrastegarnia, line 46: I am trying to understand the usage of this flag at client level. are you trying to achieve one Subscription stream per target?

adibrastegarnia (changes requested):

pkg/northbound/gnmi/v2/subscribe.go
```diff
@@ -1,14 +1,182 @@
...
+	nbStream  gnmi.GNMI_SubscribeServer
+	sbStreams map[string]gnmi.GNMI_SubscribeClient
```
adibrastegarnia, line 23: How this is used?

pkg/southbound/gnmi/client.go
tomikazi: Yes.

pkg/northbound/gnmi/v2/subscribe.go
tomikazi: Ah, yes. This is a remnant from when I thought I was going to track the sb streams. I've removed this.

pkg/northbound/gnmi/v2/subscribe.go
tomikazi: Done.

adibrastegarnia (approved):

PR #1574, opened 2022-05-20
NB gNMI Set now processes type/version override extension

tomikazi:


pkg/northbound/gnmi/v2/set.go
```diff
@@ -211,35 +217,52 @@ func (s *Server) Set(ctx context.Context, req *gnmi.SetRequest) (*gnmi.SetRespon
...
 	}
 
-	configurable, err := s.getTargetConfigurable(ctx, topoapi.ID(targetID))
-	if err != nil {
-		log.Warn(err)
-		return nil, err
+	// Use the type/version overrides if they are specified
+	var targetType configapi.TargetType
+	var targetVersion configapi.TargetVersion
+	persistent := true // TODO determine whether this is an appropriate default
```
adibrastegarnia, line 235: I think we can still get configurable for this flag and use it. The exception will be just for target type and version when extension is defined

pkg/northbound/gnmi/v2/set.go
```diff
@@ -211,35 +217,52 @@ func (s *Server) Set(ctx context.Context, req *gnmi.SetRequest) (*gnmi.SetRespon
...
 
-	modelPlugin, ok := s.pluginRegistry.GetPlugin(configapi.TargetType(configurable.Type), configapi.TargetVersion(configurable.Version))
+	// Find the model plugin using the target type and version
+	modelPlugin, ok := s.pluginRegistry.GetPlugin(targetType, targetVersion)
 	if !ok {
-		err = errors.NewNotFound("model %s (v%s) plugin not found", configurable.Type, configurable.Version)
+		err := errors.NewNotFound("model %s (v%s) plugin not found", targetType, targetVersion)
 		log.Warn(err)
 		return nil, err
 	}
```
adibrastegarnia, line 257: It looks TargetTypeVersion map in newTransaction function is not initialized with these info. Since you are here you can make the change to pass it down.

pkg/northbound/gnmi/v2/set.go
tomikazi: Fixed.

adibrastegarnia (approved):

tomikazi:
retest this please

PR #1584, opened 2022-06-07
cleanup paths after removing models dependency

SeanCondon:
Also fixing loading of RO paths
and releasing 0.10.35


SeanCondon:
retest this please

adibrastegarnia (approved):

PR #1589, opened 2022-07-29
Check kind of source entity before triggering mastership controller

adibrastegarnia:
This condition is required to avoid writing to an object which is not configurable (e.g. P4 target). Since a control relation can be between onos-config/gNMI target or P4 runtime controller/P4 switch then we need to check kind of source entity to make sure mastership is triggered on relevant  control relation.

### DIFF
--- a/pkg/controller/mastership/watcher.go
+++ b/pkg/controller/mastership/watcher.go
@@ -46,7 +46,13 @@ func (w *TopoWatcher) Start(ch chan<- controller.ID) error {
 			log.Debugf("Received topo event '%s'", event.Object.ID)
 			if relation, ok := event.Object.Obj.(*topoapi.Object_Relation); ok &&
 				relation.Relation.KindID == topoapi.CONTROLS {
-				ch <- controller.NewID(relation.Relation.TgtEntityID)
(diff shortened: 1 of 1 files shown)


adibrastegarnia:
retest this please

pkg/controller/mastership/watcher.go
```diff
@@ -46,7 +46,13 @@ func (w *TopoWatcher) Start(ch chan<- controller.ID) error {
 			log.Debugf("Received topo event '%s'", event.Object.ID)
 			if relation, ok := event.Object.Obj.(*topoapi.Object_Relation); ok &&
 				relation.Relation.KindID == topoapi.CONTROLS {
-				ch <- controller.NewID(relation.Relation.TgtEntityID)
+				srcEntity, err := w.topo.Get(ctx, relation.Relation.SrcEntityID)
+				if err == nil {
+					if srcEntity.GetEntity().KindID == topoapi.ONOS_CONFIG {
```
kuujo, line 51: These two conditions can be done in one line

kuujo: Actually, we should probably check if err != nil and log the error, otherwise check the entity type... an if-elseif statement.

kuujo (changes requested):

PR #1592, opened 2022-09-05
Handled float in onos-config set/get; Fixed proto module import

frouzbeh:


pkg/utils/values/v2/gnmiValueUtil.go
```diff
@@ -40,9 +40,13 @@ func GnmiTypedValueToNativeType(gnmiTv *gnmi.TypedValue, modelPath *pathutils.Re
 	case *gnmi.TypedValue_BytesVal:
 		return configapi.NewTypedValueBytes(v.BytesVal), nil
 	case *gnmi.TypedValue_DecimalVal:
+		// Deprecated: Do not use
 		return configapi.NewTypedValueDecimal(v.DecimalVal.Digits, uint8(v.DecimalVal.Precision)), nil
 	case *gnmi.TypedValue_FloatVal:
+		// Deprecated: Do not use
```
adibrastegarnia, line 46: Use can say use Double value instead.

pkg/utils/values/v2/gnmiValueUtil.go
```diff
@@ -40,9 +40,13 @@ func GnmiTypedValueToNativeType(gnmiTv *gnmi.TypedValue, modelPath *pathutils.Re
 	case *gnmi.TypedValue_BytesVal:
 		return configapi.NewTypedValueBytes(v.BytesVal), nil
 	case *gnmi.TypedValue_DecimalVal:
+		// Deprecated: Do not use
 		return configapi.NewTypedValueDecimal(v.DecimalVal.Digits, uint8(v.DecimalVal.Precision)), nil
 	case *gnmi.TypedValue_FloatVal:
+		// Deprecated: Use TypedValue_DoubleVal instead
 		return configapi.NewTypedValueFloat(float64(v.FloatVal)), nil
+	case *gnmi.TypedValue_DoubleVal:
```
SeanCondon, line 48: This `gnmi.TypedValue_DoubleVal` has only been added recently - I don't think it's valid to turn it in to an internal float. I think you will need to add 2 new types `DOUBLE` and `LEAFLIST_DOUBLE` in https://github.com/onosproject/onos-api/blob/fb7a353930fdf21479c8c53b3d70b49031b4e5a3/proto/onos/config/v2/value.proto#L28 

and then add a `NewTypedValueDouble` method

SeanCondon (changes requested):
@frouzbeh I don't understand why you're using gNMI type Double when the type is defined in YANG as `decimal64` (which is already supported)

PR #1597, opened 2022-10-18
Avoid skipping  the prefix in SB + fix linter issues

adibrastegarnia:
since having prefix is optionally added to the plugin that can be controlled at runtime, we don't need to discard prefix in util function (I should see what would be effect of this change). If the path in the change has prefix should be included in the SB request otherwise it won't. @SeanCondon  what do you think?


adibrastegarnia:
retest this please

PR #1603, opened 2022-12-06
implement LeafSelectionQuery

SeanCondon:


pkg/northbound/admin/admin.go
```diff
@@ -175,3 +177,62 @@ func (s Server) RollbackTransaction(ctx context.Context, req *admin.RollbackRequ
...
+		configuration.NewID(configapi.TargetID(req.Target), configType, configVersion))
+	if err != nil {
+		return nil, errors.NewInvalid("error retrieving config for %s %s %s %v",
+			req.Target, configType, configVersion, err)
+	}
+
+	if req.ChangeContext != nil &&
+		len(req.ChangeContext.GetUpdate())+len(req.ChangeContext.GetReplace())+len(req.ChangeContext.GetDelete()) > 0 {
+
+		log.Warnf("Ignoring change context for the moment")
```
adibrastegarnia, line 209: Here you can just use warn since you are not printing any other objects

pkg/northbound/admin/admin.go
```diff
@@ -175,3 +177,62 @@ func (s Server) RollbackTransaction(ctx context.Context, req *admin.RollbackRequ
 	}
 	return nil, ctx.Err()
 }
+
+// LeafSelectionQuery selects values allowable for leaf.
+func (s Server) LeafSelectionQuery(ctx context.Context, req *admin.LeafSelectionQueryRequest) (*admin.LeafSelectionQueryResponse, error) {
+	log.Debugf("Received LeafSelectionQuery %+v", req)
+	logContext(ctx, "LeafSelectionQuery()")
+	if req == nil {
+		return nil, errors.NewInvalid("request is empty")
```
adibrastegarnia, line 186: You need to use errors.Status here to return error. 
like this https://github.com/onosproject/onos-config/blob/master/pkg/northbound/admin/admin.go#L128

pkg/northbound/admin/admin.go
adibrastegarnia: In gRPC server all of the errors returned back to the user should uses errors.Status.

SeanCondon: done

pkg/northbound/admin/admin.go
SeanCondon: done

PR #1604, opened 2022-12-13
Added capability to restrict EnterpriseAdmin to modify other enterprises data

praj527:
Currently, If any user with the role `EnterpriseAdmin` is able to do modifications in any of the `enterprises` though he does not belong to it via `aether-roc-api`.

With these changes, the `EntepriseAdmin` will only be able to do the modification on the `enterprises` to which he belongs to.

### DIFF
--- a/pkg/utils/rbacevaluate.go
+++ b/pkg/utils/rbacevaluate.go
@@ -6,6 +6,7 @@ package utils
 
 import (
 	"github.com/grpc-ecosystem/go-grpc-middleware/util/metautils"
+	"github.com/openconfig/gnmi/proto/gnmi"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"os"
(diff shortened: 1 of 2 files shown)


pkg/utils/rbacevaluate.go
```diff
@@ -32,3 +33,107 @@ func TemporaryEvaluate(md metautils.NiceMD) error {
...
+		if g == "AetherROCAdmin" {
+			return "AllGroups"
+		} else {
+			groups = groups + semicolon + g
+		}
+	}
+	return groups
+}
+
+func IsAllowToDoSet(allowedGroups, prefixString string, deleteElem []*gnmi.Path, update, replace []*gnmi.Update) bool {
```
SeanCondon, line 49: I don't think this is going to work at all - it's embedding too much in to onos-config that should be handled in OpenPolicyAgent - as it's done for Get. Please see how Get builds a JSON and sends it to OPA. Also see the REGO rules in the helm chart that drives OPA

Here in onos-config there should be no mention of `enterprise` - we have lots of different uses for onos-config where the target is store, city, fabric etc

SeanCondon (changes requested):

PR #1606, opened 2022-12-15
Store Configuration paths/values in separate Atomix map

kuujo:
This PR is a proof-of-concept to improve the scalability of onos-config by storing `Configuration` path/value pairs in a separate, dynamically allocated, per-target Atomix `Map`. This is a solution that was not possible in onos-config until the recent Atomix upgrade introduced support for dynamic primitives. Storing the `Configuration` paths/values in a separate map should eliminate the large request sizes to stores that were causing gRPC size limits to be hit.

Note that this is purely a PoC and is still an imperfect implementation. This solution should scale, but at the cost of violating  the atomicity guarantees we have maintained in the northbound API thus far. Additional work will be required to restore those guarantees. But this should be enough to determine whether this is a viable solution.


SeanCondon:
Hi @kuujo  - I ran this again today with the `scaling` tests (from the main branch) and I can see that this solution performs worse than what we have already. 

For instance running the test below with 100 Sets will fail (rpc error: code = DeadlineExceeded) on my mac book after about 36 transactions, where as the 100 will complete with the `v0.11.3` onos-config.
```
helmit test ./cmd/onos-config-tests -c ../onos-helm-charts --suite scaling -n test --test TestHugeNumberOfSets
```
Is there anything more can be done to improve performance?

adibrastegarnia (approved):

SeanCondon (approved):
Better performance and scalability - LGTM

praj527:
Hi @kuujo 

I tried to run this patch on my local environment.
I am using the below version for the `Atomix` and build the `onos-config` image from this branch [kuujo:configuration-store-maps](https://github.com/kuujo/onos-config/tree/configuration-store-maps)
I saw that `onos-config` pod stuck in `CrashBackLoopOff`, I am attaching the screenshots for reference.
Please let me know, if I missed anything.
 
atomix-version
onos-config-pr-image-2023-02-14 115856

kuujo:
@praj527 I think you need to update Atomix. It looks like the sidecar that's being injected doesn't support the latest version of the API.

kuujo:
retest this please

PR #1607, opened 2022-12-16
[SDFAB-1002,1028] Implementing Cascading Delete functionality

eroshiva:
This PR implements Cascading Delete functionality inside onos-config component. Cascading delete means that when the intermediate path is being deleted, it also deletes all its children.
Integration test is also a part of that PR.


pkg/controller/proposal/controller.go
```diff
@@ -818,3 +820,24 @@ func isModelDataCompatible(pluginDataModels []*gpb.ModelData, targetDataModels [
 	}
 	return true
 }
+
+func doCascadingDelete(changeValues map[string]*configapi.PathValue, configStore map[string]*configapi.PathValue) map[string]*configapi.PathValue {
```
SeanCondon, line 824: I think a better name might be `addDeletedChildren` and the comment should go on the line before

pkg/controller/proposal/controller.go
```diff
@@ -456,7 +456,8 @@ func (r *Reconciler) reconcileCommit(ctx context.Context, proposal *configapi.Pr
 			if config.Values == nil {
 				config.Values = make(map[string]*configapi.PathValue)
 			}
-			for path, changeValue := range changeValues {
+			updChangeValues := doCascadingDelete(changeValues, config.Values)
```
SeanCondon, line 459: this `addDeletedChildren` should be conditional on a flag from modelPlugin ModelInfo

pkg/controller/proposal/controller.go
```diff
@@ -642,9 +643,10 @@ func (r *Reconciler) reconcileApply(ctx context.Context, proposal *configapi.Pro
 			changeValues = proposal.Status.RollbackValues
 		}
 
+		updChangeValues := doCascadingDelete(changeValues, config.Values)
```
SeanCondon, line 646: I think `upd` is a bit odd - suggest `updated`

pkg/controller/proposal/controller.go
```diff
@@ -818,3 +820,24 @@ func isModelDataCompatible(pluginDataModels []*gpb.ModelData, targetDataModels [
...
+
+func doCascadingDelete(changeValues map[string]*configapi.PathValue, configStore map[string]*configapi.PathValue) map[string]*configapi.PathValue {
+	// defining new changeValues map, where we will include old changeValues map and new pathValues to be cascading deleted
+	var updChangeValues = make(map[string]*configapi.PathValue)
+	for prefix, changeValue := range changeValues {
+		// if this pathValue has to be deleted, then we need to search for all children of this pathValue
+		if changeValue.Deleted {
+			for path, value := range configStore {
+				// this condition will be true, when pathValue will find itself in the store,
+				// so there is no need to store it again on line 473
```
SeanCondon, line 832: I did not understand the comment re line 473.
Also consider that a deleted object and a child might already be in the `changeValues` - it should not matter as the map will guarantee uniqueness

pkg/controller/proposal/cascading-delete_test.go
```diff
@@ -0,0 +1,114 @@
...
+		},
+		Deleted: true,
+	}
+	log.Infof("changeValues is\n%v", changeValues)
+
+	// cascading delete algorithm
+	updChangeValues := doCascadingDelete(changeValues, store)
+	log.Infof("updChangeValues is\n%v", updChangeValues)
+	assert.Equal(t, 5, len(updChangeValues))
+	log.Infof("updChangeValues has %d PathValues to delete", len(updChangeValues))
```
SeanCondon, line 113: I think you should be checking the contents too

pkg/controller/proposal/controller.go
eroshiva: Renamed

pkg/controller/proposal/controller.go
eroshiva: Changed per suggestion

pkg/controller/proposal/controller.go
eroshiva: Discussed in a call - no need to implement a flag so far

pkg/controller/proposal/cascading-delete_test.go
eroshiva: Added

pkg/controller/proposal/controller.go
eroshiva: Indeed, this comment is very confusing. I completely removed it..

eroshiva:
Pinging @adibrastegarnia and @kuujo for a review as well

Andrea-Campanella (approved):
Generally LGTM, I need to test this with the autogenerated ui, cc @teone

test/config/cascadingdelete.go
```diff
@@ -0,0 +1,125 @@
...
+	gnmiapi "github.com/openconfig/gnmi/proto/gnmi"
+	"github.com/stretchr/testify/assert"
+	"testing"
+	"time"
+)
+
+// TestCascadingDelete checks that when a root leaf or an intermediate leaf is removed, all its children are removed too
+func (s *TestSuite) TestCascadingDelete(t *testing.T) {
+	const (
+		switchName      = "testswitch"
```
adibrastegarnia, line 20: You need to change these paths to use gnmi simulator model for testing

adibrastegarnia (changes requested):

test/config/cascadingdelete.go
eroshiva: Adjusted per our discussion

Andrea-Campanella (commented):
great work! I'd say we wait to merge this until @charlesmcchan makes the release. It's a known issue, might want to do a point release with this and removing the workaround in UI and ONOS. cc also @daniele-moro

teone:
@Andrea-Campanella I don't think this level changes will have any effect on the UI.
@frouzbeh Just out of caution, if you have spare cycles today, can you try to run it with the sd-fabric models and the latest UI?

eroshiva:
retest this please

adibrastegarnia:
The integration test for cascading delete need to be fixed before we can merge this.

SeanCondon:
> @Andrea-Campanella I don't think this level changes will have any effect on the UI. @frouzbeh Just out of caution, if you have spare cycles today, can you try to run it with the sd-fabric models and the latest UI?

Agreed - this has nothing to do with the UI

adibrastegarnia:
@eroshiva 
FYI: I noticed some of the basic integration tests (specially the ones that delete leaf values) also affected (we have a bunch of them in test folder). If you were seeing the tests were passing perhaps you were deploying a wrong image that doesn't have your changes or something like that.

frouzbeh:
@teone @Andrea-Campanella 
I agree these changes has nothing to do with the gui, but I've tested and I wasn't able  to delete even leaf attributes. 
Previously, when we deleted a top-level model (e.g. switch), we had `deleted: true` for this  path `/switch[switch-id=<id>]`, but not its leaves and attributes. So the gui wouldn't show the model because the top-level model was tagged with deleted. 
With these changes,  I still can see the "deleted" switch in the gui, and I noticed there is no `deleted` tag.

Andrea-Campanella:
> @teone @Andrea-Campanella
> I agree these changes has nothing to do with the gui, but I've tested and I wasn't able to delete even leaf attributes.
> Previously, when we deleted a top-level model (e.g. switch), we had `deleted: true` for this path `/switch[switch-id=<id>]`, but not its leaves and attributes. So the gui wouldn't show the model because the top-level model was tagged with deleted.
> With these changes, I still can see the "deleted" switch in the gui, and I noticed there is no `deleted` tag.

AN update here, @eroshiva found issues in his implementation that he is working out with @adibrastegarnia. please refrain from further testing until notified. thanks for all the effort.

pkg/controller/proposal/controller.go
```diff
@@ -456,7 +456,11 @@ func (r *Reconciler) reconcileCommit(ctx context.Context, proposal *configapi.Pr
 			if config.Values == nil {
 				config.Values = make(map[string]*configapi.PathValue)
 			}
+			// ToDo - remove before merging
+			log.Infof("Commit phase - Following paths are required to be updated:\n%v", changeValues)
```
SeanCondon, line 460: it might be better to make log.Debugf, and set the logging level to activate - then they could be left in permanently

pkg/controller/utils/utils.go
```diff
@@ -5,14 +5,48 @@
...
 // GetOnosConfigID gets onos-config URI
 func GetOnosConfigID() topoapi.ID {
 	return topoapi.ID(uri.NewURI(
 		uri.WithScheme("gnmi"),
 		uri.WithOpaque(env.GetPodID())).String())
 }
+
+// AddDeleteChildren adds all children of the intermediate path which is required to be deleted
+func AddDeleteChildren(changeValues map[string]*configapi.PathValue, configStore map[string]*configapi.PathValue) map[string]*configapi.PathValue {
+	log.Debugf("Obtained changeValues are:\n%v", changeValues)
```
Andrea-Campanella, line 27: might want to remove this log as well

pkg/controller/proposal/controller.go
```diff
@@ -456,10 +457,19 @@ func (r *Reconciler) reconcileCommit(ctx context.Context, proposal *configapi.Pr
 			if config.Values == nil {
 				config.Values = make(map[string]*configapi.PathValue)
 			}
-			for path, changeValue := range changeValues {
-				_, _ = applyChangeToConfig(config.Values, path, changeValue)
-			}
+			// ToDo - remove before merging
+			log.Debugf("Commit phase (461) - values are:\n%v", config.Values)
+			log.Debugf("Commit phase (462) - Following paths are required to be updated:\n%v", changeValues)
+			updatedChangeValues := controllerutils.AddDeleteChildren(changeValues, config.Values)
```
Andrea-Campanella, line 463: I think we agreed of handing this behind a variable, maybe not all targets want a cascading delete to happen. @adibrastegarnia @SeanCondon am I correct ?

Andrea-Campanella (commented):
code mostly looks good, 2 comments. Once those are addressed I'd like @ray-milkey and @frouzbeh to test this please. I will wait for their review before any merge.

pkg/controller/proposal/controller.go
adibrastegarnia: I think cascading delete is something generic. That means when you delete an intermediate node in the config tree, it makes sense the whole subtree gets deleted. If the user doesn't want to delete the whole subtree then the user should just delete leaf values or just the paths that it cares about. I don't think enabling/disabling per target makes sense in general.

eroshiva:
retest this please

pkg/controller/utils/utils.go
eroshiva: removed

pkg/controller/proposal/controller.go
eroshiva: Log was completely removed - no need to keep it

adibrastegarnia (approved):

frouzbeh:
@eroshiva @Andrea-Campanella 
I just tested this pr with the adaptive gui, and it seems everything works as expected. I was able to delete from list, or unset values. And in the configuration all deleted paths are set with `deleted: true` as expected.

PR #1608, opened 2023-01-05
OIE-1844 add a limit to the site of a gNMI Set

SeanCondon:
Also added a new suite of integration tests for testing the scalability


eroshiva (approved):

test/scaling/hugenumberofsets.go
```diff
@@ -0,0 +1,94 @@
...
+
+const (
+	configNamePath    = "/config/name"
+	enabledPath       = "/config/enabled"
+	descriptionPath   = "/config/description"
+	mtuPath           = "/config/mtu"
+	holdTimeUpPath    = "/hold-time/config/up"
+	holdTimeDownPath  = "/hold-time/config/down"
+	longIfDescription = "this is a long description of the interface"
+	iterateCount      = 100
```
adibrastegarnia, line 28: Is this the threshold that if we go beyond this we hit the limit or for now it is just a value for testing?

SeanCondon: @adibrastegarnia I set it to 100 as I found that on my system the performance after about 160 Set operations slowed down to the point that the the gNMI timed out. 
I am going to write some more tests in the coming days to deal with different overloading challenges

PR #1654, opened 2023-06-23
Add TLA+ spec for new controller logic

kuujo:
This PR is the initial step towards simplifying and stabilizing the onos-config core code base. I'm going to slowly work in the changes as I find time to clean them up and test them so we don't have a massive PR to review and test.

The first piece is adding the updated TLA+ spec, which is used to verify both the design and implementation of the controllers in onos-config. This spec is far more thoroughly implemented and verified than it ever was before. I'm now very confident in the safety and stability of the design. The model checker confirms that the logic does not contain any consistency issues and, more importantly, contains no deadlocks, which has been too much of an issue until now.

Most importantly, the model checker is now capable of generating a full suite of tests from the spec to be run against the actual code implementing the spec. This is a feature I've been working towards on and off for more than a year now. Generating test cases from the formal model ensures that absolutely every scenario and every possible edge case in the controller logic will be covered by the unit tests. Effectively, it provides formal verification of the code itself. Every check that's performed by the model checker on the spec will be performed on the code, as well.

I'll follow up this PR with the controller changes and test runner when I have the time to clean up that portion of the changes.


kuujo:
CI is failing due to the missing copyright headers. I'm not really sure how to handle the copyright for the TLA+ specification. It's not really a code file, per se. One of its uses is to generate a PDF of the specification, and any comments I put in the spec will be printed in the PDF, so I suspect the SPDX style comments may not be suitable in this case. What would likely be more appropriate would be to just add a standard copyright notice to the footer of each `.tla` file, as would typically be done with a document. Alternatively, these files can just be excluded, but I don't think that's what we want. 

@scottmbaker what do you think?

SeanCondon:
> CI is failing due to the missing copyright headers

According to [this page](https://www.volkerkrause.eu/2021/11/13/kde-reuse-compliance-tips.html) you can have a `.license` file in the same directory - spares putting License in to file

SeanCondon (approved):

kuujo:
Perfect I'll update the PR

kuujo:
retest this please

PR #1657, opened 2023-06-27
Add model based test engine

kuujo:
This PR adds the engine for running unit tests generated by the TLC model. Tests are run by defining a test state and context struct and then calling `RunTests`. Once #1656 is merged, I will submit an additional PR adding the tests to the controllers.


kuujo:
retest this please

